PR #29, opened 2015-03-03
Joystick support

ErikDeBruijn:


daid:
Got a bunch of problems with this pull request.
1) Contains data also in your other pull request (acceleration stuff, which we talked about, as I wanted that in m/s^2)
2) EMissileWeapons(0) is wrong, that should be MW_HomingMissile
3) UI is running on the client. So the whole plasma bolt firing code is at the wrong place, on multiple accounts.
4) The "plasmaBuildup" is framerate depended in your implementation.
5) The "PlasmaBolt" class does not implement the proper way for Multiplayer replication. REGISTER_MULTIPLAYER_CLASS is never defined for it, and handing the "name" for that shouldn't be left to the creator of that object. (See the asteroid for most basic example)

ErikDeBruijn:
Re 1) I have no indication as of what the units currently are, so I thought I'd take it one step at a time and just make the existing parameter configurable. I'll try to pull apart the different commits to isolate this. Also, I still have to derive conclusions from observations, since there are hardly any units in the code. 100% speed of a fighter is 6.6 km/minute and it's set to 110, so that means the unit is m/s. Is delta the "game" time elapsed in seconds since the previous loop?
2) Okay, no problem.
3) I'll look into this!
4) I'll try to use the delta, then. There's no difference between built-up or not built up, yet. But I wanted to share small commits instead of taking longer and making it harder to merge.

For now I'll close this pull request and start working on better ones.

PR #32, opened 2015-03-08
Added getters and setters to HTTP-server

vslotman:
I've added a "get.lua" and "set.lua" which get and set game-data using the query string. 

Example Getter: /get.lua?hull=getHull()&nukes=getWeaponStorage("nuke")
This creates the following LUA-code:
object = getPlayerShip(-1)
if object == nil then return {error = "No valid object"} end
return {hull = object:getHull(), nukes = object:getWeaponStorage("nuke"), }

Returns:  {hull = 100, nukes = 42}

Example Setter: /set.lua?setShieldsActive(true)&setSpeed(200, 3)
object = getPlayerShip(-1)
if object == nil then return {error = "No valid object"} end
object:setShieldsActive(true);
object:setSpeed(200, 3);

Returns nothing, or ERROR on failure.

By default getPlayerShip(-1) is used as the object. When you pass \_OBJECT\_=someOtherObject(2) as parameter, then \_OBJECT\_ gets used.


daid:
Oh. Lots of things to comment on here. First, I like your enthusiasm.

But I have some problems with the implementation:
1) Instead of adding the odd "permission" field to the HttpConnection, you could have registered multiple HttpHandlers to the HttpServer object, one for each request type (get/set/exec), then you can just add those depending on what you want to expose.
2) You pushed the options.ini file.
3) setLoadTube, setFireTube and setInitJump don't set anything, they initiate things. And thus shouldn't be prefixed with "set".
4) setSpeed modified the "currentImpulse" which is a value updated from the spaceShip every game tick. And "rotationSpeed" which is a constant value depending on which ship you are flying. So that does not make sense. Same issue for setWarpSpeed, which directly modifies the amount of active warp, not the amount requested or the amount of speed that you have during warp. (Yes, I know the naming on these isn't very specific)

vslotman:
Well, that's what code review is for ;) 

1) I didn't really dive into the workings of that part, didn't know that was possible... I'll have a look later. Or maybe just remove the whole permissions-thing alltogether, since it's basically just a fancy feature.. Got a bit carried away :)
2) Okay, but how should I leave hints for config parameters?
3) I mainly just named them setStuff to get a clear naming convention when using permissions. I'll just rename them to their proper name.
4) What variables should I set for these settings?

vslotman:
What do you think of it now? 
I've removed all the permissions-stuff from the getters and setters and changed the names of the script-calls. Also removed my options.ini and changed the reference from sFALSE to an empty string.

Github tells me that it has merge conflicts, however I can merge locally without any issues... Do you have an idea what's happening here?

nallath:
Hmm that is strange. @daid, any idea what might cause this?

daid:
Most likely github refuses to auto-merge unless the fork has all the commits of the origin.

vslotman:
I've fixed the merge issues :) I guess I forgot to do a pull before merging branches... Doh!

PR #37, opened 2015-03-10
Cleaned up the file a bit, added descriptions for all ships. Made sizeCl...

ErikDeBruijn:
...ass definitions explicit without changing them instead of assuming the default in the C++ code.


daid:
And you changed the number of missiles to 2 for one ship?

ErikDeBruijn:
Apparently I did. This does make a lot of sense when the plasma weapon is integrated again (which adds tremendous fire power if you can spare the energy), but implementation of the joystick in a way you'll more likely be happy with has been a bigger adventure than I initially thought. So I can't also add plasma right now.

PR #41, opened 2015-06-18
cleanup of HTTP getters and setters

vslotman:
I cleaned up my previous pull-request after I ended up in a horrible merge-hell. 
Info from previous request:

I've added a "get.lua" and "set.lua" which get and set game-data using the query string.

Example Getter: /get.lua?hull=getHull()&nukes=getWeaponStorage("nuke")
This creates the following LUA-code:
object = getPlayerShip(-1)
if object == nil then return {error = "No valid object"} end
return {hull = object:getHull(), nukes = object:getWeaponStorage("nuke"), }

Returns: {hull = 100, nukes = 42}

Example Setter: /set.lua?setShieldsActive(true)&setSpeed(200, 3)
object = getPlayerShip(-1)
if object == nil then return {error = "No valid object"} end
object:setShieldsActive(true);
object:setSpeed(200, 3);

Returns nothing, or ERROR on failure.

By default getPlayerShip(-1) is used as the object. When you pass _OBJECT_=someOtherObject(2) as parameter, then _OBJECT_ gets used.

### DIFF
--- a/src/httpScriptAccess.cpp
+++ b/src/httpScriptAccess.cpp
@@ -1,6 +1,8 @@
 #include "httpScriptAccess.h"
 #include "gameGlobalInfo.h"
 
+#define sOBJECT "_OBJECT_"
+
 bool HttpScriptHandler::handleRequest(HttpRequest& request, HttpServerConnection* connection)
 {
     if (request.path == "/exec.lua")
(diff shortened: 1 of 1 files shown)


daid:
Looks good.

PR #43, opened 2015-07-30
Wormholes

vslotman:
I've added a wormhole. It works kinda like a black hole: It pulls objects in, and spits it out at a user-defineable location. 
When creating a wormhole in the GM-screen, its target is set to (0,0). Target can be changed by selecting the wormhole and clicking RMB at the new location.
The position of the target is indicated with a line.


ErikDeBruijn:
Awesome!

daid:
Cool stuff. Looks fine. I do see a small potential issue, but just fixed that myself. Nothing that can cause huge problems.

PR #44, opened 2015-07-30
Fixed possible bug and added logging in channel-parsing

vslotman:
When I define a channel with #X, it gets mapped to #X-1
I've also added some logging.


daid:
I think, as far as I can find, DMX512 starts counting channels at 1, that's why the - 1 is there.

daid:
But Artemis seems to start at 0, so better match that.

PR #70, opened 2016-01-03
Added custom radar images

Fouindor:
Added the possibility to use custom radar images for ships and stations. These images can be set using the new setRadarTrace(string filename) function.

If a ship is not identified, it will display using the default "RadarArrow.png" image. If identified by the science station and that a custom image has been set, it will display using the custom image.

Differs from my previous pull request in that I have added that if a radar image is not set, it will default to the standard image for the object.

### DIFF
--- a/scripts/shipTemplates.lua
+++ b/scripts/shipTemplates.lua
@@ -2,23 +2,19 @@
 template = ShipTemplate():setName("Small Station"):setModel("space_station_4"):setType("station")
 template:setHull(150)
 template:setShields(300, 0)
-template:setRadarTrace("RadarBlip.png")
 
 template = ShipTemplate():setName("Medium Station"):setModel("space_station_3"):setType("station")
 template:setType("station")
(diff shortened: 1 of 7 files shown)


daid:
Looks good. See a few minor things that could be improved in your coding skills, but overall it's a good patch.

"bools" for example should be set to "true" or "false", instead of 1 and 0. Does the same, but makes it more readable.

daid:
Did some extra work on your patch. Also added functions to overrule the custom icon on a ship-by-ship basis during scenarios.

Fouindor:
Thank you ! :)
I'm still learning a lot of things about "real" coding ^^

Indeed, having a function that sets an icon on-the-fly is good, especially for designating a "special" ship without setting a new template.

daid:
Also allows you to "mark" a ship on the radar, by using the same icon with a circle around it.

Learning is good, that's why I tell you what I did and why :-)

PR #73, opened 2016-01-05
fixed type not being set

Fouindor:


Fouindor:
Could not spawn player ships from the server menu in latest version. It seems that the type was not correctly set following latests patches. Tried with the added line, seems to work fine.

daid:
oops...

PR #74, opened 2016-01-05
Ability to change beam weapon color

Fouindor:
Ship beam weapon color can be set with setBeamWeaponColor(int index, string texfile)
Template beam color can be set with SetBeamColor(int index, string texfile)

If no color is provided, default texture is "beam_orange.png".

I'm not sure that the solution I provided is the most optimized but it seems to work :)


nallath:
I'd say that beam color would be "orange" and that the code that selects the image automatically prepends the beam_ and adds the .png.
This results in easier to read code, as the beam color value actually gives you a color (eg; blue, orange, green) instead of a 'mangled' beam_orange.png when you request the color.

daid:
@nallath true, but what if you want an electric beam look?
Example: http://www.layerpaint.com/admin/new_cropper/tutorial_content_images/1852_tid_main.jpg

Also, the yellow beam texture seems to be a bit too light, and isn't black at the edges.

Fouindor:
Maybe I should change it to setBeamWeaponTexture(int index, string texfile) and SetBeamTexture(int index, string texfile) to avoid confusion.
Modified the yellow beam texture, should be better : http://i.imgur.com/pvj7BaN.png

daid:
Yes, setBeamWeaponTexture would be better.

Fouindor:
Ok, done.
(sorry for all the commits, modified via web)

daid:
Applied a few small fixes.

You missed a rename in shipTemplate.cpp
I renamed "texture" to "beam texture", in case we also want to change the fire-ring texture later on.
I added bounds checks for the setBeamTexture function, as they where missing, which means you could crash the game with incorrect indexes.

Fouindor:
Woops, thanks.
Yes, I think it is a good thing, I'm really tempted to make a lightning beam just to see how it would do :)

PR #77, opened 2016-01-06
EdgeOfSpaceSuggestedChanges

Wrongtown:
I _think_ I might actually have managed to get this done at last? I appreciate the guidance @daid & @Kwadroke have provided, I'm definitely still coming to grips with this set of tools.

As noted in my comments I was unable to find anything indicating that a WAIT function (or similar) exists in .lua and so I hacked together a kind of Rube Goldberg machine to serve the purpose. If there's a simple method that can be called to provide pauses between events then I'd love to be introduced to it.


Fouindor:
For the wait, add a variable that keeps track of time in your update function :
-- mission_timer progress
mission_timer = mission_timer + delta

It will be the time in seconds. At the start of when you want the wait to start, set it to 0. Add a condition on it where you want and it should be ok.

Wrongtown:
Well that's not nearly as fun as my clumsy spaceship-based clock ;)

I think I've *almost* got it but I'm getting complaints about trying to perform arithmetic on a variable that doesn't have an integer value so maybe I'm being too literal.

Just before I want to use the timer, I have the following (with **inspection_progress** instead of **mission_timer**):

    inspection_progress = 0
    inspection_progress = inspection_progress + delta

after which I resolve the rest of the event and have the following check immediately after that:

    	if Central_Command.mission_state == 2 then 
    		if inspection_init == 1 then
    			if inspection_progress >> 10 then
    		Technical_Officer:sendCommsMessage(Player,[[10 seconds passed.
    
    	Fouindor is pretty great.]])
    			end
    		end
    	end

Which is where the stack starts screaming about my abuse of logic. What am I missing here? Or should

Fouindor:
The stack starts screaming because >> is not correct ^^
It should be : inspection_progress > 10 (> is the correct operator : http://www.lua.org/pil/3.2.html)

However, inspection_progress = inspection_progress + delta should be placed after the  "if inspection_init == 1 then", because : if you started counting, then you count.

Hope it helps :)

lunerfox:
Looks like it has to do with your use of the >> operator; did you mean to use > instead?

daid:
(You had two pull requests, both contained the same code, so I merged the other one and will close this one)

Wrongtown:
Cheers all, sorry about the duplicate requests, clearly I need to be more careful with my branching.

PR #81, opened 2016-01-07
Added Science console zoom button + fixed yellow beam texture

Fouindor:
Tested the button, works ok.
Made the beam texture darker with proper borders


Fouindor:
I did not manage to create several pull request, sorry.

PR #85, opened 2016-01-08
Fighter docking

Wrongtown:
As per my earlier comments this seems pretty great but for all I know it was intentionally removed. If so the request can be rejected, if accidentally removed this should put it back in.


nallath:
I don't think this was very intentional. Me and Daid discussed the power issues at some point in time and we quickly came to the conclusion that we need some sort of "absolute" power units (eg; let a cruiser generate (and use) 20x times as much power as a fighter). This way having multiple fighters will be a drain on the cruiser, but not launching the 20 fighters could even be helpfull (20 times a small generator powering your batteries might actually help a bit)

Wrongtown:
Well that's good to hear because the idea of one or more fighters latching on before a jump makes me all warm and fuzzy inside :) 

I like the sound of that if I understand correclty; if they're all full of juice then they're pushing it back into the cruiser, but if lots of hungry babies come home to mama she could get drained awful fast?

Maybe the cruiser (Relay?) should have to acknowledge power couplings before the fighter can start getting energy? I'm thinking sometimes you might want to signal for a retreat, dock all the fighters and jump somewhere far away to lick your wounds. That won't work if by the time the last fighter connects the cruiser doesn't have enough in the reserves to make a jump.

daid:
It was never added, so it was never intentionally removed.

There are some issues with the current ship to ship docking. Energy and shields for example are recharged at an increased rate when you are docked, no matter what you are docked to. Which is why I reserved the size classes for future use.

Wrongtown:
Ah I see, sorry I thought I saw it in an earlier Pull Request, must have been one closed without merge.

Wrongtown:
By the way should this be closed (un-merged) and added as an Issue/Feature Request for later attention?

nallath:
Well, if we want to add this, we should add it "right". This would mean re-doing the docking stuff correctly. Space stations will probably need to have power as well (but much greater than a regular ship)

I love the feature of being docked. It makes the scenic space opera stuff possible of having a carrier jump into the fray and release a bunch of fighters to attack enemies.

daid:
As @nallath says. We need to work out the energy mechanic better first (code wise). Then we can fix up the docking behavior better.

Currently only player ships have energy, and this is causing more complexity in other areas of the code that use energy as well.

Wrongtown:
Closing as Issues #92 and #93 have been raised.

PR #87, opened 2016-01-09
Added the Ghost from the Past scenario

Fouindor:


daid:
Didn't check every bit of code, but in general it looks fine :-)

PR #89, opened 2016-01-11
Added Engineering Crew variables

Fouindor:
Created 3 engineering crew variables : 
- engineering_crew_max, which is the max number of engineers in the spaceship. Defaults to 50, can be set in the template by setEngineeringCrewMax(int number)
- engineering_crew, which is the number of engineers currently on board. Equal to engineering_crew_max on ship spawn. Can be set on-the-fly via setEngineeringCrew(int number)
- engineering_crew_injuried, which is the number of engineers injuried.

Right now these values only display and have no incidence on repair rate. Furthermore, there is no mechanism to wound crew as of now.


Fouindor:
Wow, something went wrong. I'll reopen when I see what is wrong with my sync.

Fouindor:
Seems that I didn't updated correctly my fork, re-submittting modifiactions.

PR #90, opened 2016-01-11
Added engineering crew variables

Fouindor:
Created 3 engineering crew variables :

- engineering_crew_max, which is the max number of engineers in the spaceship. Defaults to 50, can be set in the template by setEngineeringCrewMax(int number)

- engineering_crew, which is the number of engineers currently on board. Equal to engineering_crew_max on ship spawn. Can be set on-the-fly via setEngineeringCrew(int number)
- engineering_crew_injuried, which is the number of engineers injuried.

Right now these values only display and have no incidence on repair rate. Furthermore, there is no mechanism to wound crew as of now.


nallath:
I didn't check this completly, but could you also add a variable to the server that enables/disables the option? A lot of the more advanced stuff can be enabled/disabled there and this feels like one of those functions.

daid:
engineering_crew_max is the template is a float, while in the ship it is an integer?

I'm unsure on how this ties into the current repair mechanics. As far as I can see your additions do not do anything yet except for showing a value on some stations.

Issue #97, opened 2016-01-13
REQUEST - Physical Probes


Wrongtown:
As referenced in #96 the Probes launched by Relay don't "physically exist" in the game, I've had a couple of players comment "Where is our probe? Should be here but I can't see it."

A model for the probe would add to immersion, and also allow for the following mechanics to be developed:
* Destruction of probes by other forces.
* Collection of probes to re-use.
* Re-stocking probes at stations.

**Example**
```lua    
--Probe, for augmenting the normal sensor suite. There's a fine line between exploration and espionage.
    template = ShipTemplate():setName("Astrometrics Probe"):setModel("SPAAAACE_core")
    template:setHull(20) --Maybe even lower?
    template:setShields(5) --Maybe even lower?
    template:setSpeed(1000, 10, 15) --I'm not sure how fast probes currently move.
    template:setDescription([[The Astrometrics Probe is an invaluable tool for mapping out sectors where no-one has previously gone as far as we are aware. Tiny and incredibly fast, the probe is essentially a missile with a payload of sensors and a transmitter rather than destructive elements.  Due to the extremely lightweight build there's not much in the way of armor or shielding; probes discovered by hostile or defensive entities do not provide survey information for very long at all.]])
```

daid:
Just need a proper model for them. As the "exist" and "not exist" at the same time. They are there, but do not have a 3D model.

Wrongtown:
That was an excellent excuse for me to check out the various Artifact models (still hadn't scripted up a museum of the little blighters) but sadly none of them really fit the bill do they?

By the way if anyone has been meaning to take a look and lazy like me, I've attached a .txt copy of the lua script I created with all the unusual objects lined up for your observation. Just change the extension to .lua and drop it into the /scripts/ directory and it will be added to the bottom of the list.

[scenario_99_artifacts.txt](https://github.com/daid/EmptyEpsilon/files/88565/scenario_99_artifacts.txt)

Fouindor:
Maybe one of SolCommand's models for communication probes :
https://public.blu.livefilestore.com/y3pa_fhFR8hiZSndVwdLoIkkAXmJBI15a0qwWIRqe_eC6SP-Eu0uNWQUHcKfe20rvGDgUfePwiAwb6FQh3zyMpmxn8wYV1QxLI99OAQSewy9QY/communication_satellite_v2.jpg?psid=1
[...]

PR #104, opened 2016-01-18
Science now able to see with probes

Fouindor:
Added the capacity for Relay to feed probes to science.
Science can now switch between the standard ship view or a 5-km short range radar centered on the selected probe.
Tested the best I can, it seems to be working correctly. The only inelegant thing is when the probe is moving.
To make the probe selectable in relay, I had to make it targetable. I changed the probe faction to independant, in order to not have the enemies attack them.


daid:
Instead of "canBeTargeted" you can "canBeSelected" to make probes selectable. Introduced this to support of selecting objects that have a description.

Other then that, "local" is quite a badly named variable. As I have no clue what it's suppose to do.

Wrongtown:
As I think I suggested in #97 I would say let enemies target and destroy the probes. It's more realistic and adds to the fun if you have to weigh up whether sending a probe into enemy turf is going to work for you IMHO.

Though thinking about it just now that does raise one extra "nice to have" for Relay:

*As Relay officer I want to be notified when a probe is being attacked so I don't lose track of my probes.*

Fouindor:
Thanks for the correction. The "local" variable of the radar was added to inhibit re-centering on ship when observing a probe with. Would "inhibit_centering" be better ?

Indeed, making enemies able to kill probes was on my mind as it would be a further improvement, but there is several concerns as of now :
- Probes cannot be killed for now (didn't had time to implement this, as for new they are game-breaking, that's why I switched them to independant)
- Probe do not have a 3D model (not gamebreaking per se, but still nice to have)

Wrongtown:
:+1:

Fouindor:
Ok, cleared things a bit and added a check that does not let you send a probe that is not on its target (i.e., moving). Tested and seems to work ok. Also, solved a situation in Ghost from the Past scenario where losing Stakhanov would result in a Human (!) victory

daid:
I can fix the targeting, do not worry about that (as I also need to fix up the AI to not prioritize probes too much)

I do have a few small things. "bool" variables should be set to "true" and "false" instead of 1 and 0.
Also, you code doesn't work across the network. I can explain what needs to be done for this, but you'll have to understand a bunch of concepts.

Fouindor:
D'oh, didn't thought about testing on different clients. Given what you already corrected in the past in my submissions, it should not be working because I do not replicate the tracked object, resulting in : if the relay links a probe to Science, Science is none the wiser, because it is not replicated (Relay update ITS my_spaceship->tracked_object, but not the server's and none of the other player's).

I'm sorry to waste your time, working on an open project like this is a good way for me to learn some things :)

I will try to correct this (one day, I will get it right ^^) once I get back on my computer.

Fouindor:
Okay, I’m sending yet another revision  :)
This time :
-	Fixed the bool status, as instructed.
-	Using the linked probe as a variable of a player spaceship was quite clumsy, and absolutely not adapted to network replication. Instead of the probe, its multiplayer ID is used. Seems to work about right.

I have another question: can multiplayer IDs be reused ? If yes, I would need to add that when a probe dies, it must check if it is the object pointed by its owner. If yes, set the pointed ID to NULL. Else it would lead to an issue with my probe implementation:
•	Relay launches probe (ID #112)
•	Probe dies
•	A new object takes ID #112
•	Science can observe the new object, as it is valid (!)

I’ve also had a bit of a Github mishap, but everything seems in order now.

daid:
MultiplayerIDs are never re-used during the runtime of a server, so this is fine on that aspect, good job! Love it that you get this far with just a little help from my side!

Almost ready for action now. There is just 1 bug related to multiplayer in there now. If the server is relay, and the client is science, it works now. But if the server is science, and the client is relay, the server never knows that "linked_object" should be changed.

This is by far the least "coder friendly" piece to fix. But seeing how far your got with little info, I give you 1 hint to look at "playerspaceship::commandSetTarget"

Fouindor:
So if I understand what you are saying correctly, if a replicated variable is modified on the server, it is automatically updated on all clients, but when a client modifies a replicated variable, it must push manually the change using sendClientCommand ? (I thought quite naively that if the variable was replicated it was updated no matter the case)

Thanks for the hint, I see (or at least suppose to see ^^) where I should be going :)

Fouindor:
That (I hope so ^^) should do the trick. Tested it in the 2 configurations.

daid:
Looks good now. When I have the time, I'll start working on AI ships targeting and destroying nearby probes when they have nothing else to do.

daid:
I've added that AI ships can target and destroy probes:
https://github.com/daid/EmptyEpsilon/commit/52d2e9ef1392e0c2e202f6aef9598c29d4b6b3d4

They will only target probes within 5km, and only when they have no other target that is more important.

Should protect enemy bases from "hostile" probes.

Fouindor:
Thanks, looks nice ! :)
I think I will submit a partial code for translation soon for review if possible, to see if its an efficient way to implement it (to know if I must start again for scratch or not ^^)

daid:
I did the start of translations for a work project a few weeks back. Most of the work was in wrapping all the strings in a function and writing the translation file generation&update scripts.

Fouindor:
I've had a certainly more archaic approach, adding a language.cpp and language.h files which hosted all translation as nested switch/cases :

string fetchGUITranslation(int16_t targ){
switch (targ) {
[...]
    case   CANCEL_LABEL:
    {
        switch (game_language){
            case LANG_EN:return "Cancel";
            case LANG_FR:return "Annuler";
            case LANG_ES:return "Cancelar";
            default:return "Cancel";
        }
    }

    case   ANSWER_LABEL:
    {
        switch (game_language){
            case LANG_EN:return "Answer";
            case LANG_FR:return "Répondre";
            case LANG_ES:return "Contestar";
            default:return "Answer";
        }
    }
[...]
}

All translations are stored in the .cpp file. Even if it is clumsy, it works :)
But there is no external scripts to update/generate in my approach.

daid:
Sorry, but that's a horrible way to solve it. Because it puts a lot of manual work in adding extra texts.

Fouindor:
No problem, I need feedback to learn ;)
So better would be to have an external file per language, loaded on start or on language change.
(In my vision, a bit like how options are stored/called : load the option file and then go fish properties)

One point that still confuses me is scripts for translation file creation/update. What are they for ?

daid:
In general, you usually put a macro or function in your source code that wraps all English texts.

For example:
```
new GuiButton(this, "START_SERVER", "Start server");
```
Becomes:
```
new GuiButton(this, "START_SERVER", _("Start server"));
```
It's almost the same code, but the translatable string is wrapped in the "underscore" function.

A script/program (usually xgettext) extracts all strings that are wrapped like that and puts them in a localication (.po) file. So you do not need to run the program, or even do any manual work to generate these files.

Then tools like "poedit" can be used to edit this po files to generate the translations.

Fouindor:
Ok, thank you for your patience and the infos. I did not know there was already such a framework for translation management, I know what to read now  :)

Fouindor:
Learned a lot about translations :)
Using libintl (for gettext), xgettext for pot generation and poedit for generation translation, it seems to work fine. However, the problem is that even if I try to change locale with setlocale, gettext always returns the translation of my default system locale, I can't even fall back to default strings unless I delete the translation folder associated with my default system locale. Would you kindly have any clues on that, please ?

daid:
Ah, yes, on linux I've noticed this problem as well.
This snippet from my current work project should help:

```
void setLanguage(const char* language)
{
        //Set the "LANGUAGE" environment variable, this overrules all other language selections, and allows for language selections that are
        // not system wide installed. (setting the language with setlocale does not work, as this checks if there is a system wide language with that name)
        setenv("LANGUAGE", language, 1);
        //We also need to set the LANG environment variable. Else the LANGUAGE setting is ignored.
        // However, the LANG environment variable needs to be a language that is available for libc as well. So we use C.UTF-8, which is always available.
        setenv("LANG", "C.UTF-8", 1);
        setlocale(LC_ALL, "");
        bindtextdomain("jedi-display", "/usr/share/griffin/locale");
        textdomain("jedi-display");
}
```

This is the best way I found that I could forcefully overrule the language. (yes, it has hardcoded paths and names, but that's fine for what I'm working on)

PR #118, opened 2016-01-30
OSX .app bundle

kfprimm:
Hi,

I've got the application built, packaged as an app bundle, and running on El Capitan without crashing.

Here's an overview:

1. Modified `chdir` code in `main` to point to the proper directory.
2. Changed `CMakeLists.txt` so it invokes `BundleUtilities` to generate the bundle structure on OSX.
3. Still investigating retina display issue. 

---

The first issue is pretty-straightforward. [This code](https://github.com/daid/EmptyEpsilon/blob/02c7c110198d0f9abd1f4e9abbfbd058225d33a0/src/main.cpp#L53-L61) is correctly changing the directory to the `EmptyEpsilon.app/Contents/Resources` directory where the `resources`, `packs`, and `scripts` directories should be in the app bundle. If the program is not bundled, nothing happens.

However, it seems that when your project directory name equals the executable name (in this case `EmptyEpsilon`) `CFBundleGetMainBundle` is returning a valid `CFBundleRef`. It seems OSX is guessing that the executable is in a bundle because of that. This is somewhat consistent with other behavior as it will make the same assumption with an `.icns` file in `Contents/Resources`.

I did some quick research, but did not turn up much. Hopefully someone else has experience with this. 

My (hopefully) temporary solution is to check if the executable is in root of the alleged "bundle". If it is, `chdir` to the executable directory, otherwise change to `Contents/Resources`.

---

CMake provides some scripts to make automatically building a bundle straightforward. I did need to override the default `Info.plist` template so I could add `NSHighResolutionCapable` for (future) retina support.

In order to get a complete app bundle, you need to run 'make && make install'. `make` builds the executable, builds the directory structure, and adds `Info.plist`. `make install` copies the data directories to `Contents/Resources`.

So, `CMAKE_INSTALL_PREFIX` is also being overridden to the current directory so `make install` dumps the files in the bundle. I don't know if there is a better way to do, but it was straight forward to implement.

---

Currently, high resolution support is disabled as SFML 2 doesn't seem to support it fully, or at least automatically.

I'll be looking into it further, but I figured I would go ahead and submit a pull request for this so others can build & test on OSX.

Thanks!


daid:
If this works reliably, I'll try to add OSX builds to the official builds as well. Good job!

PR #120, opened 2016-02-01
Addition of translation support.

Fouindor:
Hello,
I wanted to submit a "naive" translation support for menus and GUI : a new option in the menu is present, to use system locale or not. Default is to not use sys. locale, so the game will be in english. If the use of system locale is enabled and a translation is available in your language, then it will use the translation.

Important : for gettext, you need to have libintl.

Also added pgettext (already available on UNIX systems ?), a function recognized by xgettext, which allows to append a context to a translation. This allows for different translations of the same word. e.g. :
"Back" in menu buttons, as in the "return" meaning, has been translated to "Retour"
"Back" in main screen control, as in "back of the ship", has been translated to "Arriére"

I've supplied only the french translation for now. Tested and it seems ok. This solution isn't the most optimized or elegant and has some things missing : as I do not use cmake, I did not update it in this commit. Also as I do use Poedit (and not a standalone version of xgettext) to generate translation files, I do not have created a script to generate .pot files automatically.


Fouindor:
I was worried that this addition would break Linux support, but it is compiling fine on 64-bit Ubuntu (SFML requires GCC >= 4.9 however). All you need is the gettext package.

However, there seems to be oddities in the encoding.

So, for windows you need :
libintl : http://gnuwin32.sourceforge.net/packages/libintl.htm
libiconv2.dll, required by libintl, found in http://gnuwin32.sourceforge.net/packages/libiconv.htm

Fouindor:
As I want to work on something else and as there is file conflict, I will close this for now. Thanks daid for the translation code tips (it will be useful for me anyways :smile: )

PR #124, opened 2016-02-20
Fixed #123

### LINKED ISSUE #123: GM screen crash
Hello, when selecting GM screen, the game crashes

My OS : Windows 7
Version : Latest sources. Also tested with @kwadroke's autobuilds, happens with :
- Latest (0aa7374)
- 85e0058

But does not happens with f69274d, so the GM screeen-breaking commit must be 85e0058.

Link to RPT : http://s000.tinyupload.com/?file_id=67458864984668680142

How to make bug happen : Just select GM console on any scenario

Fouindor:
Arrow-drawing code was not checking if a spaceship was set or not. In the case of GM screen, it tried to access a non-set object, which made the game crash.


Fouindor:
Solved by 831004e

PR #131, opened 2016-03-01
os x support for dmx.

HenriVesala:
Not really apiece of art but it seems to be working.

Fake break parity didin't work with my test setup.
Don't really know anything about this...


daid:
Not everything needs to be art :-) and if it works, that's a start.

Did a small patch. https://github.com/daid/EmptyEpsilon/commit/6b54aaf1df5fdc1492e9297592030d4946b8fc08

I moved the IOSSIOSPEED define into the cpp file, as it is only needed there, and does not need to be in the header file. (The header file just defines the external interface of the SerialPort class, which does not need to know about this define)

And I removed a bit of code duplication by making the Linux and OSX code use the same code where possible.

PR #132, opened 2016-03-01
dmx channels were mapped one channel too high.

HenriVesala:
So 1 in config would become channel 2 in dmx device. 
Not a major problem but quite confusing.


daid:
I had a hard time finding out if DMX was 0 or 1 based. Some sources start counting at 0, some at 1.

So, naturally I started to count as 0, as a good programmer :-)

PR #141, opened 2016-03-23
App Icon Updates

InterestingJohn:
Ok, I think I managed to tackle most of the icon updating, let me know if any of it wasn’t done correctly.


InterestingJohn:
Added the slightly updated starfield as well, it's not perfect yet, but it's _better_ at least.

PR #143, opened 2016-03-23
Updated nav and body styles

InterestingJohn:
Updating styling to look more like the new in-game UI


InterestingJohn:
The vertical tabs under "Mission Scripting" are a bit wonky, but I think this update works, overall.

daid:
Yes, looks a lot better already.

The tabs on the mission scripting could be "fixed" with a smaller font I guess. I never got that CSS to work as I wanted (dynamic scaling those elements to the text size)

PR #161, opened 2016-04-02
Added new shield, hull and self-destruct icons

InterestingJohn:
Let me know if I did any of this wrong, it's my first time digging into the UI code.


daid:
Looks fine, except for a small issue in the self destruct button. But I can fix that :-)

PR #165, opened 2016-04-02
Added a function to broadcast faction messages. AI ships now inform when taking new orders.

Fouindor:
Added SpaceShip::addFactionBroadcast(string message), function is accessible by script.


daid:
The name implies that the message is added to the ships log of the same faction, however, the function itself broadcasts to all friendlies and neutrals as well. Wouldn't it make more sense to have 3 different functions?
addFactionBroadcast
addFriendlyBroadcast
addNeutralBroadcast

Fouindor:
While I see how it is not aptly named, I don't see the usefulness of splitting it in 3 functions. Is it ok if I just rename it addBroadcast ?

nallath:
or even better; Create a function addBroadcast(enum type, string message)

daid:
>While I see how it is not aptly named, I don't see the usefulness of splitting it in 3 functions.
While you might not see the use, someone else might.
For example, the AI order changes, I wouldn't want to broadcast those to neutrals as well. Why does a neutral needs to know this? I want to update my friendlies with what I'm doing.

>Is it ok if I just rename it addBroadcast?
That implies that it is broadcasted to everyone. Which is also not true.

Fouindor:
Ok, no problem, thanks for the clarifications, I'll make the modifications and submit them later today :)

Fouindor:
Changed the function, Broadcast has now 3 thresholds :
- 0/FVF_Friendly : Communicate to allies only
- 1/FVF_Neutral : Communicate to allies and neutral
- 2/FVF_Enemy : Communicate to everyone, including enemies

Played with it a bit, seems to work ok so far, both in program and missions :
http://i.imgur.com/kL2QnNH.png
http://i.imgur.com/x0uQ2Sn.png

PR #178, opened 2016-04-08
Added engineering column icons

Fouindor:
My new players were particulary confused about the new engineering system display. Added icons for legibility. Tested on different resolutions, seems ok.


Fouindor:
http://i.imgur.com/AwUcWe0.png

daid:
Had the same problem with new players yesterday. Hope this will help a bit. I think it would be good to add the same icons to the sliders on the right as well.

daid:
I also need to fix that the sliders on the right now are far from the rest of the stuff in widescreen mode. As gestalts principles say, things that are close together belong together. And these controls belong together but are not close together now.

InterestingJohn:
Why'd the beam weapons get so blurry?

I'm also trying to decide if I like that concept for the reactor or not, _should_ it be different from the "Energy" icon? Aren't those two things intrinsically tied together?

daid:
The icon got a bit blurry because I changed a bit how the icons on the ship where drawn, so icons on buttons and icons on the ship can use the same image files, that match the template you provided.

The reactor one was just something I threw together. It can also be the same as the energy icon. Wasn't sure on it yet.
So that one, the warp/jump drive and maneuvering still need work. And the beams just need a small update to be the proper resolution like the rest.

InterestingJohn:
#180 Should fix the blurry beam icon. :)

I've got a pretty full weekend ahead of me but I'd like to try and really tackle the rest of the stations next week, as well as finish designing any additional icons that need to be made.

I also want to figure out a good solution for the alerts, I tested my circular design and it worked _ok_ on the screens where the radar is in the middle, but screens like engineering, science and relay look odd, so I need to re-think the alert design and we may need to adjust code for it depending on what the solution ends up being. Should I start issues for these things?

daid:
There is no problem in having an alert image per station. Or one for screen with a radar, centered around the radar (I want to move the science radar a bit more to the center anyhow) and one for raderless or "square" screens that looks differently and maybe is more "corner" based.

Code wise it's all no problem. I can also put the image behind the controls instead of on top of them.

InterestingJohn:
Awesome! That'll help inform my designs going forward. Also putting the image behind the controls will open up a TON more options.

I was wondering about the science radar placement, do you think it's worth having different positioning based on screen aspect ratio? Because left alignment makes complete sense on a 4:3 screen, but on a 16:9 or 16:10 screen it could easily be centered like the rest of the stations without issue.

Issue #136, opened 2016-03-09
Request - Advanced Engineering Station


lunerfox:
Right now, the engineering station is able to instantly route power and coolant to any subsystem on the fly, allowing for instant changes to min/max the ship's capabilities. 

Such dramatic responses end up feeling game-y, since realistically, coolant and power plants need time to distribute their resources. Which is why I'm proposing the addition of an **Advanced Engineering** mode.

Similar to the advanced science interface, the engineering station puts limitations on routing power and coolant, rewarding advanced planning and strategy.

All changes to subsystem coolant and power levels is based on set points, which the subsystem will match given some time based on an S-curve.

The Engineer may dedicate additional reactor power and coolant to get faster responses, at the cost of having less coolant to go around.

Fouindor:
This is a really good idea. But why make a S-curve distribution progression ? Wouldn't a linear progression be good enough ? Maybe instead of "advanced engineering", have a "Distribution speed" parameter which can be Slow/Fast/Instant.

lunerfox:
@Fouindor Yes I think an S curve might be unnecessarily complex. I would add that distribution speed should be different for adding power/coolant vs. removing.

It should always take longer to increase power (More power, more resistant to change), typically systems want to

leoszeto:
(Continuing from the last comment)

Typically systems are biased towards the de-energized state, so any setpoints lower then the current setpoint should have a faster response.

It's likely sufficient to just have 2 linear rates (resourceRateUp & resourceRateDown), modified by power level of the reactor (More overall power = Better system response)

Ex: 
resourceRateUp = resourceRateUpBase + resourceRateUpPwr x PwrSysLvl

daid:
On engineering itself, as a bigger picture. Currently, it's mostly a straight copy from how the Artemis system worked.

I have a few ideas that could overhaul the whole way engineering (both system management and repairs) work. But the main issue here is keeping the complexity of engineering in par with the complexity of other stations. Currently it's already the station with the most controls and the most things it needs to pay attention to.

lunerfox:
Completely agree with the design of engineering station being mindful of complexity; Initially I was thinking of the presets Artemis used as a way to reduce it, but that also brings in other issues during play.

Daid, what are some of your thoughts on how engineering station could be overhauled?

daid:
One complete overhaul idea is to make it like a sort of 2D "Space Engineers"/"Big Pharma"
[...]

PR #197, opened 2016-04-18
Made power and coolant distribution not instantaneous per #134

Fouindor:
Distribution speed hardcoded as of now


Fouindor:
Per #136, not 134...

daid:
Looks sane. But wouldn't it make sense to just remove the old commandSetSystemCoolant functions instead of adding new ones?

Fouindor:
They are used in the tutorial. Furthermore, I think it could be handy to keep these options for other purpose (like simulating a system down in a mission).

Fouindor:
Thought a bit, and it indeed makes more sense. Replaced the obsolete commands by new SpaceShip functions, setSystemCoolant and setSystemPower

daid:
Ah, yes, wanted to suggest that :D

kwadroke:
I would like to be able to take systems down/offline in a mission.

daid:
You could just keep forcing the power to 0.

And the tutorial uses some tricks like setting the max impulse speed to 0 to disable systems as well.

kwadroke:
That was my plan until we had a way to take systems down. 
But it would be nice to have an indicator saying it was offline.

Fouindor:
Are the changes ok ?

daid:
Didn't have time to check it earlier, looks good now. See you also fixed some minor other things. which is fine.

Fouindor:
Thanks, sorry if I sounded pushy, I was worried that some issues remained.

PR #216, opened 2016-05-07
ignore the options.ini file, output .bin directory and vim temp files

pryankster:
This is a silly one, feel free to reject, but 'git update' showing files that I know will never be checked in gets my OCD going :-)

--pryankster


daid:
I have about a 100 files that are not ignored and thus show up, but I don't mind patches on the gitignore :-)

PR #218, opened 2016-05-08
Port to Visual Studio 2015

pryankster:
Changes to port to Visual Sudio 2015 (Community Version).   Biggest part of port is moving template specialization definitions to .hpp files.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,6 +1,8 @@
 #include <string.h>
+#ifndef _MSC_VER
 #include <unistd.h>
 #include <sys/stat.h>
+#endif
 #include <sys/types.h>
 #include "gui/mouseRenderer.h"
 #include "gui/debugRenderer.h"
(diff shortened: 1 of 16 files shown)


pryankster:
What was the problem with the pathFind changes?  The algorithm should have been the same, just made the code (IMHO) a little more readable, and removed the dynamic allocation that MSVC doesn't like.

I'm willing to debug and fix, but it looks like it's working fine on my build.

--pryankster

daid:
First off, it did not initialize the data properly, old code uses C style memset to set it all to zero.

After I fixed that I still got random crashes on moving crew members in the new Pirana ship. Not sure why, but I tested the old code, and that worked without a glitch. So I rather reverted it then searching for the bug.

pryankster:
Found the issue(s):  first, I had forgotten that C++ default constructors don't initialize, second was that the map indexing was doing x*width + y instead of y*width + x, and thirdly, the assert was checking y < width instead of y < height.

PR #230, opened 2016-05-10
Pryankster msvc take2

pryankster:
Second take at repairCrew.cpp fix.  This adds the constructor for PathNode, fixes the assert() in PathMap::Node(), and corrects the multiplication (y*width + x) instead of (y*height + x)

Sorry for the bad pull req last time.


daid:
Great, that seems to have fixed it :-)

PR #232, opened 2016-05-11
Use new power/coolant request functions in tutorial

oznogon:


daid:
Don't think you tested this, as the functions are named "commandSetSystemPowerRequest"/"commandSetSystemCoolantRequest"

oznogon:
Updated.

Issue #212, opened 2016-05-04
Headless server playing audio


kwadroke:
When you run Empty Epsilon with the ``headless=scenario_01_waves.lua`` parameter, audio (music) is played, even without starting a client.

In my opinion, the Main Screen client should play the music files, not the headless server.
(I'm wanting to run this as dedicated server for events and Internet play.)

To replicate:
Run ```EmptyEpsilon.exe headless=scenario_01_waves.lua``` in windows
or
```./EmptyEpsilon headless=scenario_01_waves.lua``` on Linux (untested)

oznogon:
Confirmed that the music begins playing on OS X by running `./EmptyEpsilon headless=scenario_01_waves.lua` from `EmptyEpsilon.app/Contents/MacOS`.

daid:
As intended, the server is running the music. Else you would have two sets of music when you have two main screens.

gguillotte:
So the workaround/solution is to turn the music volume to nil on the headless server?

oznogon:
So the workaround/solution is to set the music volume to 0 when running headless?

kwadroke:
So, what about if you are GM'ing a game? Since the GM is has to run the server, I doubt the GM would care about the music.

MyrddinE:
This is actually the issue we are running in to for our online games. The GM hears music... 6 clients hear nothing. While the option makes perfect sense for LAN play, it's not very good for Internet usage. Perhaps a Music button that clients can click? With a musical note on it? And potentially a server option to block clients from seeing the button when on LAN (or conversely, hide it by default and only show it when the server enables it) so you don't get competing music playing when unwanted.

oznogon:
The combat/ambient music switch is triggered on the server by changes to the threat level, which is [determined by and occurs on the server](https://github.com/daid/EmptyEpsilon/blob/68451f9317f377a65c99ab88b3aab20e84807df1/src/epsilonServer.cpp#L18-L23). Those `threat_level_low` and `threat_level_high` functions trigger music playback in the server only, so it assumes the server is always part of the game presentation (probably as the main screen).

Running `headless`, however, means the server itself probably isn't part of the presentation. GMs on the server also might not want music on their station, but instead only the main screen client.

If it's possible for the `threat_level_low` and `_high` functions to trigger a function on screens in addition to the server, and if screens are capable of playing music, it should be possible to play and change music:

[...]

PR #255, opened 2016-05-23
Move music to clients and add option to toggle music

oznogon:
-   Move threat estimation from server to crew screens.
    -   This also moves music playback and switching from server to clients, which resolves at least part of #212 (headless servers playing music).
-   Create `music_enabled` option in `options.ini` and the options screen. Values:
    -   0: Client never plays music during game.
    -   1: Client always plays music during game.
    -   2 (default): Client plays music only when running the main screen.
-   Add music_enabled selector to option menu.

I'm not sure if this introduces issues with Android, which doesn't bundle music.

### DIFF
--- a/src/epsilonServer.cpp
+++ b/src/epsilonServer.cpp
@@ -15,23 +15,12 @@ EpsilonServer::EpsilonServer()
     for(unsigned int n=0; n<factionInfo.size(); n++)
         factionInfo[n]->reset();
 
-    threat_estimate = new ThreatLevelEstimate();
-    threat_estimate->setCallbacks([](){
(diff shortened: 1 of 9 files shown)


daid:
The implementation looks mostly sound. I just have a small problem with the default, the option menu now specifies a default of "2", but if another code path reads it first it does not have this default yet, and thus defaults to an empty string.

If there is no music, (as on android) you can still safely request a music file set, it just does nothing.

oznogon:
Thanks, @daid! I added a commit to have PreferencesManager check for that setting in main.cpp, and set a default if the setting is empty.

PR #262, opened 2016-05-30
Added a change log

OldDragon2A:


daid:
Are you going to keep this updated? Because I'm not.

OldDragon2A:
I was planning too.

nallath:
Well, we can always remove it if it doesn't get updated anymore.

PR #263, opened 2016-05-30
Updated the site

OldDragon2A:
Moved to CDN versions of the javascript and supporting stylesheets.
Fixed the indentation of the code samples and added HighlightJS for syntax coloring.
Added a page for displaying the change log.
Fixed the page forwarding for the forums.


nallath:
Nice work :)

daid:
I actually reverted this, as it visually made the font on my machine very ugly. I also cannot review the changes in the html file as it marks the whole file as changed.

OldDragon2A:
Since I moved one of the font definitions to the main stylesheet, my first
thought would be your browser was using a cached version of stylesheet.

The HTML is mostly the same, just with code blocks rather than styled
paragraphs. The javascript is the majority of actual change.

Issue #261, opened 2016-05-29
BUG - Multiple repair crews can occupy the same position


oznogon:
Multiple repair crews can be in the same position on a ship, effectively stacking themselves into a single spot. This allows engineers to place more repair crews into a room than it should hold. For instance, with this bug it's possible to fit three crew members into the two-position rooms on an Atlantis X23.

To reproduce:

- On the Engineering station, select a repair crew, then click a position in the ship.
- Before the crew reaches its destination, select another crew and click the same position.

Expected result:

- The second crew to arrive either finds an open position in the same room or stops in an empty position outside of that room.

Observed result:

- The second crew stacks on top of the first crew to occupy the same position.

Here's two crews stacked in this manner between the reactor and jump drive rooms.

![crewpos](https://cloud.githubusercontent.com/assets/19192104/15631264/88551bcc-2514-11e6-960f-bcd37dd1114a.png)

kwadroke:
I've seen that problem myself occasionally.

daid:
Yep, can happen. It's more fun if you use the player version of the Flavia (light transport), as it has a lot of repair crew. And a layout that can require quite a bit of walking.

PR #264, opened 2016-05-30
Disallow crew overlap (#261)

Fouindor:
Tested, seems to work OK.


daid:
I've fixed it in a slightly different way, keeping the concept the same. But I check it server side now, and made the function a member of the repairCrew class.

daid:
(Still, thanks for the pointer that this is an easy fix this way :+1: )

PR #267, opened 2016-05-31
Expand top-down controls.

oznogon:
-   Allow for an unlocked camera.
    -   Add keyboard controls for panning and zooming.
    -   Add mouse wheel zoom controls.
-   Add a selector for picking a player ship to lock onto.


daid:
The idea of the topdown view is actually to run it on a different monitor/tv where people can follow the action "outside" of the main game. So I'm not sure how happy I am with controls on top of this view.

I did want to expand it so it zooms/pans automatically to better follow action happening.

oznogon:
Thanks for the feedback, @daid. I've added keyboard controls to toggle the camera lock and cycle through player ships, and I've hidden the GUI elements by default. They can still be revealed with the H key if a spectator wants them.

daid:
Sounds good :)

PR #269, opened 2016-05-31
Add SIL OFL v1.1 for Bebas Neue.

oznogon:
This adds the SIL OFL v1.1 license file, per the [creator's license](http://dharmatype.com/license):

> When you will download Bebas Neue, Bebas Kai , please follow SIL Open Font License 1.1.

Per SIL OFL 1.1:

> Original or Modified Versions of the Font Software may be bundled, redistributed and/or sold with any software, provided that each copy contains the above copyright notice and this license. These can be included either as stand-alone text files, human-readable headers or in the appropriate machine-readable metadata fields within text or binary files as long as those fields can be easily viewed by the user.


daid:
Ah, missed that part. When the font was suggested, I did check if we could use it, but I must have missed this detail. Great for pointing it out!

PR #280, opened 2016-06-04
Log viewer fixes

oznogon:
-   Fix autoplay, and provide a link to toggle it.
-   Add an html DOCTYPE.
-   Cap the negative mousewheel delta value to avoid setting an impossible (0, negative) zoom level.
-   Scale objects to the zoom level. (See #279 for player ship scaling.)
-   Display the logged data on projectiles and beam/missile impacts.
-   Show the zoom scale and current X/Y coordinates next to the time.


daid:
Love it. Just did some work on it to provide a base, I'm glad someone else is also spending some time on it :D

PR #284, opened 2016-06-06
Add option to toggle game log generation

oznogon:
Set `game_logs=1` by default in options.ini. Game logs can be disabled by switching that setting to 0.

Also tweaks description margins on server selection screen.

### DIFF
--- a/src/epsilonServer.cpp
+++ b/src/epsilonServer.cpp
@@ -20,8 +20,7 @@ EpsilonServer::EpsilonServer()
     {
         state_logger = new GameStateLogger();
         state_logger->start();
-    } else if (state_logger)
-        state_logger->destroy();
+    }
 }
 
 EpsilonServer::~EpsilonServer()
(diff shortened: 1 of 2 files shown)


src/epsilonServer.cpp
```diff
@@ -15,8 +16,12 @@ EpsilonServer::EpsilonServer()
...
         factionInfo[n]->reset();
 
-    state_logger = new GameStateLogger();
-    state_logger->start();
+    if (PreferencesManager::get("game_logs", "1").toInt())
+    {
+        state_logger = new GameStateLogger();
+        state_logger->start();
+    } else if (state_logger)
+        state_logger->destroy();
```
daid: The last part about destroying the logger can never trigger. As this is during construction of the server object. The server holds the state_logger when it is created and will destroy the logger as well when the server gets destroyed.

As this is "at creation time", there is never a logger in this member.

oznogon: Yeah, that makes sense. I've removed it.

Should the game state logger start in gameGlobalInfo->startScenario() instead of server creation? If logging starts there, it could be toggled from the server creation screen. It doesn't look like anything's logged until the scenario's started anyway.

daid:
Closing this in favor of #285

PR #285, opened 2016-06-06
Move game state logging from server creation to start of scenario

oznogon:
Game state logging starts when the EpsilonServer object is created, and us destroyed when the object is destroyed. However, the logger doesn't log anything that happens until a scenario starts. This isn't ideal for a few reasons:

-    Starting the server and closing it without starting a scenario results in a 0-length log file left in the logs folder.
-   Running multiple scenarios on the same server object results in all scenarios logged to the same file.

This PR:

-   Moves game state logging into the startScenario function of GameGlobalInfo. The logger starts when the scenario starts and is destroyed when the scenario is destroyed. A single server object should create a unique log file each time it starts a scenario.
-   Sets a `game_logs` Boolean option in PreferencesManager, with a default value of `true`/`1`.

### DIFF
--- a/src/epsilonServer.cpp
+++ b/src/epsilonServer.cpp
@@ -1,7 +1,6 @@
 #include "epsilonServer.h"
 #include "playerInfo.h"
 #include "gameGlobalInfo.h"
-#include "gameStateLogger.h"
 #include "main.h"
 
 EpsilonServer::EpsilonServer()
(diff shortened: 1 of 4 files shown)


daid:
> I'd need to test to confirm, but I also suspect logging doesn't work as expected when running multiple successive scenarios on the same server object.

I can confirm this one, the session of Saturday had about 10 games in a single 100MB log file. So moving it makes sense from that perspective.

I don't think this needs to be a toggle button on the server creation screen. As it does not effect the game directly.

oznogon:
Thanks, @daid. I've removed the server creation screen button from this PR.

PR #289, opened 2016-06-07
Move shields button from system info on Eng+

oznogon:
Prevent the shields toggle on Engineering+ from covering the new system efficiency information added to Engineering.


daid:
Do you sleep/eat? :-)

PR #303, opened 2016-06-10
Ship selection screen improvements

### LINKED ISSUE #194: Suggestion: On Role Selection, Show Counts
It would be far easier to set up ships, avoid multiple people accidentally entering the same view, and coordinate role changes if the Role Selection screen showed the number of users who have selected that role next to the role button, and the ships showed the number of users ready on that ship.

This is the simplest version of the suggestion... there could be enhancements. For example, this version would not show any difference between five people choosing five different roles, and one user choosing five positions (except the ship count wouldn't match the role count). But this would help Internet groups get set up, which is currently far more confusing and tedious than it is in person when you can just hand them a tablet set for Helm.

ee-readycount

oznogon:
-   Resolve #194.

    -   Check which stations are occupied on a player ship, then
        displays a count on the ship and an indicator on the station.

    -   Adds the faction to player ship names to help PvP scenarios.

        Indicators

-   Add public single-entry setter functions to GuiEntryList to
    avoid having to destroy the entire list when changing an
    entry's name or value. (This supports updating the
    ship/position button text.)

-   Convert the ship window angle selector to a slider, because
    the selector can't fit all options on the screen at once.

    Ship window angle slider

-   Add comments to `shipSelectionScreen.cpp`.

### DIFF
--- a/src/gui/gui2_entrylist.cpp
+++ b/src/gui/gui2_entrylist.cpp
@@ -27,6 +27,31 @@ GuiEntryList* GuiEntryList::setOptions(std::vector<string> options, std::vector<
     return this;
 }
 
+void GuiEntryList::setEntryName(int index, string name)
+{
+    if (index < 0 || index >= (int)entries.size())
+        return;
+    entries[index].name = name;
+    entriesChanged();
+}
+
+void GuiEntryList::setEntryValue(int index, string value)
+{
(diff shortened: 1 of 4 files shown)


CaptLapadite:
Well done oznogon ! Much appreciated :+1:

PR #304, opened 2016-06-11
Let Relay put comms on main screen

### LINKED ISSUE #18: Docked communications with SpaceObjects on main screen
When docked there's using no action. We could show the communication on the main screen. Also, it would be nice if a choice needs to be made. Such as investing in ship upgrading. The upgrades could be available at select stations.

A large station could have an alien cloaking device or additional missile bays, or another beam weapon (costing many credits).

A cloaking device should be available only at a pirate trading hub or on a planet, not through official channels. This encourages deeper space exploration instead of just fighting off ingress enemies.

oznogon:
Resolves #18.

-   Add a comms overlay to the main screen, defaulting to hidden.

-   Add MSS commands to hide and show the main screen comms overlay.

-   Add a button to the main screen controls to toggle onscreen comms.
    The button appears only if the player has a station capable of opening comms
    (Relay, Ops, or Single Pilot).

    comms-1

    comms-2

-   Add a function to `commsOverlay` allowing its contents to be manually hidden.
    This works around a potential issue where the contents of the overlay aren't
    hidden when the overlay is hidden; they remain visible but aren't updated.

-   Fixes some button IDs and makes more buttons on the comms overlay variables
    for future customization.

### DIFF
--- a/src/screenComponents/commsOverlay.cpp
+++ b/src/screenComponents/commsOverlay.cpp
@@ -12,11 +12,14 @@
 GuiCommsOverlay::GuiCommsOverlay(GuiContainer* owner)
 : GuiElement(owner, "COMMS_OVERLAY")
 {
+    // Panel for reporting outgoing hails.
     opening_box = new GuiPanel(owner, "COMMS_OPENING_BOX");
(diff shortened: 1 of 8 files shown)


oznogon:
Closing this because I think there's a better way to handle main screen overlays.

PR #310, opened 2016-06-13
Alert overlay smaller

Wrongtown:
Suggest alteration to my last attempt at the Alert Overlay image; I've just removed top and bottom bars then ran it through pngcrush. As usual looking to make things cleaner based on feedback on bridgesim.net but welcome your opinion.


daid:
I'm still not sure what to do with the alert underlay. It's a tough one to get right. But I think less in more in this case.

PR #312, opened 2016-06-14
Reduce width of missile tube controls

oznogon:
Reduce overlap of missile tube controls over radar.

tubes


src/screenComponents/missileTubeControls.cpp
```diff
@@ -142,8 +142,7 @@ void GuiMissileTubeControls::onDraw(sf::RenderTarget& window)
             rows[n].loading_bar->hide();
         }
     }
-    for(int n=my_spaceship->weapon_tube_count; n<max_weapon_tubes; n++)
-        rows[n].layout->hide();
```
daid: This should not have been removed, as this causes the bug Kwadroke noticed.

PR #314, opened 2016-06-14
Restore state change clearing loop.

oznogon:
Fixes the issue introduced in #312.


daid:
Note, I understand how this code could look odd, but it's written like this to allow for dynamic changes of the amount of missile tubes.

PR #313, opened 2016-06-14
Use PNG alpha for radar drop shadow overlay

oznogon:
Replace the opaque background image with a transparent/alpha-shaded overlay. This allows the background color configured in `colors.ini` to appear.

This also adds custom shadows for the Science/Ops screens, which has a larger radar shifted left, and the single-pilot overlay, which has a smaller radar.

PR #315, opened 2016-06-14
Hide the radar background gradient on database view

oznogon:
Resolves the issue brought up by @Fouindor in #313.

-   Assigns the background gradient to a variable across stations so it can be more easily cleared or toggled.
-   Toggle the gradient when toggling the database view from Science.


Fouindor:
Thanks ! :)

PR #322, opened 2016-06-16
Show time left to reach a target [science screen]

elptito:
- when targeting an object, if rel_velocity > 0.01 and direction near the heading (diff < 15°), show the time left to reach the target (format : minutes:secondes)


daid:
I actually intentionally did not provide this information, so science has to do some math in his head from the relative speed.

elptito:
As you want. Personally, I like do math for work, but not (toot much) for a game.

kwadroke:
Make this an optional setting perhaps?

CaptLapadite:
If you wish to add this, it could also appear only to scanned targets, with the info only in science. 
Giving a little something more to science I guess..

kwadroke:
After thinking about this, it might be good for Probes on the Science screen.

elptito:
I don't understand what you mean ? Display the time left to reach a probe ?

kwadroke:
@elptito, No. For the Probe View. 
When you use the Probe View on Science having the time to reach target information might be more valuable since it could be off your regular radar and calculating it could be more difficult or at least take longer.

elptito:
Ha ok. bah, I think that is could be always helpful. Maybe more for probe view indeed.

PR #324, opened 2016-06-17
Add CPU tugs

oznogon:
-   Add AI_Pickup, which instructs a CPU ship to move to a target and trigger the target's docking request. This allows AI tugs and carriers to pick up idle, immobile, unmanned, or otherwise disabled ships.
-   Add Small Tug and Small Jump Tug support frigates.
-   Add a tow request to the station comms script, which dispatches a Small Tug to pick up the requesting player.
-   Add a fly-towards request to the ship comms script.
-   Add basic getter functions for whether artifacts and supply drops can be picked up.

### DIFF
--- a/scripts/comms_ship.lua
+++ b/scripts/comms_ship.lua
@@ -37,6 +37,21 @@ function friendlyComms(comms_data)
 			end
 		end
 	end)
+	addCommsReply("Go to a waypoint", function()
+		if player:getWaypointCount() == 0 then
+			setCommsMessage("No waypoints set. Please set a waypoint first.");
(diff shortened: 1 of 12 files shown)


daid:
I don't know. There is no reason this behavior isn't possible already with just some scenario scripting. It could also mess up certain missions quite badly if you can suddenly request other ships to move other idle ships.

I think you want to "simulate" a tractor beam with this AI function, but I kinda fear it will cause problems.

Fouindor:
I think this is a very good suggestion, maybe just disable the station towing service by default.

PR #325, opened 2016-06-17
Add toggle for sharing energy with docked ships

oznogon:
Let ship templates set whether to distribute energy to other ships
docked with it (enabled by default; no change compared to current
behavior).

For instance, to disable energy sharing with docked ships, add this
to a ship's template:

    template:setSharesEnergy(false)


Fouindor:
Good idea !

daid:
I took your patch and enhanced a bit on it:
https://github.com/daid/EmptyEpsilon/commit/9a4c00f4b4110f31c3c365900845692efc889c33

I did rename the function to setSharesEnergyWithDocked.
I also added a setRepairDocked, so you can allow or disallow ships to repairs docked ships.
And, I've moved this all to the "ShipTemplateBasedObject", so it effects stations as well. So you can now make "dead" stations, that do not repair or give off energy.

PR #328, opened 2016-06-21
CPU ships now recharge missile when docked to station.

Fouindor:
For upcoming AI changes


daid:
Good plan :-) long standing todo in the AI code

PR #329, opened 2016-06-21
Add a Intern Log for the engineering station

elptito:
I propose to add a intern log only seen by the engeneer, to display all intern modifications of the ship. I use it to inform the ship about a virus attack, or some ship failures. I add this function : PlayerSpaceship::addToShipLogIntern(string message, sf::Color color) to add this kind of informations.

Some automatic informations fill the log : 

- shields enabled or disabled
- Missile fire
- Change of the shields frequency
- Change of the beam system target
- change of the beam frequence
- warp speed
- jump initialisation
- damage received
- red or yellow alert
- docking situation
- auto destruction situation

I know there is conflict with the base. According to me, this is only about color at the beginning of the log. I am not familiar with git, but if you are interested by this PR, I can check to remove this bug.


src/spaceObjects/playerSpaceship.h
```diff
@@ -59,6 +59,20 @@ class PlayerSpaceship : public SpaceShip
         bool operator!=(const ShipLogEntry& e) { return prefix != e.prefix || text != e.text || color != e.color; }
     };
 
+    class ShipLogInternEntry
```
daid: Isn't this a 1 on 1 copy of the ShipLogEntry?

src/spaceObjects/playerSpaceship.cpp
```diff
@@ -1138,6 +1209,11 @@ void PlayerSpaceship::onReceiveClientCommand(int32_t client_id, sf::Packet& pack
     case CMD_SET_ALERT_LEVEL:
         {
             packet >> alert_level;
+            string alert_level_string = alertLevelToString(alert_level);
```
daid: First convert the enum to a string and then check the string? Why not check the enum directly?

src/spaceObjects/playerSpaceship.h
```diff
@@ -81,6 +97,7 @@ class PlayerSpaceship : public SpaceShip
     CommsScriptInterface comms_script_interface;  //Server only
 
     std::vector<ShipLogEntry> ships_log;
+    std::vector<ShipLogInternEntry> ships_logIntern;
```
daid: Inconsistent naming, "ships_log_intern" would be the proper name.

src/screenComponents/shipsLogInternControl.cpp
```diff
@@ -0,0 +1,85 @@
+#include "playerInfo.h"
+#include "spaceObjects/playerSpaceship.h"
+#include "shipsLogInternControl.h"
+
+#include "gui/gui2_panel.h"
+#include "gui/gui2_advancedscrolltext.h"
+
+ShipsLogIntern::ShipsLogIntern(GuiContainer* owner)
```
daid: This looks like a copy&paste from the normal ships log. It's better to have a single class that can serve both, instead of copy&pasting.

src/spaceObjects/playerSpaceship.h
elptito: Yes, only the names changed (see my other remark below)

src/spaceObjects/playerSpaceship.cpp
elptito: Yes, I didn't think about that.
I will change it tonight

src/spaceObjects/playerSpaceship.h
elptito: ok sorry

src/screenComponents/shipsLogInternControl.cpp
elptito: It is a copy &paste indeed. I don't know now how to do a single class but I will learn !

daid:
Note, I put the comments there not to criticize, but in the hope you are willing to learn. Overall, it is a good idea :-)

elptito:
Don't worry, I didn't think that you criticized :) . And I want to learn more so it's perfect, thanks for the remarks, I will make a new PR maybe this night.

Issue #323, opened 2016-06-17
SUGGESTION - NEW SOUND


AlexBail:
Hi there, 

I just made some sound for Empty Epsilon on my free time:

https://www.wetransfer.com/downloads/a9a15420c5c6c297a80a98231623210c20160617021239/349f89

Here is the list for now:

button
explosion
Laser
missile launch
shield down
shield up
*Vocal 1 to 10
*Vocal Incoming Transmission
*Vocal nuclear Missile Launched
*Vocal Probe Launched
*Vocal Red Alert
*Vocal Self Destruction
*Vocal Ship Docked
*Vocal Ship Undocked
*Vocal System Overheat

*those still don't have code to trigger the sound

I'm planning to make new sound for:
Nuclear Explosion
Red and Yellow Alert
Something to add with the Vocal Ship docked/undocked

And I'm currently working on ambiance for the space ship. Still it'll probably need more coding if we don't want it to be too much static

I'm working in collaboration with some of my friend on an event called "Dussel 3". and we would need some help to add some code for the remaining sound and the others to come.

Thanks
Alex.

nallath:
:O Niice!

oznogon:
To surface some comments on the now-closed PR #359:

@daid:

> I am thinking about actually adding some generic server->client call that can say "play sound X only crew position(s) Y for ship Z"

@lunerfox:

> The more generic implementation will also give way to adding location specific sounds; since a setup may not always want all sounds from the same location; For example, engine sounds might be better if played from behind the bridge; red alert sounds might also be location specific...etc

@oznogon:

> There are three classes of sounds to consider: ship, station, and feedback.
>
> Ship sounds are what's audible anywhere on the player ship--engine and shield hum, alarms, weapons fire and impacts. In a LAN game, the best place for these are on the main screen. In an Internet game, each station on a ship should play these.
>
> Station sounds are primarily audible on a specific station. We don't have any of these yet, but these would be things like target locks, scans, turret motion, comms chatter, or repair work. In a LAN game, these should play only on the appropriate station, where they wind up being the bridge background noise when played together. In an Internet game, they should be most audible on the appropriate station and backgrounded or not played on other stations.
>
> Feedback sounds are responses to UI interactions, like button presses, and should play only on the stations. Unlike ship and station sounds, these don't need to interact with the server.

[...]

PR #330, opened 2016-06-22
Added sounds by @AlexBail 

Fouindor:
Shields and self destruct, see #323


AlexBail:
There's already a sound for shield up and down. It would be cool maybe if we had different sound when we're hit with shields on and directly on the hull. 

I can work on these sounds, but it'll need some coding that i can't do

Self Destruction could be nice too. I add it to my list!

daid:
Ah, yes, hadn't got around to do this. I usually do EE stuff in the train with sound off...

PR #332, opened 2016-06-22
Implement turreted beams

### LINKED ISSUE #227: [Request] Lasers as turreted systems
As a separate but related idea in #226 , Laser beams could be designed as being turreted. 

As it stands, an enemy ship is either Inside the firing arc, or outside. This leaves little room to do any sort of "Evasive Maneuvers" , where a ship is inside the firing arc, but actively avoiding being shot at.

My suggestion is to think of the firing Arc as the possible range and angle that a gun is **able** to hit, but the turret needs time to actually bring the guns to bear. The actual angle that a gun is pointed at can be shown via a line, similar to the way the missile trajectories are shown.

When an enemy ship is targeted, all turrets will attempt to turn towards the heading. as long as the line passes through the ship, the gun is trained on the ship and can fire normally.

Not sure if firing lines from other ships should be visualized, or if so, how to visualize them.

Each gun can have a default turret turn speed, but should be modifiable via script (Large Capital ships should have big guns that take a long time to turn, while fighters should have turrets that turn almost instantly.

oznogon:
Resolves #227.

-   Add variables, getters, and setters for turreted beams.
-   Add beam rotation mechanics.
-   Draw turret bounds on short-range radar.
-   Implement configurable turret arc, direction, and rotation rate.
-   Implement GM screen tweak UI sliders.

To implement, add `setBeamWeaponTurret(<INDEX>, <ARC>, <DIRECTION>, <ROTATION_RATE>)` to a shipTemplate, with the index of the beam to make turreted.

For example, this creates two 10-degree beams, one on each side of a ship, within 90-degree turret arcs that rotate at 0.25 degrees per tick (~15 deg/sec):

``` Lua
variation:setBeamWeapon(0, 10, 270, 1500.0, 6.0, 8)
variation:setBeamWeapon(1, 10,  90, 1500.0, 6.0, 8)

--                          Idx, Arc, Dir, Rotation
variation:setBeamWeaponTurret(0, 90, 270, 0.25)
variation:setBeamWeaponTurret(1, 90, 90, 0.25)
```

turret

### DIFF
--- a/src/beamTemplate.cpp
+++ b/src/beamTemplate.cpp
@@ -1,9 +1,12 @@
+#include <iostream>
 #include "beamTemplate.h"
 
 BeamTemplate::BeamTemplate()
 {
     direction = 0;
     range = 0;
+    turret_direction = 0;
+    turret_rotation_rate = 0;
     cycle_time = 0;
     damage = 0;
     beam_texture = "beam_orange.png";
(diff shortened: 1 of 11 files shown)


daid:
That is really really cool :)

oznogon:
One of the neat side-effects I noticed in testing is that with a sufficiently fast rotation rate (which is capped at 25 deg/tick), a narrow-arced beam in a wide turret arc effectively behaves the same as a wide-arced normal beam, but the moving beam arc provides more visual feedback showing what the beam's firing at, both of your own ship and other friendly or deep-scanned ships. The beam rotation also shows up in the game logs.

Also, it shouldn't be difficult to also degrade turret rotation rates when the beam system is damaged, or consume energy or generate heat when the turret rotates.

Wrongtown:
Can't wait to check this out, talk about motivation to streamline my build process at home ;)

elptito:
I have a crash when I use the Odin, when I try to display the beam range : 

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Abandon (core dumped)
```

oznogon:
^ moved to an issue for easier tracking

PR #334, opened 2016-06-23
Added new AI repair state

Fouindor:
I've added a new AI state : Repair. When a ship enters in repair mode, it set its previous order aside, tries to find a friendly station in range dock to it, and resume its previous orders when it has full hull and missiles.

An exemple is added in Roaming mode, when the ship has neither missiles nor beam (ok, now a ship will switch to repair mode only when they have no beams and no missiles)

There is still work to do on when to switch to repair mode (when hull is very damaged, ...).

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -428,45 +428,44 @@ void ShipAI::runOrders()
         }
         break;
 
-        case AI_Repair:            //Repair at the nearest station
-        if (owner->docking_state == DS_NotDocking)
+    case AI_Repair:            //Repair at the nearest station
+        P<SpaceStation> supply_station;
+        bool missiles_full = 1;
+        for(int n=0; n<MW_Count; n++)
         {
-            P<SpaceStation> supply_station;
-
-            if (owner->hasJumpDrive() || owner->hasWarpDrive())
-                supply_station = searchFriendlyStation(100000.0);
-            else
-                supply_station = searchFriendlyStation(15000.0);
-
-            if (supply_station)
+            if  (owner->weapon_storage[n] != owner->weapon_storage_max[n])
             {
-                float dist = sf::length(owner->getPosition() - supply_station->getPosition());
-                if (dist < 950 + supply_station->getRadius())
-                {
-                    owner->requestDock(supply_station);
-                }else{
-                    flyTowards(supply_station->getPosition());
-                }
-            }
-            else
-            {
-                owner->orderResume();  //No friendly station in range to dock, revert orders.
+                missiles_full = 0;
             }
         }
-        else if (owner->docking_state == DS_Docked)
-        {
-            bool missiles_full = 1;
-            for(int n=0; n<MW_Count; n++)
-            {
-                if  (owner->weapon_storage[n] != owner->weapon_storage_max[n])
-                {
-                    missiles_full = 0;
-                }
-            }
 
-            if (missiles_full && owner->getHull() == owner->getHullMax())
+        if (missiles_full && owner->getHull() == owner->getHullMax())
+        {
+            owner->orderResume();
+        }
+        else
+        {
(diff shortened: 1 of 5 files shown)


Fouindor:
http://i.imgur.com/yDM9pb3.png

http://i.imgur.com/eS4f2nX.png

src/ai/ai.cpp
```diff
@@ -413,6 +423,44 @@ void ShipAI::runOrders()
             owner->orderRoaming();  //Nothing to dock, just fall back to roaming.
         }
         break;
+
+        case AI_Repair:            //Repair at the nearest station
+        if (owner->docking_state == DS_NotDocking)
+        {
+            P<SpaceStation> friendly_station = searchFriendlyStation(15000.0);
```
oznogon: Would it be a good idea to tie the range to the server's long-range radar setting and/or the ship's jump range, instead of a fixed 15U?

src/spaceObjects/cpuShip.cpp
```diff
@@ -213,6 +214,20 @@ void CpuShip::orderDock(P<SpaceObject> object)
...
+    if (orders != AI_Repair)
+    {
+        previous_orders = orders;
+        orders = AI_Repair;
+    }
+}
+
+void CpuShip::orderResume()
+{
+    orders = previous_orders;
```
oznogon: There might need to be a fallback here in case the previous orders are no longer relevant (such as defending a ship or object that no longer exists).

src/ai/ai.cpp
```diff
@@ -346,7 +348,15 @@ void ShipAI::runOrders()
                 flyTowards(owner->getOrderTargetLocation());
             }
         }else{
```
oznogon: Will unarmed ships with roaming orders within 15U of a friendly station always broadcast that they're out of ammo and go into Repair mode? If so, this statement might need a narrower condition.

oznogon:
I like this concept! It gives ships an option to retreat, which they don't currently have unless it's explicitly scripted.

It might be a good idea to provide some sort of flag or function so scripts can override this AI's default conditions, say to have a ship retreat to a specific station or ship, or to not seek repairs and instead "fight to the death".

src/ai/ai.cpp
Fouindor: Yup, it is a tricky situation. In the current code (before this commit), an unarmed vessel going in Roaming mode will do nothing. Roaming mode action conditions and prioritization should be changed. (I also want to make very damaged ships seek repairs after battle).

src/ai/ai.cpp
Fouindor: Yup, if a ship has either a jump or warp core, it should be able to seek further, like 100U.

src/ai/ai.cpp
Fouindor: I've encountered this case, Roaming mode action condition and prioritization should be changed.

Fouindor:
I'll try to get a commit fixing these points, thanks for the feedback.

src/ai/ai.cpp
Fouindor: Corrected.

src/spaceObjects/cpuShip.cpp
Fouindor: From what I see, all AI orders involving objects have a failsafe, if they target a non-existing object, they fall back to the Roaming AI. So I don't think it's necessary. I will test that later and see if any there is any funky behavior to be sure.

daid:
I don't think "repair/restock" should be an order. It's more something the AI decides to do and more of an AI state.

Fouindor:
Ok, reverted.

Fouindor:
I realize that I may have misinterpreted what should be fixed. Should Repair be the separate AI_Repair order or should the repairs should be handled directly  by each AI order?

Issue #344, opened 2016-06-25
Energy spent by Shields not display


elptito:
In the engeneer station, if the shields are on, the energy decreases but the energy variation is not known : 

![image](https://cloud.githubusercontent.com/assets/13868149/16359268/df6ae9b2-3b2d-11e6-9a2a-82d5b9c31561.png)

I could be great to have this information to adapt the power of the reactor.

oznogon:
For reference, active shields use 1.5 energy/second per [playerSpaceship.h line 34, constant `energy_shield_use_per_second`](https://github.com/daid/EmptyEpsilon/blob/master/src/spaceObjects/playerSpaceship.h#L34).

oznogon:
So this is a little more involved than expected... active shields draw power separately from the function that calculates power draw (and also provides the number to Engineering). In other words, the energy draw of active shields isn't factored into a ship's total energy draw in the same manner as other systems.

Further complicating things is the fact that shield energy draw is calculated differently from system energy draw, namely that system draw is reduced to 1/12.5 before being subtracted, but shield energy draw isn't.

#347 works around this by adding shield energy draw directly to the Engineering screen value updates, but now I'm not sure what the expected or desired behavior of active shield energy draw should be. It definitely isn't 1.5/second, despite the constant's name, but also the value on the Engineering screen doesn't represent power draw per second, either. A value of 1.0 doesn't result in gaining 1 energy per second.

oznogon:
Opened #350 to fix this closer to the root of the issue.

PR #347, opened 2016-06-26
Factor shields into engineering energy usage indicator.

oznogon:
Workaround for #344. Does so in the Engineering screen calculation rather than in getNetPowerUsage, which might be preferable.

### DIFF
--- a/src/screens/crew6/engineeringScreen.cpp
+++ b/src/screens/crew6/engineeringScreen.cpp
@@ -134,7 +134,7 @@ void EngineeringScreen::onDraw(sf::RenderTarget& window)
 {
     if (my_spaceship)
     {
-        energy_display->setValue(string(int(my_spaceship->energy_level)) + " (" + string(my_spaceship->getNetPowerUsage()) + ")");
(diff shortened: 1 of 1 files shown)


daid:
Where does the 12.5 come from? Seems like an odd number...

oznogon:
playerSpaceship's update function [L353](https://github.com/daid/EmptyEpsilon/blob/master/src/spaceObjects/playerSpaceship.cpp#L353):

``` cpp
        if (shields_active)
            useEnergy(delta * energy_shield_use_per_second);

        energy_level += delta * getNetPowerUsage() * 0.08;
```

The update function reduces the result of getNetPowerUsage to 8% (1/12.5) when modifying the energy level for systems energy draw, but doesn't reduce energy_shield_use_per_second by the same amount when assessing active shield energy draw. In other words, the value of energy_shield_use_per_second has 12.5 times more weight than the value of getNetPowerUsage().

The Engineering screen uses getNetPowerUsage() to get the value that it displays. To line the shields' actual energy use up with the system energy draw, I multiplied the value of energy_shield_use_per_second by 12.5.

oznogon:
Closing in favor of moving the shield energy draw into getNetPowerUsage().

PR #350, opened 2016-06-28
Move active shield energy draw to getNetPowerUsage()

### LINKED ISSUE #344: Energy spent by Shields not display

oznogon:
Resolves #344.

-   Increases `energy_shield_use_per_second` from 1.5 to 18.75 (1.5 * 12.5) to compensate for net power usage being reduced to 1/12.5 during energy draw assessment.
-   Reduces net power in getNetPowerUsage() instead of drawing via useEnergy(). This shouldn't affect gameplay since there were no consequences of useEnergy() failing.
-   Updates Engineering screen to display the power draw interval in energy used per second, instead of per 1/12.5 of a second.

I'm not sure whether [L411](https://github.com/daid/EmptyEpsilon/blob/master/src/spaceObjects/playerSpaceship.cpp#L411) (calculating whether to reset warp factor due to lack of energy) needs to be modified. It doesn't use the `energy_shield_use_per_second` constant, just an inline `1.5`.

### DIFF
--- a/src/screens/crew6/engineeringScreen.cpp
+++ b/src/screens/crew6/engineeringScreen.cpp
@@ -134,7 +134,8 @@ void EngineeringScreen::onDraw(sf::RenderTarget& window)
 {
     if (my_spaceship)
     {
-        energy_display->setValue(string(int(my_spaceship->energy_level)) + " (" + string(my_spaceship->getNetPowerUsage()) + ")");
+        // getNetPowerUsage returns energy use per 1/12.5 (0.08) of a second.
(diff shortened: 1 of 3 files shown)


daid:
L411 is just that jumping with shields up uses 150% energy compared to a jump without shields, for the energy used with the jump.
Note, I've fixed this in a different way, I actually calculate the real delta now in the engineering station by looking at previous and current values. This will then also account for energy used by other things like jumps, beams, warp.

Issue #15, opened 2015-02-20
Building on Linux VM


ErikDeBruijn:
Tried building on clean Ubuntu 14.10 within a virtualbox VM.
```
mkdir Dev ; cd Dev
sudo apt-get install libsfml-dev git build-essential codeblocks
git clone https://github.com/daid/EmptyEpsilon.git
cd Emp*
python make_cbp.py
./EmptyEpsilon 
libGL error: pci id for fd 7: 80ee:beef, driver (null)
OpenGL Warning: Assertion failed: thread && thread->netServer.conn && thread->netServer.conn->type==CR_VBOXHGCM && thread->bInjectThread, file /mnt/tinderbox/add-4.3/src/VBox/Additions/common/crOpenGL/pack/packspu_misc.c, line 562
Segmentation fault (core dumped)
```

I upgraded to VirtualBox 4.3.22, installed guest pack and disabled 3D accelleration. This allowed me to run it, although it's fairly slow when displaying the main screen.

PR #358, opened 2016-07-01
Add looping engine sound, use stopSound when dropping shields

### LINKED ISSUE #274: REQUEST - Loopable sound effects
Can sound effects be looped? It looks like only the music channel loops.

If SFX could be looped, that would open up options for engine noise and shield hum, repeating alarm klaxon sounds, and ambient system sounds. Engine and shield feedback in particular would help provide some passive feedback about when those systems are active.

oznogon:
Requires [SeriousProton PR #15](https://github.com/daid/SeriousProton/pull/15), which would resolve #274.

-   Play a looping engine sound when impulse request is not 0. `engine.wav` is white noise generated using [sox](http://sox.sourceforge.net/):

    ```
    sox -b16 -c2 -r 44100 -n engine.wav synth 5 whitenoise band -n 100 24 band -n 300 100 gain +20
    ```

-   Stop the sound effect to raise shields when they're lowered, to prevent playback from overlapping.


oznogon:
Needs work, closing.

PR #359, opened 2016-07-01
Add looping engine sound

### LINKED ISSUE #274: REQUEST - Loopable sound effects

oznogon:
Requires [SeriousProton PR #15](https://github.com/daid/SeriousProton/pull/15), which would resolve #274.

-   When the impulse system is working, loop an engine sound on the main screen. `engine.wav` is white noise generated using [sox](http://sox.sourceforge.net/):

    ```
    sox -b16 -c2 -r 44100 -n engine.wav synth 5 whitenoise band -n 100 24 band -n 300 100 gain +20
    ```

    ... then edited in Audacity to fade a few frames at each end to prevent popping.

-   Scale the engine sound's pitch and volume to the engine's current throttle, energy allocation, and damage status. The engine roars at 300% power and full throttle compared to a low hum at idle and a quiet hum at underpowered idle.

-   Add a public `impulse_sound_file` string to `shipTemplateBasedObject` for future ship-specific engine sounds.

-   Stop the shields-up sound if the shields are dropped before the sound finishes playing.


daid:
Sounds are still a bit of a mess. As the shield sounds now run on the server. While the engine sounds run on the main screen (which I think is perfect)

I did add a small patch to SP, to prevent stopSound being called with an incorrect index. Saves you -1 checks before stopSound calls, and always allows the output from playSound to be put in stopSound

kwadroke:
How would a 2nd (3rd,4th,etc) bridge get the shield sounds if they are played on the server? Shouldn't all ship sounds be played on the main screen? Not counting the sounds played on stations, of course.

daid:
Like I said, sounds are a bit of a mess. They should be played on the main screen, but are currently played on the server for this bit.

oznogon:
Yeah, it's not as easy to move the shield sounds to the main screen. The main screen doesn't have an easy way of knowing _when_ the shields are raised or lowered, only whether the shields are active at a given moment.

While there's a potential workaround in checking whether shields are active every tick and comparing that state to the previous tick, it'd be more efficient for the server or shields toggle button to simply send the player ship's main screen client the same command.

daid:
I am thinking about actually adding some generic server->client call that can say "play sound X only crew position(s) Y for ship Z"

lunerfox:
The more generic implementation will also give way to adding location specific sounds; since a setup may not always want all sounds from the same location; For example, engine sounds might be better if played from behind the bridge; red alert sounds might also be location specific...etc

oznogon:
There are three classes of sounds to consider: ship, station, and feedback.

Ship sounds are what's audible anywhere on the player ship--engine and shield hum, alarms, weapons fire and impacts. In a LAN game, the best place for these are on the main screen. In an Internet game, each station on a ship should play these.

Station sounds are primarily audible on a specific station. We don't have any of these yet, but these would be things like target locks, scans, turret motion, comms chatter, or repair work. In a LAN game, these should play only on the appropriate station, where they wind up being the bridge background noise when played together. In an Internet game, they should be most audible on the appropriate station and backgrounded or not played on other stations.

Feedback sounds are responses to UI interactions, like button presses, and should play only on the stations. Unlike ship and station sounds, these don't need to interact with the server.

PR #382, opened 2016-08-07
Cursory, basic, but we-still-found-them-useful build instructions

doismellburning:


daid:
Actually: https://github.com/daid/EmptyEpsilon/wiki/Build-from-sources

PR #388, opened 2016-08-18
Can not hack friendlies or unidentified ships

Fouindor:


daid:
I think the first check can be simplified with "isFriendly(other)"

Fouindor:
Thanks ! :)
(I may be a little bit rusty ^^)

PR #407, opened 2016-11-09
Added get/setEnergy script functions

vslotman:
I added getEnergy and setEnergy script-functions


daid:
It's now in conflict with the other merge I already did.

Also, ships no longer have a fixed 1000 max energy, Ships can have a customized max, see setEnergyStorage in the ship template.

vslotman:
Fixed! :)

PR #408, opened 2016-11-09
Give CMake error if DRMINGW_ROOT not set if CRASH_LOGGER is enabled

vslotman:
I added some stuff to the CMakeLists to give an error when the DrMingW root-dir is not found.


vslotman:
@ii8: You're right, it's rather useless on e.g. Linux... Making a new pull-request since this one is already merged.

Issue #417, opened 2016-12-13
Debian Package


paddatrapper:
I would like to package EmptyEpsilon for Debian, however this is made difficult by the fact that it is spilt over two repos. Is it possible to add SeriousProton as a git submodule in EmptyEpsilon? It would also simplify building from source, as the path to SeriousProton can be set in CMakeLists.txt and not necessarily passed as a argument:

    $ git clone https://github.com/daid/EmptyEpsilon
    $ cd EmptyEpsilon
    $ git submodule update --init
    $ mkdir _build && cd _build
    $ cmake ..
    $ make

paddatrapper:
Packaged as library and game

PR #418, opened 2016-12-13
Added SeriousProton as submodule

paddatrapper:
Fix for Issue #417. Still supports setting the SeriousProton path in the command line, but if not set it defaults to `${CMAKE_SOURCE_DIR}/SeriousProton/`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "SeriousProton"]
+	path = SeriousProton
+	url = https://git@github.com/daid/SeriousProton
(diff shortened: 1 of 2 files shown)


daid:
I seriously hate git's submodules system. Sorry.

paddatrapper:
How is it a bad system? It is certainly better than having to manually clone two separate repos and point one to the other...

daid:
Why is it better? It's different, but not better. It still requires extra commands to get/update the code. And it requires more work from my side to update the submodule reference all the time.

Of all the good things that git added, submodules is not one of them IMHO, as it just adds an extra layer of commands/arguments that you need to use and another layer for mistakes when using it, without providing something that you cannot do without submodules.

paddatrapper:
It is better in that it allows a tagged release to point to a specific version of the submodule which is known to work with it. Though the way I'm busy trying to package EmptyEpsilon makes this redundant - shared library for SeriousProton which EE then builds against

PR #424, opened 2017-02-04
Add docking permission

greatestKeks:
I added docking permission dialog to Relay.

### DIFF
--- a/src/spaceObjects/spaceship.cpp
+++ b/src/spaceObjects/spaceship.cpp
@@ -5,6 +5,7 @@
 #include "spaceObjects/beamEffect.h"
 #include "factionInfo.h"
 #include "spaceObjects/explosionEffect.h"
+#include "spaceObjects/spaceStation.h"
 #include "particleEffect.h"
 #include "spaceObjects/warpJammer.h"
 #include "gameGlobalInfo.h"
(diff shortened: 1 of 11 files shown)


daid:
I like the general idea of docking permission, But I'm not sure if this is the best way. Maybe the docking button simply shouldn't work if you do not have permission?

greatestKeks:
This was a possibility I considered. Another option is to increase the multiplier for buying equipment at the station.
I chose to take away 'reputation' because it seems to be fitting and might add a bit of fun.

greatestKeks:
I forgot to ask: Do you want me to implement some changes in my code?

JordanLongstaff:
I tried implementing this, and for some reason, instead of saying "Permission granted" and adding my player ship to the permission list, it simply shows "?" and does not add the player ship.

JordanLongstaff:
I'd suggest increasing the reputation penalty. 10 isn't a whole lot.

src/spaceObjects/spaceship.cpp
```diff
@@ -706,6 +707,17 @@ void SpaceShip::collide(Collisionable* other, float force)
             docking_offset = sf::rotateVector(getPosition() - other->getPosition(), -other->getRotation());
             float length = sf::length(docking_offset);
             docking_offset = docking_offset / length * (length + 2.0f);
+            // if Player has no permission to dock give a penalty of -10 reputation
+            P<SpaceStation> station = dock_object;
+            if (station && !station->hasDockingPermission(this)) {
+                P<PlayerSpaceship> player = P<SpaceShip>(this);
+                player->removeReputationPoints(10);
```
amir-arad: need to check `if(player)`
it might be some bad AI trying to dock a cpu ship, then the game will crash

JordanLongstaff:
I think it would be wise to, instead of forcing the player to get permission to dock or suffer a reputation penalty in every game, make it dependent on the scenario. Expose functions to Lua scripting that allow scenarios to specify whether docking permissions should be required or not.

FaramosCZ:
> I like the general idea of docking permission, But I'm not sure if this is the best way. Maybe the docking button simply shouldn't work if you do not have permission?

It would be better to vizualize the fact that the permission was not granted.
For eamlpe the "Request Dock" button would be overwited by red or organge text "docking denied", when in dock range of a station which disallowed the dock.

Issue may arise when two stations are next to each other, only one denied the docking and the player ship is between them.
In such case, the enabled button "Request Dock" should IMHO get priority and the players would dock to which-ever-one is available.

It migh be also very useful for players to know from faraway (sicence? relay?) if a station denies docking for them.

daid:

PR #425, opened 2017-02-04
Add ship info to science

foesi:
Make a ship description only visible when scanned. Also added another description property to ships which only appears if the ship is deep scanned. This is currently only possible for ships. Maybe this should be extended to ships and stations. For other objects it is maybe worthwhile that the description can be seen all the time.

### DIFF
--- a/src/menus/autoConnectScreen.cpp
+++ b/src/menus/autoConnectScreen.cpp
@@ -39,7 +39,7 @@ AutoConnectScreen::AutoConnectScreen(ECrewPosition crew_position, bool control_m
         LOG(INFO) << "Auto connect filter: " << key << " = " << ship_filters[key];
     }
 
-    if (!PreferencesManager::get("instance_name").compare(""))
+    if (PreferencesManager::get("instance_name") != "")
     {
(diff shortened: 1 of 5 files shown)


czenker:
The feature of this pull requests was implemented in a recent version: https://github.com/daid/EmptyEpsilon/blob/master/CHANGELOG.md#2017-11-03

So this might be closed I guess.

src/menus/autoConnectScreen.cpp
```diff
@@ -124,7 +124,7 @@ bool AutoConnectScreen::isValidShip(int index)
 
     for(auto it : ship_filters)
     {
-        if (it.first == "solo")
+        if (it.first.compare("solo"))
```
daid: Why did you change this? And even more importantly, it's broken now.

foesi:
Sorry. I did a mistake. In Java there is a difference between equals and == and I did not look it up for c++.
I reverted it so now at least this problem should no longer exist.

foesi:
@czenker said this was already implemented.

PR #432, opened 2017-02-14
fix joystick bug

### LINKED ISSUE #426: Bug - Joystick control allows to recharge combat meneuver and lower the heat
Currently, you can use a Joystick to use the CombatManeuverBoost. Not only forwards, but also backwards - which is pretty cool. But as the code does not expect negative values, this leads to a strange and exploitable effect. So by moving backwards, you can not only recharge the combat manoeuvrer bar, but even cool down the maneuvering system instead of heating it up.

There would be two ways to fix this:

**1) Removing the ability to make a reverse combat maneuver boost, consistent with the mouse/touch controls by removing the following lines:**

helmsScreen.cpp, line 78+79
tacticalScreen.cpp, line 75+76

which in both cases looks like:

```
//else
//                    value = (y_position+20) * 1.25 / 100;
```

singlePilotScreen does not seem to have joystick support at all (btw, is there a specific reason?), so nothing to change here.

**2) The ability to boost backwards per joystick is something you would probably expect and should not be too powerful, as long as only the absolute value of the boost is used for exhaustion calculation. This can be achieved by adding two fabs functions**

spaceship.cpp, line 606:
`combat_maneuver_charge -= fabs(combat_maneuver_boost_active) * delta / combat_maneuver_boost_max_time;
`
line 629:
`addHeat(SYS_Impulse, fabs(combat_maneuver_boost_active) * delta * heat_per_combat_maneuver_boost);`

aBlueShadow:
Prevent joystick control from refreshing combat maneuver bar and from cooling down the maneuvering system when moving backwards.

This fixes #426 , using option 2)
Even if 1) is preferred and done later on, using absolute values would make the code more robust.

### DIFF
--- a/src/spaceObjects/spaceship.cpp
+++ b/src/spaceObjects/spaceship.cpp
@@ -603,7 +603,7 @@ void SpaceShip::update(float delta)
     if (combat_maneuver_boost_active != 0.0 || combat_maneuver_strafe_active != 0.0)
     {
         // ... consume its combat maneuver boost.
-        combat_maneuver_charge -= combat_maneuver_boost_active * delta / combat_maneuver_boost_max_time;
+        combat_maneuver_charge -= fabs(combat_maneuver_boost_active) * delta / combat_maneuver_boost_max_time;
         combat_maneuver_charge -= fabs(combat_maneuver_strafe_active) * delta / combat_maneuver_strafe_max_time;
 
         // Use boost only if we have boost available.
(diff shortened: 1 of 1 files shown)


daid:
I agree on that this is better no matter if option 1 or 2 is preferred.

tdelc:
There is still a problem with the joystick : with a backward maneuvering, the refreshing combat maneuver bar reload for me.

aBlueShadow:
Hm.., here both works with that commit (heat increases and refreshing bar depletes when boosting backwards).
I have to admit, I only have a dpad (so only maximal input values) to test here, though I don't think that would matter much.

What about the heat? Does the system cool down (fast) when you boost backwards?.

tdelc:
I think the heat grow up, but i will test that this night.

PR #435, opened 2017-02-19
HotKey Menu for User settings

greatestKeks:
Additional menu point in Options is a Hotkey Menu. 
Hotkeys can be viewed and changed for any category. For categories with a lot of entries this might be a problem, as I not yet have limited the window width.
There is a check if user set SFML key exist and also the key names are made case insensitive.
This menu might in future also allow for Joystick SFML commands to be set by user.


daid:
Boost is a no go.

PR #437, opened 2017-02-20
hotkeyMenu in Options

greatestKeks:
Adding HotkeyMenu for player to set Hotkeys in-game. It is implemented as sub-menu to Options.
Boost dependency has been removed.

To Do: Adapt Menu to Window width.


src/gui/hotkeyConfig.cpp
```diff
@@ -310,3 +335,42 @@ void HotkeyConfigItem::load(string key_config)
...
+                        i++;
+                    }
+                }
+            }
+        }
+    }
+    return false;
+}
+
+bool HotkeyConfig::iequals(const string& a, const string& b)
```
daid: This function has no reason to be in the hotkey config.
The strings are also the class exposed from serious proton, which have a "lower" function. So the whole code can be replaced with "if (a.lower() == b.lower())"

greatestKeks: Ok, thanks!

This resolves my original problem quite nicely. I will make the changes.

JordanLongstaff:
I like this feature, but I just tested it and I'd like to know why the background is a hot pink. Is it a failure to load the standard naval grey background?

oznogon:
This can be closed now that #1050, which incorporates it, was merged

PR #448, opened 2017-03-13
Add option "Can be destroyed ?" 

tdelc:
I added two functions to set if a ship or a station can be destroyed or not. With this option, GM could avoid an unfortunately death of a players' ship.
Option can be set by the script or by tweak in the GM screen.


daid:
Ah, cool, I know the LARP people will love this :-)

amir-arad:
Woot!! I'm LARP people and I love it!

PR #449, opened 2017-03-16
Updated the change log

OldDragon2A:
Sorry updating this took so long, changing jobs has kept me pretty busy.


daid:
Nope, I also changed the OpenAL dll in the windows release, but that seems to cause more problems then it solves :(

PR #457, opened 2017-05-16
Singlepilot and tactical input,: catch up witch 6-crew station inputs.

aBlueShadow:
This adds the targetting hotkeys to tactical and singlePilot, also it adds joystick support to singlePilot.
So all crew-setups should now have comparable input capabilities.

This was basically made by a bunch of copying and pasting. 
Some modular attempt would probably  more elegant, putting joystick mechanics and shared hotkeys in a separate file, but as another user already anounced a major overhaul of the hotkey system, it is probably better not to interfer too much witch that.

But feature-wise,  unless I missed something, tactical should now be completly controllable with keyboard and joystick (or an according custom console)


Wrongtown:
Sounds very cool

PR #460, opened 2017-06-02
New radar screens

tdelc:
I propose 3 news radar screens to improve immersion during a LARP for example. Or to provide to the Captain some news interessant screens.

I only create one file (radarScreen.cpp) with the 3 variations into it.


CaptLapadite:
Would you be kind enough to share a printscreen of theses new screens ? :) 
thanks !

tdelc:
Oh yes sorry :) It is tactical, science and relay screens. I don't add anything else to keep screen simple. 

capture du 2017-06-02 20 41 45
capture du 2017-06-02 20 42 11
capture du 2017-06-02 20 42 25

PR #461, opened 2017-06-06
getWaypointPosition

tdelc:
This is just a function to get the position of a waypoint. Currently, there is no way to get that (or miss I that ?)


oznogon:
https://github.com/daid/EmptyEpsilon/blob/fc4cff5330011a43ced1b71dc22793aed46321f6/src/spaceObjects/playerSpaceship.h#L292

```
player = getPlayerShip(-1)

# return number of waypoints
player:getWaypointCount()

# return coordinates of waypoint 1, or `0.000,0.000` if the waypoint doesn't exist
getPlayerShip(-1):getWaypoint(1) 
```

Screen Shot 2020-02-29 at 11 34 52 PM

PR #474, opened 2017-10-07
Add several artemis missions (DO NOT MERGE BEFORE CLARIFYING LICENSE ISSUE)

foesi:
All of the missions are converted via an enhanced version of artemis_mission_converter.py which I will push at a later point of time.
The missions need a lot of rework and should be tested by a lot of players thats why I made the pull request. I can not find bugs and other discrepancies myself.

But the most important point is if we are allowed to publish them within EmptyEpsilon.
We should clarify this license issue.
Unfortunately there were no license information in any of the missions.
I see two ways to deal with it:
1. Ask the author to put it under a creative commons license or at least ask for permission
2. Rewrite the communication to not copy it word by word. I am far away from beeing a legal expert and I could be wrong but I think this should suffice.


oznogon:
I would strongly advise against adding copyrighted missions not opened by this project to the repository. You could add the conversion script and instructions on converting the missions, which would allow people to test the results. But the only safe path forward is to get permission, which IMO would not expect to get.

foesi:
The conversion script was already created by @daid a and I only added some enhancements.
Unfortunately the script can only convert parts of the artemis missions and you have to do a lot of "manual" work.

But I reached out to FutileChas and Andrew Lacey who created the missions I converted. Maybe they will publish their work under a Creative Commons lincense or similiar so we can use it.

foesi:
Hey guys. I want to give you a short update.
I have an ongoing mail conversation with Andrew Lacey the author of the Ardent and STAR missions.
He said he like the idea that we add it to the EmptyEpsilon repo and gave me a permission to do so.
I asked him if he can repeat it here in the discussion and even if he can make it more specific and maybe publish it under Creative Commons.

On the other hand I just want to explain to you what my motivation is for trying to merge some Artemis missions.
We like to play EmptyEpsilon but are not creative enough to write own mission. But there are a ton of Artemis missions out there which can be converted. But the current conversion is really poor and lacks a lot of features and immersion. To keep the same quality like the original Artemis missions a lot of people have to play and take a look at the converted missions. And where is a better place than this repository.
I am aware that there is a legal issue with copyright and licensing. That's why the first way I proposed of dealing with this, is to ask for permission because the authors already published it for free but often just didn't add a formal license nor a intention how to use it.
The second solution I proposed because I knew copyright and licensing are hard issues but you can not copyright an idea so legally it may can suffice if you just use the idea and not any of the original content. But because I am no legal expert I like to play it safe too and I can understand if we reject missions were the author didn't give the excplicit permission.

daid:
I'm not to worried about the legal thing. I know a lot of open source projects are quire pedantic about this. But for mission files like this, people generally publish does out of love for the game.
There is 1 rule that I'm quite strict about. No Star Trek lore. They are very protective of this IP, and I don't want to get into trouble for a mission mentioning Klingons or anything else (I'm not very familiar with Star Trek, so I have to trust you on this one)

When I wrote the mission conversion script, I did so to get some more test missions for myself. The other reason was because there are quite some tools for Artemis, while there are none of those for EE. So the idea was that maybe we could piggy back on those tools. But nowadays setting up the objects in a world for EE is quite easy with the GM screen.

However, I quickly discovered that some things are vastly different between EE and Artemis. Which doesn't make all features convertible. There is also the difference in ship types, which doesn't translate 1 on 1 very well. So your estimate that every mission needs some handiwork is right. EE also has a few extra features that Artemis lacks and help in the story-telling part IMHO.

Finally, I would love to see your additions to the script, the script is quite old and pre-dates some of the newer features in EE.

PR #477, opened 2017-10-28
Added random object creation helper functions using Perlin noise

vslotman:
I've added a placeRandomObjects function which uses Perlin noise to create a natural looking random area of space.


daid:
:) that could make for some better randomized maps.

PR #483, opened 2017-11-14
Change pacing for beam aiming practice

aBlueShadow:
The beam practice part of the helms tutorial was probably one of the most tedious parts of all EE tutorials until now. Shooting down an immovable ship with decent hull and shields is pretty boring. So I considerably weakened that training ship to get it destroyed much faster.
While I was on it, I also corrected a small typo and put in a reference to the color change while firing.


daid:
Good point :-)

PR #484, opened 2017-11-14
Advanced probe

tdelc:
I propose a new 3D screen for probe. When relay launch a probe, it is like a camera is fixed on the probe, and turn around the probe. One extra screen is needed for this extension.

(maybe it could be add in the code that the button on relay screen is visible only is the screen is selected ?)


k9lego:
Was this ever merged?

PR #485, opened 2017-11-15
Separate coms

tdelc:
I propose to give the possibility to separate Comms from relay. I think that @daid you prefer to keep this unchanged, but I did the change for my fork and so I propose it for you or another.


nallath:
We explictly kept comms & relay together as comms tends to be extremely boring in most matches.

tdelc:
My aim is not to separate into two players, but to provide to one player the two screens. One extra PC is needed, but it is more comfortable for the relay officer.

daid:
@nallath interresting point was made at http://bridgesim.net/discussion/313/review-for-the-last-convention-inoubliable
That running engineering with 2 screens, one for damage control and one for power/coolant control gave a better and easier to control overview. This pull request just puts relay in the same option flow.

tdelc:
I realize that my PR (and the previous one) include more than 250 commits. It is not real because there are some previous commits for test (github is still a mystery for me...) . If you are interested by one or the two PR, I can re PR it with less commits.

To follow @daid, my players always ask me for more and more screens, even with the same number of players. In their opinion (and I am agree), this enhance the immersion.

nallath:
Aah right. That makes sense. I thought it was going to be a whole new station. Putting it on multiple screens makes sense.

daid:
I can "squash" the commits when merging, which means they end up as a single commit in the final result. So don't worry too much about the 250+ commits.
(I still  have to look into the actual changes, in a quick glance it looks fine)

src/spaceObjects/scanProbe.cpp
```diff
@@ -49,7 +49,7 @@ void ScanProbe::update(float delta)
 
 bool ScanProbe::canBeTargetedBy(P<SpaceObject> other)
 {
-    return (getTarget() - getPosition()) < getRadius();
+    return true;
```
daid: This change is problematic for the AI. This (slightly odd) logic is in place to prevent AI ships from actively seeking out probes and destroying them. But only seek and destroy probes when they are near the ships.

src/screens/extra/commsScreen.h
```diff
@@ -0,0 +1,23 @@
...
+class GuiToggleButton;
+class GuiSlider;
+class GuiLabel;
+
+class CommsScreen : public GuiOverlay
+{
+private:
+
+public:
+	CommsScreen(GuiContainer* owner);
```
daid: (indentation)

src/screens/crew6/relayScreen.cpp
```diff
@@ -263,6 +283,8 @@ void RelayScreen::onDraw(sf::RenderTarget& window)
         hack_target_button->disable();
         link_to_science_button->disable();
         link_to_science_button->setValue(false);
+		link_to_3D_port_button->disable();
```
daid: (indentation)

src/menus/shipSelectionScreen.cpp
```diff
@@ -434,11 +452,17 @@ void ShipSelectionScreen::updateCrewTypeOptions()
         crew_position_button[damageControl]->show();
         crew_position_button[powerManagement]->show();
         crew_position_button[databaseView]->show();
+      	crew_position_button[relayOfficerNC]->show();
```
daid: (indentation)

src/menus/shipSelectionScreen.cpp
```diff
@@ -109,6 +122,7 @@ ShipSelectionScreen::ShipSelectionScreen()
     cinematic_view_button = new GuiToggleButton(stations_layout, "CINEMATIC_VIEW_BUTTON", "Cinematic view", [this](bool value) {
         game_master_button->setValue(false);
         window_button->setValue(false);
+		probe_button->setValue(false);
```
daid: (indentation)

src/menus/shipSelectionScreen.cpp
```diff
@@ -96,11 +99,21 @@ ShipSelectionScreen::ShipSelectionScreen()
     window_angle->setSize(GuiElement::GuiSizeMax, 50);
     window_angle_label = new GuiLabel(window_angle, "WINDOW_ANGLE_LABEL", "0 degrees", 30);
     window_angle_label->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
+	
+	// Probe view button
+    probe_button = new GuiToggleButton(stations_layout, "PROBE_BUTTON", "Probe view", [this](bool value) {
+        game_master_button->setValue(false);
+        window_button->setValue(false);
+		topdown_button->setValue(false);
```
daid: (indentation)

src/menus/shipSelectionScreen.cpp
```diff
@@ -86,6 +88,7 @@ ShipSelectionScreen::ShipSelectionScreen()
     window_button = new GuiToggleButton(window_button_row, "WINDOW_BUTTON", "Ship window", [this](bool value) {
         game_master_button->setValue(false);
         topdown_button->setValue(false);
+		probe_button->setValue(false);
```
daid: (indentation)

src/menus/shipSelectionScreen.cpp
```diff
@@ -75,6 +76,7 @@ ShipSelectionScreen::ShipSelectionScreen()
     // Game master button
     game_master_button = new GuiToggleButton(stations_layout, "GAME_MASTER_BUTTON", "Game master", [this](bool value) {
         window_button->setValue(false);
+		probe_button->setValue(false);
```
daid: (indentation)

amir-arad:
Hi
We have a LARP planned for two weeks from now, which relies heavily on EE. 
We had our own approach to separating comms from relay and This PR may be a big win for us (haven't got a chance to try it out).
I see now that this PR may very well be be merged in time for our larp. Is it reasonable to assume there will be a version release immediately after?
can I help this PR in any way?

src/spaceObjects/scanProbe.cpp
tdelc: Ok, I did'nt think about that. I think I did this to select and active the 3D vision just when the probe is launched. So I am not sure this feature is relevant anymore and that the relay officier have to active camera manually.

tdelc:
Sorry about all the indentations. I already see that but my text editor do a lot of strange things... I will check about that the next time.

Oh, and I realize that 3 PR of mine are already merged together... sorry about that, github is not my friend... So I think that this PR include separate coms, advanced probe (with 3D vision) and extra radar screen. If you want to select just one, i can re-PR it.

Last thing : some players said me that the automatic rotation of the probe is not a good idea, mostly when the probe is moving. I like it but I am not sure everyone will like it.

daid:
Which editor are you using?  I'm using "programmers notepad" for quick edits. Which never does "strange things".

I don't mind the 3 merged into one problem. And if the auto-rotation is a bad thing, we can always fix it.  However, to prevent the rotation during probe moving, we could look at the velocity of the probe, and if it's non-zero, view in the direction of the velocity. So you won't get the rotating view until it reached it destination. Until it is there, it shows the flight direction.

tdelc:
it depend : notepad ++ with windows, gedit with linux. I will check about it when i have some time.

You are right about the auto-rotation, it is a good idea. And another thing I don't implement : I think the button "Camera Probe" could be hidden if the station "Probe view" is not activated. With that, every thing is similar for players who don't use the new station.

src/spaceObjects/scanProbe.cpp
amir-arad: Hi,
I have another way to solve this: 
I use the Selectable criteria instead of Targetable, and I make all probes selectable by any object of the same faction (which makes sense to me). so targetable is left with the old slightly odd logic.
see:
https://github.com/amir-arad/EmptyEpsilon/commit/11c201fcffab094c4ab86babc249aa5485594c91

oznogon:
fe8c5c975386efb3b264ce30cf2bffa5bba973bd and PR #765 add separate relay into standalone comms/log and a relay view without comms.

PR #488, opened 2017-11-22
Enable to select selectable objects in Operations

aBlueShadow:
Like in the science station, all selectable objects now can be selected from operations and will show their description, not just those that are "targetable".


daid:
Makes sense as it replaces science+relay.

PR #492, opened 2017-12-12
Trying to clean the GUI code

Lucasdmnt:
The changes includes :
* Add const to getter
* Fix typo
* Merging identical loop

### DIFF
--- a/src/gui/gui2_advancedscrolltext.cpp
+++ b/src/gui/gui2_advancedscrolltext.cpp
@@ -17,12 +17,12 @@ GuiAdvancedScrollText* GuiAdvancedScrollText::addEntry(string prefix, string tex
     return this; 
 }
 
-unsigned int GuiAdvancedScrollText::getEntryCount()
+unsigned int GuiAdvancedScrollText::getEntryCount() const
 {
     return entries.size();
 }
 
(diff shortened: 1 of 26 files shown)


Lucasdmnt:
I'm kinda new to all the github things so i don't know if i have to split this PR into 3 or not.

On an other hand, @daid, i'm not sure why this two loop https://github.com/daid/EmptyEpsilon/blob/5f7b23caf2eedf233fb63645b1a1f58694a43e3c/src/gui/gui2_container.cpp#L22 and https://github.com/daid/EmptyEpsilon/blob/5f7b23caf2eedf233fb63645b1a1f58694a43e3c/src/gui/gui2_container.cpp#L46 where necessary. I tried to merge them and i didn't see any difference. Maybe i don't pay enough attention. I will remove the change if there is a need for them

daid:
As I look at it, yes, those could be combined. The first loop handles positioning and destruction. The second loop does the drawing. As long as no drawing code depends on positioning of other elements, then there should be no difference. Else you could have 1 frame glitches (which sometimes happens anyhow right now)

Lucasdmnt:
Thank you for the verification. And what about the split in multiple PR ?

PR #493, opened 2017-12-12
Change description of black hole

aBlueShadow:
Now that black holes are visible (sort of), I changed the according text of the science db.
Also added a bit explanatory text and defined the radar indicator as a "danger zone" rather than as the event horizon. So we circumvent the problem that is hard to tell when the event horizon is crossed (for the game visuals and even more in reality). Also it's logical that the ship is getting in danger way before crossing the horizon.


daid:
Makes sense :+1:

PR #498, opened 2017-12-15
Redirect mouse clicks on main screen radar

NinjaSpectre:
When mouse clicks occurred on the GuiRadarView elements nothing was being done with them.  The user would have to verify that the mouse pointer was not on the element in order for the click event to be handled.


daid:
I made a different fix, where the radar element simply doesn't capture the mouse click if no handling is attached to that rader element.

PR #507, opened 2017-12-28
Expose calculateFiringSolution to script (for use with commandFireTube)

benhell:
Added "getFiringSolution" script method that takes a tube and a target and returns the angle at which the missile weapon should go (for use as the second argument to commandFireTube).  

This emulates the behavior of firing a missile weapon through the UI.  With this change, a script writer can write the following:

      us = getPlayerShip(-1)
      target = us:getTarget()
      tube_number = 0
      target_angle = us:getFiringSolution(tube_number, target)
      us:commandFireTube(tube_number, target_angle)


daid:
Wouldn't it just make more sense to have a function like this:

```
us = getPlayerShip(-1)
target = us:getTarget()
tube_number = 0
if us:fireTubeAtTarget(tube_number, target) then print("Tube fired!") end
```
?

benhell:
Easy enough.  Revised to "commandFireTubeAtTarget" and modeled after "commandFireTube."  Usage is as you note above, except there is no boolean return type (which is implicit in your description but inconsistent with "commandFireTube" signature).

benhell:
Apologies, accidentally closed the request.

Tuhis:
@daid any update on this? Could this be merged into master?

daid:
I missed the update on this. Due to my work also being on github, I occassionally get spammed with notifications from github and miss things.

PR #511, opened 2017-12-30
enhance max distance (zoom out in radar grid) by x100

amir-arad:
radar grid draws sectors of relevant size


tdelc:
I think this PR could be problematic for arcade sessions of EE. Players could be a little bit lost in this screen. Maybe could you add the max distance as option in the server screen ?

amir-arad:
all it does is allowing the plater / GM to zoom out more. I didn't think it creates a problem, as the player can simply zoom back in.

If it needs to be configurable at the game level, I think it should be part of the map's properties, as it has to do with the size of the map. I already feel that there's room to define the coordinates system at the map level, this can be part of it.

@daid ?

oznogon:
Just built this and have some feedback:

-   The zoom slider doesn't handle zoom levels less than 1.0X; the far left results in zoom 1.0X. This means the extra zoom is inaccessible from the slider, and therefore also inaccessible if a scrollwheel or wheel-emulating gesture is unavailable.

    issue1

-   On relay with default radar ranges, labels on radar objects are clipped by the radar visibility mask starting at around zoom 0.5X and smaller. Larger radar ranges only defer this issue to lower zoom levels.

    issue2

-   On relay with default radar ranges, the lowest zoom levels (<0.3X) have such little visible information that they're effectively useless except for faster map navigation to extremely distant sectors. Would being able to center the map on a sector by entering its name in an input be more effective for that use case?

-   "Zoom" numbers are easy to grasp in the default range, but start to lose their meaning if applied to much larger spaces. The loss of scale is compounded by the new sector folding behavior, which makes massive blocks of sectors look similar to individual sectors in scale. Reporting scale in a more concrete way tied to game mechanics, such as a scale bar in units/kilounits/megaunits instead of "zoom 0.0X", might be more useful, especially to Relay, who can't readily determine distances when zoomed out to less than 1.0X.

amir-arad:
@oznogon I've looked at centering the map on a sector by entering its name. I've found out that sector names doesnt always have printable characters, and translating sector names to position is not single-valued. IMO #512 or some similar fix is a blocker for that kind of feature

PR #512, opened 2017-12-30
sector names logic rewrite

amir-arad:
human friendly, unique sector names with only alphdecimal (upper case) digits
[demo](https://media.giphy.com/media/l0HUlyMxowPHUABQk/giphy.gif)


amir-arad:
I would also ditch the `+5` on x any y. perhaps implement some some global configurable coordinates system?

src/gameGlobalInfo.cpp
```diff
@@ -202,17 +202,30 @@ string getSectorName(sf::Vector2f position)
...
+    }
+    while (sector_y > 0)
+    {
+        int digit = (sector_y - 1) % 26;
+        row = (char)('A' + digit) + row;
+        sector_y = (int)((sector_y - digit) / 26);
+    }
+
+    char column[(sector_x / 10) + 1];
+    sprintf(column, "%d", (int)sector_x);
```
daid: Perfect example of C code being bad for people. The buffer you allocate can be really big if the sector_x value is really large. Which is a waste.
There is also no need for sprintf, as the SeriousProton "string" class has a constructor that converts an integer into a string. (which was used by the original code)

daid:
More configuration = bad.

Also, if you want to change the naming of the sectors, which is fine by me. But then a bunch of the mission scripts have regressions as they mention sectors by name, and thus also need to be fixed.

src/gameGlobalInfo.cpp
amir-arad: yes, I was wondering about that magic... will fix.

amir-arad: regarding the size of the buffer, if the scroll is infinite, then either the sector names repeat themselves or they keep growing. I think growing is the way to go.

amir-arad:
how about making it configurable via lua, but with backward compatible defaults? both the sector names and the coordinates?

oznogon:
I'm not a fan of mandatory configuration, and even less excited about changes that could potentially break existing scenarios.

I do like the idea of exposing sector naming conventions to Lua to make changes like this per scenario, while also keeping the default behavior. I can't speak for anyone else, but I haven't seen or made any scenarios with maps large enough to require deviating from the current behavior.

Not sure how this would affect the logging/replay systems, but those are already out of sync with the game.

src/gameGlobalInfo.cpp
amir-arad: oh, I  think i got what you mean about the buffer. it should be `std::log10(sector_x)`, right?

NinjaSpectre:
Might I suggest changing the while loop into the following:
`while (sector_y > -1)`
`{`
`row = string(char('A' + (sector_y % 26))) + row;`
`sector_y = int(sector_y / 26) - 1;`
`}`
This eliminates the sectors that look out of place by the equator (start with a number instead of alpha)

amir-arad:
I'm working on #518 as a way to opt-in to the new format per scenario

amir-arad:
redundant by  #518

PR #513, opened 2018-01-07
Updated the change log

OldDragon2A:


daid:
You might also want to log the changes in SeriousProton? It's not often that there are fixes there, but this update had the Surface touchscreen fix:
https://github.com/daid/SeriousProton/commit/455c79199410c3cb651a1f6e8b92b705df6b2cf6

PR #516, opened 2018-01-10
Enable scripts to read the order of a CpuShip

czenker:
I would like to expose the orders of a ship to be read in scripts. 

The issue that bugs me though is, that getOrder returns an int to the script at the moment. Any pointers on how to make this a string? I could not find any example code from the project where this has already been done. :)

### DIFF
--- a/src/spaceObjects/cpuShip.cpp
+++ b/src/spaceObjects/cpuShip.cpp
@@ -38,6 +38,12 @@ REGISTER_SCRIPT_SUBCLASS(CpuShip, SpaceShip)
     REGISTER_SCRIPT_CLASS_FUNCTION(CpuShip, orderAttack);
     /// Order this ship to dock at a specific object (station or otherwise)
     REGISTER_SCRIPT_CLASS_FUNCTION(CpuShip, orderDock);
+    /// Get the order this ship is executing
+    REGISTER_SCRIPT_CLASS_FUNCTION(CpuShip, getOrder);
+    /// Get the target location of the currently executed order
+    REGISTER_SCRIPT_CLASS_FUNCTION(CpuShip, getOrderTargetLocation);
+    /// Get the target SpaceObject of the currently executed order
+    REGISTER_SCRIPT_CLASS_FUNCTION(CpuShip, getOrderTarget);
 }
 
 REGISTER_MULTIPLAYER_CLASS(CpuShip, "CpuShip");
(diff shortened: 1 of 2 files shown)


oznogon:
The orders are in an enumerator: https://github.com/daid/EmptyEpsilon/blob/5f7b23caf2eedf233fb63645b1a1f58694a43e3c/src/spaceObjects/cpuShip.h#L7-L19

I think you'd have to add a function or map that associates each value with a string, then expose that to scripts. You could do that in the EE source, but it might be more flexible to make that association in the script, with the understanding that if the orders enumerator's contents change later in EE, you'd need to update that Lua script accordingly.

I didn't test this, but I'd imagine it would be like:

``` lua
ai_states = {"AI_Idle", "AI_Roaming", "AI_StandGround", "AI_DefendLocation", 
             "AI_DefendTarget", "AI_FlyFormation", "AI_FlyTowards",
             "AI_FlyTowardsBlind", "AI_Dock", "AI_Attack"}

exuari_RT4_guard1:orderIdle()

order_text = ai_states[exuari_RT4_guard1:getOrder()]
```

I'd expect `order_text` to then contain "AI_Idle". And those strings in `ai_states` could be anything, like "Idling" or "Attacking", etc., depending on what you're using a string value for.

oznogon:
Actually, looking again at cpuShip.cpp... maybe we could expose the strings already used for each order's comms broadcast? Like:

```
      this->addBroadcast(FVF_Friendly,"Searching for targets."); 
```

could be

```
      order_broadcast = "Searching for targets.";
      this->addBroadcast(FVF_Friendly, order_broadcast); 
```

with an exposed `string getLastOrderBroadcast() { return order_broadcast; }` function?

czenker:
Thanks for your feedback, @oznogon 

There are also strings for the enum in `getAIOrderString` function already that seem to do what we are looking for.

https://github.com/daid/EmptyEpsilon/blob/d3b6e69e8e8d04f920440d11f70320dfbe3d11b3/src/spaceObjects/cpuShip.cpp#L248-L264

I was also thinking about introducing a new method like `getLastOrderBroadcast` that you suggest, but it would be inconsistent with the way other enums like `EAlert` or `ECrewPosition` are handled for the Scripting API:

https://github.com/daid/EmptyEpsilon/blob/45aa7e54a2ce987cd3eb3c6a374eed10ce8b3ffc/src/spaceObjects/playerSpaceship.hpp#L4-L8

I think writing a `convert` function for `EAIOrder` would be the way to go, but I failed at writing it yesterday. What do you think?

daid:
A convert function for EAIOrder is indeed the proper way to solve it. With the template specialization, it's always a bit tricky on where to place the function to be properly picked up.

czenker:
Then I will give the `convert` approach another try this evening. Let's see what my StackOverflow-CopyPaste-C++-Skills can do.

czenker:
Hah! Look what a refreshed brain is able to do. :tada: 

For your consideration, @daid.

PR #517, opened 2018-01-11
Added an faction aura in the radar, once we know what faction the ship is.

LayfonWeller:
This adds a semi-transparent circle to ship (displayed under) and station of the color of the faction, so that they can be identified visually.

I also had to change the color of "Human Navy" faction has the aura was too hard to distinguished from the "Independent" faction

### DIFF
--- a/scripts/factionInfo.lua
+++ b/scripts/factionInfo.lua
@@ -3,7 +3,7 @@ neutral:setGMColor(128, 128, 128)
 neutral:setDescription([[Despite appearing as a faction, independents are distinguished primarily by having no strong affiliation with any faction at all. Most traders consider themselves independent, though certain voices have started to speak up about creating a merchant faction.]])
 
 human = FactionInfo():setName("Human Navy")
-human:setGMColor(255, 255, 255)
+human:setGMColor(128, 128, 255)
 human:setDescription([[The remnants of the human navy.
 
 While all other races were driven to the stars out of greed or scientific research, humans where the only race to start exploring the galaxy because their homeworld could no longer sustain their population. Some other races view humans as a sort of virus or plague due to the rate at which they can breed and spread.
--- a/scripts/factionInfo.lua
+++ b/scripts/factionInfo.lua
@@ -3,7 +3,7 @@ neutral:setGMColor(128, 128, 128)
 neutral:setDescription([[Despite appearing as a faction, independents are distinguished primarily by having no strong affiliation with any faction at all. Most traders consider themselves independent, though certain voices have started to speak up about creating a merchant faction.]])
 
 human = FactionInfo():setName("Human Navy")
-human:setGMColor(128, 128, 255)
+human:setGMColor(0, 255, 255)
 human:setDescription([[The remnants of the human navy.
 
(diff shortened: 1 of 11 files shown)


LayfonWeller:
Here a quick screen shot. Can't fiddle with things to have everything neet right now.
If need be, (ie screenshot of station), I can take other screenshot later tonight.

screenshot from 2018-01-11 13-19-05

oznogon:
Thanks. I can see the green auras fine, but I don't see the other auras without zooming in more on the screenshot. Maybe if they were less transparent or a more contrasting tint against that dark background?

I wonder if there's a different way to solve the problem of identifying faction as well as friend-or-foe. Would tinting the labels instead of adding an aura work? Or outlining the ship icon with the faction color instead of the transparent aura?

LayfonWeller:
The aura are now less transparent, Human Navy is now cyan.

Here are some new image :
ee_shortrangeradar
ee_morevisiblecolors
ee_relaystation
ee_science

I haven't tried tainting the labels nor tried to outline the ship icon with the color aura. If need be I could try coloring the label, but I think that would harder to spot. It currently have no idea how I could outline the ships icon.

daid:
I didn't comment on this pull request yet. But I'll be blunt. I don't like the feature. Sorry.

The faction data is intentionally limited in visibility to give science a bit more to do.
I also don't think the aura's look nice. Generally there are not a lot of different factions in a single mission anyhow.

LayfonWeller:
It is something that is optional, when creating a game. You can disable it (or enable it)
It's an added optional feature, in it's own function, with it's own trigger. 

It was build for LARPs, and in the LARP I go in there is 3 player faction and often 2 other factions.
Player factions are "enemy" to each other, due to the organizer waiting use to choice if we fight or collaborate. When invested by swarms of enemy it happens that a nuke is released at the wrong target.

This feature is there to help prevent that.

I prefer the first look, with the aura pretty transparent, personally. Having a glow effect means that from far away to get a merged of color that helps ID what factions have fleets where and how they are mixed.

I'm open to other looks, or options.

While I agree that pretty much all post needs more stuff to do, in my games, science is not one for the lazy... With helm, it's probably the role that does the most in and out of combat. And if science didn't fully scan the ship, then the aura would not appear, so it exist only if science is doing it's job.

But I'm open to add more things/limits to it : like it can only be seen in short range radar... Or only for weapon officer.

faernElghinn:
I agree with the spirit of the feature: went used in a LARP, faction become mighty important (and the sensor officer is never idle). 

I personally prefer the first transparency, but would select the colours so that all faction are equally visible. Or maybe change the colours of their name tag to the faction once scanned.

The accidental nuke did happen to our team too.

CaptLapadite:
This feature could be of good use for larps with multiple waring factions indeed. 

In a hope to clarify the view and the possible messiness of many auras... could we simply affect the color of the faction to the Callsign text ? Or reduce the aura size. 

Or even more, replace the aura for a simple circle what would be a tiny bit smaller from the shield circle.
This feature also add a visual feedback to the science officer as for which units are done scanning.

PR #518, opened 2018-01-13
custom sector names

amir-arad:
- allow customization of the coordinates system from lua script with these features:
   - translate (x,y) position to sector name 
   - translate sector name to (x,y) position
   - validates sector name form
   - determines the size of a sector
 - add such customization to the basic scenario
 - add sector name field in GM and relay screens that shows the name of the current sector, and allows editing it if the coordinates system supports it.
 - add validation feature to `GuiTextEntry` and use it for the sector name input
   - incl. new color in`colors.ini`


amir-arad:
I see two problems with this solution: 
 - it only works on the game server. 
 - passing a lua function as a string is bad ux

amir-arad:
The feature is working currently in basic scenario in relay screen only. 
I feel that validation UX feedback is still missing, but I wonder if that's a blocker, and I haven't found a reference for a similar use-case in the game.

@daid @oznogon I think I've reached a milestone and would appreciate a review

oznogon:
@amir-arad Thanks for the heads up, I'll take a look (hopefully today)

oznogon:
Looks fine as a prototype. I agree that validation and acknowledgement could help, even if it's as simple as a color flash on the field..

amir-arad:
@oznogon I'm working on a `GuiValidatedTextEntry` ui component to give user feedback on invalid input. I'm currently figuring out the color theme and it's configuration.

What else is missing for this PR to pass as non-POC?

amir-arad:
@daid ?

daid:
(Sorry, busy life is keeping me busy, little free time)

First of, I like the idea of having validation on text fields, also helps on other locations there text entry is done.

I'm not sure how much I like the custom functions for sector names. Feels very over-complicated, even more with functions in strings. For maximum flexibility that most likely nobody else will ever use due to the high complexity...
I understand why they are strings (with network code and everything, in theory compiled lua functions could be transferred, but I'm not even going to try to suggest this as solution, because of the insane complexity of getting this to work within the current scripting and multiplayer framework)

Just thinking out loud here. So don't see this as a "everything is horrible" from my side. Just my first thoughts. I also need to go trough the patch a bit better and put some remarks here and there. It's quite a big patch.

amir-arad:
@daid I've played with LUA's `string.dump`, as it seems more elegant and suitable, but I gave it up when trying to read the result in the cpp side.

I get what you mean when you say it feels over-complicated.
I'm involved in a larp project that relies on very big maps, so this feature can help with that (and later, go-to waypoint with keyboard shortcuts), and it seems that our need for sector naming overhead is not shared by the majority of the current users community, so this feature being an optional extension point that is determined in the map is IMO the most effective way to accommodate both use-cases.

amir-arad:
another thought : Ideally, the coordinates system will be exposed in LUA as well, so that communication scripts can use it instead of hard-coding sector names. however this is much easier to achieve inside the map's logic, as it's the map that loads both the coordinates system and the communication scripts.

Nevrose:
Hi, this is my first post here.

More than a sector Name, I'd like two things that would I think better serve the prupose :

- Define a visible areas (you build a polygone with a color that is displayed in the background (behind the sector grid and spaceobjects icons). With a fonction that helps to tel you if a (x,y) coordinate or a space object is WITHIN ou WITHOUT the area. Visible areas are displayed in the communication view.

- Add a label spaceobject that is only displayed in the communication view.

amir-arad:
the benefits of this approach:
1. it's backwards compatible
2. it allows each map author to have their own sector naming scheme (for compatibility with other games etc.)
3. (derives from 2) - it allows for custom sector names - like having a sector named "danger zone 1"
4. having the sector logic available in lua means that the dialog strings can use the same logic to resolve object position to sector name (as a way to write cleaner comms code for static locations, or as a feature for dynamic locations - like report the current sector where some moving object is located)
downside : complicates the code

k9lego:
Any chance we could highlight certain sectors or draw a area with a red line, as to mark it as "danger zone." It would allow a script to mark a group of sectors, or 5 whole sectors and half of two as dangerous.

Nevrose:
About coloring a danger zone, wasn't some of that covered with the new Zones functions ?
[Issue 529: REQUEST Areas & Labels](https://github.com/daid/EmptyEpsilon/issues/529)

> daid commented on 20 Mar
> Started work on this. Basic colored zones seem to work now, labels not yet.
> 
> Simple test:
> 
>     Zone():setPoints(
>         -1000, -1000,
>         1000, -1000,
>         1000, 1000,
>         -1000, 1000
>     ):setColor(255, 0, 0)

k9lego:
Gotcha! Thanks for pointing me to that

daid:
On 2, It allows, but most likely it won't, the usage is so complex and error prone, that only your usage will be "stable"
On 3, the new zones solves this better IMHO.
On 4, objects already have the getSectorName function. ```print(player:getSectorName()```

amir-arad:
thanks @daid 
so if we're left with only 1 and 5, I suggest making it a second hard-coded (CPP) second coordinates strategy.
this way it's still backwards compatible (the map needs to activate it explicitly), and in the new strategy you get the "goto" ability (debatable whether the text field will still be displayed when not editable, for consistency)

If it's not a big burden on the code, `sectorSize` can still be configurable, but only when the map activates the new strategy (default remaining 20000)

Issue #520, opened 2018-01-21
SpaceObject:takeDamage does not allow to specify the origin


czenker:
Damaging a ship with multiple shields in a script with `SpaceObject:takeDamage` will always damage the shield in the direction of the center of the map.

The new `ExplosionEffect` is great and allows to create new sources of damage ingame. Players might be confused if the wrong shield is damaged.

Pull request incoming... stand by...

czenker:
Also added `frequency` and `system_target` while I was at it.

This is the code I used for testing.

```lua
-- Name: Testing damage
-- Description: Testing stuff
-- Type: Mission

function init()

    local test = PlayerSpaceship():setFaction("Human Navy"):setTemplate("Atlantis"):setImpulseMaxSpeed(250):setRotationMaxSpeed(50)

    addGMFunction("energy damage", function()
        test:takeDamage(10, "energy")
    end)
    addGMFunction("emp damage", function()
        test:commandSetShields(true)
        test:takeDamage(10, "emp")
    end)
    addGMFunction("directional emp", function()
        test:commandSetShields(true)
        test:takeDamage(10, "emp", 20000, 0)
    end)
    addGMFunction("400Hz energy", function()
        test:commandSetShields(true)
        test:takeDamage(10, "energy", 0, 0, 0)
    end)
    addGMFunction("600Hz energy", function()
        test:commandSetShields(true)
        test:takeDamage(10, "energy", 0, 0, 10)
    end)
    addGMFunction("target system damage", function()
        test:commandSetShields(false)
        test:takeDamage(10, "energy", 0, 0, nil, "warp")
    end)

end
```

daid:
Code changes look fine, but I'm not spotting a pull request, only a commit referring   this issue?

PR #522, opened 2018-01-29
Add directional damage to scripts

czenker:
Pull Request for issue #520 .

Thanks for the reminder, daid.


daid:
Note, instead of lua_isnumber, you could also have used lua_gettop, which returns the amount of parameters that is given to the lua function.

PR #527, opened 2018-03-01
Refactor mission converter script

foesi:
As mentioned in #474 I tweaked the mission converter script a little bit.

### DIFF
--- a/scripts/shipTemplates_Corvette.lua
+++ b/scripts/shipTemplates_Corvette.lua
@@ -80,6 +80,24 @@ variation:addDoor(8, 4, false);
 --variation:addDoor(2, 2, false);
 --variation:addDoor(2, 5, false);
 
+-- small enemy corvette
+
+variation = template:copy("Cruiser Q8")
+variation:setDescription([[The enemies version of the Atlantis X23.]])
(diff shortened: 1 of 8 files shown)


scripts/shipTemplates_Corvette.lua
```diff
@@ -80,6 +80,24 @@ variation:addDoor(8, 4, false);
 --variation:addDoor(2, 2, false);
 --variation:addDoor(2, 5, false);
 
+-- small enemy corvette
+
+variation = template:copy("Cruiser Q8")
+variation:setDescription([[The enemies version of the Atlantis X23.]])
```
daid: But the "Atlantis X23" is already an enemy ship? The "Atlantis" is the player variant of that...

foesi: I took a look into it and remembered that I used to have a list where I could see the hullID and the correspnding stats of the ship (In Artemis). If I remember correctly hullID 5002 was a ship similar to the players ship (in Artemis) and that is why the template was created. My proposal is to either delete the template or give it a new and more meaningful description. But I can not come up with a good idea in the moment.

PR #528, opened 2018-03-04
Fixed E1.31 DMP layer packet octet 118 value to 0xa1

Tuhis:
It seems that there were small typo in E1.31 data packet format on DMP Layer. For reference please see http://tsp.esta.org/tsp/documents/docs/E1-31-2016.pdf page 5 which defines data packet format.


daid:
You are right, seems the software I tested against ignored that field...

PR #540, opened 2018-06-07
Add a race type mission script: Fermi 500

Xansta:
I wrote another mission script based on the Regatta race suggestion made by jrwarwick in the starship bridge simulation network forum under the "Ideas for mission scripts and scripting system improvements" topic. It's not exactly as described, but many of the ideas are there.

Try it. Test it. Let me know what you think.


daid:
I haven't tested it, but the script looks fine.

daid:
I noticed you also had some updates on your other scenario in here.

PR #541, opened 2018-06-07
Correction of GM comms

tdelc:
When player want to contact in GM mode, there was a bug with my previous PR, sorry.


daid:
Can happen ;)

PR #546, opened 2018-07-11
Added automatic weapon tube reloading feature

JordanLongstaff:
Pretty straightforward. Each tube now has a toggle button for automatic reloading of missiles. If a missile is fired while auto-reloading is on, another missile of the same type is immediately loaded if available. If there are no available missiles of that type, the tube will be loaded when a missile is available. If a tube is unloaded, auto-reloading is automatically turned off. If auto-reloading is turned off while a tube is empty, it will forget what type of missile it wanted. If auto-reloading is turned on while a tube is empty, the tube will not begin loading until a missile type is selected and loaded into the tube.

For rear-facing tubes, which can only hold mines, this will be quite useful. For front-facing tubes, this will be useful in situations of standardizing on a particular attack pattern.

WARNING: Untested!


daid:
Seems you put more things in this pull request then you intended to put in?

JordanLongstaff:
That's better. I didn't realize GitHub would retroactively add new commits to the pull request.

JordanLongstaff:
The recent changes to this repository are making things complicated. I will have to remake this pull request from scratch.

PR #550, opened 2018-08-20
Mostly implemented clock like bearing system.

JonathanILevi:
Me and my friends what to use a different bearing system that is, like a clock, based on twelve rather than 360.
This commit just effects the radar indicators and the Helms' tool-tip.  I am working on making it work everywhere.  Changing what radar system is used is a setting in `gameGlobalInfo` and right now the only way to change what type, is to recompile with a different value.

I hope this new bearing system would be appreciable to others.

### DIFF
--- a/scripts/scenario_00_basic.lua
+++ b/scripts/scenario_00_basic.lua
@@ -19,22 +19,48 @@ require("utils.lua")
 --      Returns a relative vector (x, y coordinates)
 --   setCirclePos(obj, x, y, angle, distance)
 --      Returns the object with its position set to the resulting coordinates.
+--   distance(a, b, c, d)
+--      Returns the distance between two objects/coordinates
+--   angleRotation(a, b, c, d)
+--      Returns the bearing between first object/coordinate and second object/coordinate. 
 
 -- Global variables for this scenario
 local enemyList
 local friendlyList
 local stationList
+local addWavesToGMPosition      -- If set to true, add wave will require GM to click on the map to position, where the wave should be spawned. 
+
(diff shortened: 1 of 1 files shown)


daid:
Interresting, sounds like a simple addition.

Instead of setting this as a game_global_info, you could set it as a client configuration from the options.ini, example:
https://github.com/daid/EmptyEpsilon/blob/master/src/main.cpp#L183

JonathanILevi:
I went with game_global_settings because the entire ship should have the same setting or it is a little weird.  But, if you would prefer I could put it in options.ini.

JonathanILevi:
It would also work as a ship setting (so different ship can be different).  Along with warp/jump.  What if there was an interface for selecting per ship settings when a ship is created?

JonathanILevi:
> I think this would work even better if EmptyEpsilon had a "Relative Display Mode" option whereby the ship always points up on the Helm screen (this is a feature in Artemis SBS). Or even if this isn't implemented, at least have the clock numbers revolve around the compass to keep 12 always in the direction the ship is facing (so the directions are always relative to that direction).

That is a cool idea.  I will see if I can implement that when I get around to it (:

JonathanILevi:
The clock like bearing should be working everywhere except the GM screen (which I do not think is important currently).

I added a file for the bearing logic `screenComponents/logic/heading.x` which has show functions that convert a bearing number into the appropriate string.  I am not sure that this is the best place for this file; do you have an idea for a better place? 

The only thing yet to do is a way of changing the setting.

PR #557, opened 2018-08-28
fix overlay in radar screen painting everything black in mac

### LINKED ISSUE #554: MacOS Science/Radar Station Rendering All Objects as Black
screen shot 2018-08-24 at 7 32 19 pm

It seems as if the objects on the Radar are rendering all black in the Science station on the 1 player ship.

This was a scenario with two Macs that had built their own game from source and altered the cmake to version zero and a Windows machine at Tactical. A Mac was the server.

I haven't had the opportunity to play on other devices to confirm what it "should" look like. 

Objects are clickable and scannable, but always black and essentially invisible. You can see the outline of one rectangular object at the bottom right of the green cloudy area.

amir-arad:
this change should theoretically do nothing - it uses the mask's alpha channel instead of color (both of them are set to all 1's) for the BlendMode's destinationFactor. 
This fixes an issue in OSX that painted all the objects in black, with the price of not rendering the mask itself on the screen.
fixes #554

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -81,8 +81,7 @@ void GuiRadarView::onDraw(sf::RenderTarget& window)
     if (fog_style == NebulaFogOfWar || fog_style == FriendlysShortRangeFogOfWar)
     {
         drawRenderTexture(mask_texture, forground_texture, sf::Color::White, sf::BlendMode(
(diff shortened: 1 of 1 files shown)


amir-arad:
:+1: release version please? I haven't yet managed to build for windows and I'd like to use this in mixed OS network

Issue #556, opened 2018-08-25
First Attempt at a Windows (Code::Blocks and MinGW) Compile


msbeck:
I'm trying to get things compiled on windows in order to make a Windows build. I'm coming up against a few things. I had a lot of other errors, but sorted though those. I've made it as far as building the **debug** exe and running it, but it seems like the DLLs aren't working as expected.

![bin directory of dlls](https://user-images.githubusercontent.com/2982395/44622848-ee2fe380-a886-11e8-9269-fecc5182b6cf.PNG)
All my DLLS

![entry point not found](https://user-images.githubusercontent.com/2982395/44622850-f38d2e00-a886-11e8-8624-33bda63c5bb3.PNG)
![entry point not found 2](https://user-images.githubusercontent.com/2982395/44622851-f556f180-a886-11e8-8384-7b40f7b91796.PNG)
![entry point not found 3](https://user-images.githubusercontent.com/2982395/44622853-f6881e80-a886-11e8-834a-b3e67dd3c925.PNG)
![entry point not found 4](https://user-images.githubusercontent.com/2982395/44622854-f7b94b80-a886-11e8-9c2c-7624812de972.PNG)
Missing Entry Points

![project directory](https://user-images.githubusercontent.com/2982395/44622858-09025800-a887-11e8-8dd9-a38bbac90341.PNG)
Project Directory

![toolchain executables](https://user-images.githubusercontent.com/2982395/44622859-10296600-a887-11e8-904c-da5c17ae46d6.PNG)
![search directories - compiler](https://user-images.githubusercontent.com/2982395/44622860-115a9300-a887-11e8-9a17-220c1794415b.PNG)
![search directories - linker](https://user-images.githubusercontent.com/2982395/44622861-128bc000-a887-11e8-90fd-66cf1823bc6d.PNG)
Compiler settings

![sfml 2 3 2](https://user-images.githubusercontent.com/2982395/44622866-1fa8af00-a887-11e8-8ac0-2da17b89ad98.PNG)
![codeblocks16mingw](https://user-images.githubusercontent.com/2982395/44622867-20d9dc00-a887-11e8-9597-b06f356a1678.PNG)
SFML 2.3.2 and CodeBlocks 32 bit Version 16.01

Do I have the wrong DLLs included?

[...]

PR #560, opened 2018-09-17
code::blocks build with sfml 2.5.0 and drmingw 0.8.2

amir-arad:
following the advice in #556 This is the updated code-blocks project configurations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,8 +30,13 @@ thumbs.db
 
 # output directory
 .bin
+.obj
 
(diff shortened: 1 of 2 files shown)


JordanLongstaff:
If this pull request gets merged, the wiki should also be updated.

amir-arad:
> If this pull request gets merged, the wiki should also be updated.

yes. however I can't add a wiki update to the PR, so one thing at a time

PR #561, opened 2018-09-19
updates to CMakeLists.txt to make it more manageable 

amir-arad:
These changes allow working on EmptyEpsilon from [vscode](https://code.visualstudio.com/). It's a very smooth dev environment.
I've made a [guide and some configuration files](https://gist.github.com/amir-arad/169993b47e97034277e0e5dfe18b1397) in how to set up the environment, there's feature in that setup that depends on the changes in this PR.
the changes for CMake are minor:
* add `COMPILE_DEV_HELPER` option to set the `DEBUG` preprocessor value (opt in)
* add messages to print the value of options 
* fix a bug where `DRMINGW_ROOT` is required even when `ENABLE_CRASH_LOGGER` was not defined by the user

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,6 @@ cmake_minimum_required(VERSION 2.8.12)
 
 option(ENABLE_CRASH_LOGGER "Enable the drmingw crash logging facilities" OFF)
 message(STATUS "ENABLE_CRASH_LOGGER is " ${ENABLE_CRASH_LOGGER})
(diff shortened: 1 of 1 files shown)


amir-arad:
Honestly I don't know. This is the first and only CMake project I know, so i'm guessing the standards from what I see here and read online in CMake tools documentation. I just assumed you didn't couple it with CMAKE_BUILD_TYPE  for a reason.
I like the idea of fine grained control over the build from outside of CMake, but I can't justify it in this case.

daid:
Main reason is lazyness :-) as the DEBUG define I use during development, and then I use the code::blocks build, the CMake I only use during the release process.

PR #567, opened 2018-09-25
New screens

amir-arad:
* 3D probe screen
* 3 types of radar (3U, 5U, grid)  screens
> I propose 3 news radar screens to improve immersion during a LARP for example. Or to provide to the Captain some news interessant screens.
> 
> I only create one file (radarScreen.cpp) with the 3 variations into it.


amir-arad:
really sorry for the noise, my mistakes

PR #568, opened 2018-10-01
Added automatic weapon tube reloading feature

JordanLongstaff:
Here's my second attempt to add this feature. Very simply, it adds an "AUTO" button to each missile tube that, if on, will reload a missile tube as soon as a missile is fired from that tube, with the same type of missile if one is available. Also, as of now this feature has been tested by me, and the only known bug is that sometimes the loading bar fails to appear during the reloading process.


JordanLongstaff:
I've added an impromptu fix for the loading bar issue. What seems to be happening is that the client is not recognizing that the tube is in a loading state, either because it is not told or because the tube is switched back to a loaded state before the information can be relayed to the client. So I added a "fix" in the update method which re-modifies the tube's state while auto-reloading is active. There can still be a slight delay in the update, though. I would like to make the transition seamless if possible.

daid:
More likely, I see that you make no distinction between client and server when starting the auto-load, causing the client to make a decision to start an autoload on the client. At which point the client data is out-of-sync.

General rule on EE's network code, decisions are only made on the server, and the result of that decision is replicated trough the network.

JordanLongstaff:
Fair enough, but it's hard to tell exactly where the distinction between client and server lies. I think I know what part of the code you are referring to (it's in the fire method, which checks for auto-reloading in order to decide whether to reload the tube or leave it empty); I was under the impression that this was a server-side function because it's triggered when the server receives a command packet from the client. In which case I don't know how the decision to reload gets sent back to the client.

If you're referring to auto-reloading on re-stock, I remember testing that and it worked just fine. But maybe that's not a guarantee, I don't know.

JordanLongstaff:
Can you specifically refer to me where the client "makes a decision to start an auto-load on the client"?

JordanLongstaff:
I think I found the real problem! The weapon tube's "delay" parameter, which is the measure of how much progress is being made in loading a tube, has a 0.5 second server update delay value. I'm guessing this is intended to make the loading bar itself look more seamless on the client, but what ends up happening with the auto-reloading is that the tube fires, then sets the tube state to Loading and sets the "delay" parameter to equal the standard load time, but when the server sends this update to the client, it sends the state update but not the delay update, at which point the client thinks the loading has finished and sets the tube's state to Loaded. I tried removing the server update delay and it fixed the issue.

JordanLongstaff:
So with that in mind, would it be OK to remove the server update delay or would that make the loading bar too choppy? This was an observed side effect in my testing.

daid:
The delay is also in place to prevent a large network load, as the server will send the update every thick otherwise.

The bug is that the client shouldn't change the state of the tubes, pretty much, ever.

JordanLongstaff:
Which it's not doing AFAIK! The PlayerSpaceship::fire method is called by the server, correct? Well, that's the method that then calls startLoad if auto-loading is set. I don't know where the client would be attempting to change the tube's state independent of the server. My only guess would be that it's somehow calling PlayerSpaceship::update and thinks, erroneously, that the loading has already finished and sets the tube's state to Loaded. It doesn't do this if the update delay on the loading delay variable is removed.

JordanLongstaff:
As suspected, the WeaponTube::update method is run on the client. It is what changes the state to Loaded. By your own statement, the client should not be changing the state of the tubes, ever.
https://github.com/daid/EmptyEpsilon/blob/a3a2220b8f122f113bbab393b269ad9bdc0f91d2/src/spaceObjects/spaceshipParts/weaponTube.cpp#L214
I can only guess that this bug is the result of a timing issue where the client mistakenly thinks a tube load has just finished when it has just started.

JordanLongstaff:
As suspected, WeaponTube::update is being called by the client, That is how the state changes to Loaded on the client. By your own statement, the client shouldn't change the state of the tubes, ever.

https://github.com/daid/EmptyEpsilon/blob/a3a2220b8f122f113bbab393b269ad9bdc0f91d2/src/spaceObjects/spaceshipParts/weaponTube.cpp#L211-L215

I can only guess that this is the result of a timing issue where the client erroneously thinks a tube load has just finished when it has actually just started.

JordanLongstaff:
As suspected, WeaponTube::update is called on the client. That's how the tube state is set to Loaded. By your own statement, the client shouldn't change the state of the tubes, ever.

https://github.com/daid/EmptyEpsilon/blob/a491819aaddafa5d8d3e907ed6aa15b1693fb737/src/spaceObjects/spaceshipParts/weaponTube.cpp#L220-L232

amir-arad:
sometimes there's a confusing mixup in update logic between server and client. 
@daid why does `update` even invoked in the client side?

JordanLongstaff:
I don't know what it is in general, but for weapon tubes I can only imagine it's to create the effect of the loading bar filling up even though the client has to wait 0.5 seconds for a progress update. There may be other objects that have update delays on certain variables. I guess what would be needed then to prevent the client from updating the wrong things and thereby getting out of sync would be to add conditions on being called on a server vs. a client, which should be easy enough to implement.

JordanLongstaff:
Yup...I just did a test and I can see clearly now that I am spot on: the client thinks the tube is already full and sets the state to Loaded unaware of the server's knowledge of the actual situation.

amir-arad:
@JordanLongstaff cool! how about suggesting a fix PR for that? maybe even separate to the auto-reloading feature

JordanLongstaff:
And done! This pull request now requires this other pull request to be made in SeriousProton:
https://github.com/daid/EmptyEpsilon/pull/568

JordanLongstaff:
I just resolved some conflicts with this PR, but I don't know why Git is suddenly thinking that I'm making changes to three entire files when all I'm really doing is adding a couple of lines. I hope it's not a line termination issue.

JordanLongstaff:
Fixed it. 😛

JordanLongstaff:
It's worth mentioning, this feature intentionally does not auto-reload HVLI ordnances. I don't know if there are players who would use HVLI prominently enough to warrant auto-reloading HVLI, but if anyone wants me to change this feature so that HVLI ordnances are auto-reloaded, I can do so. It should be easy enough.

JordanLongstaff:
This should be ready for merge now if desired. If we want the ability to auto-reload HVLI, that can be done in a future PR.

JonathanILevi:
I am not convinced that auto loading would be a good feature.

The problem is that it would make playing without auto-loading inferior.  And, therefore, practically take away the choice of the weapons officer to load something else.

One idea which would not have that problem is having a sort of prepared to load slot for each tube which would be automatically loaded when the tube is emptied.  (And then the prepared to load slot would be emptied, but allows the weapons officer plenty of time to prepare another missile.

As it is, I am afraid, I would not like this feature.

JordanLongstaff:
The feature comes from my experience playing Artemis; in my group, different sequences of missile combinations (e.g. EMP followed by nuke) were common, and still are. I wanted to have a feature that made it possible to have certain type of missiles always be ready ASAP. One element of EmptyEpsilon's tubes that makes this work well is that ships can have a rear tube that can literally only load one type of missile (mines), so there's never a point to having that tube empty if it can be avoided. This feature would eliminate that concern.

The feature doesn't sacrifice the weapons officer's control of tube contents. If a tube needs to have different contents, the weapons officer can turn that tube's auto-reload off (BTW, it's off by default). It also turns off automatically if a tube is already loaded and the weapons officer decides to unload it.

PR #572, opened 2018-10-13
Philips Hue Light Count

rdbahm:
One of the listed "TODOs" in philipsHueDevice was to determine how many lights were returned in the JSON response.

- I implemented this as a standalone string parser instead of adding a JSON parsing library, as I'm not sure on this project's stance on adding additional libraries. If adding a library for this purpose makes sense at this time, I'd be happy to do that instead.
- The code does not count the lights returned by the Hue. Instead, it finds the highest light index and returns that value. This is specifically  because other parts of the code (setChannelData for one) depend on the light_count being greater than or equal to than the light index - otherwise the light won't update.

### DIFF
--- a/EmptyEpsilon.cbp
+++ b/EmptyEpsilon.cbp
@@ -297,6 +297,8 @@
 		<Unit filename="../SeriousProton/src/httpServer.h" />
 		<Unit filename="../SeriousProton/src/input.cpp" />
 		<Unit filename="../SeriousProton/src/input.h" />
+		<Unit filename="../SeriousProton/src/json11/json11.hpp" />
+		<Unit filename="../SeriousProton/src/json11/json11.cpp" />
 		<Unit filename="../SeriousProton/src/logging.cpp" />
 		<Unit filename="../SeriousProton/src/logging.h" />
 		<Unit filename="../SeriousProton/src/lua/lapi.c">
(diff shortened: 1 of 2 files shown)


daid:
Mmm, I understand this works, but it's not very pretty (I assume you agree on that)

I'm also not sure yet if I want to pull in a json library. I've used json11 (1 header+1 source file) with great success in another project. And that only adds a small amount of extra code.
https://github.com/dropbox/json11


On libraries, as you might have noticed, in the SeriousProton repo, there are copies of a bunch of libraries, to minimize external dependencies and simplify the building of the code for different targets. It's not the linux way, but it makes life a lot easier on windows.
So if we add a library, the proper place would be to add it there.

rdbahm:
Totally agree - I wasn't sure your policy on adding new libraries.

I've set it up to use json11 in the philips hue device to find the key.

rdbahm:
Not much activity on this one recently, which is fine (I know rule 1, "this project is a dictatorship." 😄) - if you don't want to go down the path of adding an additional library to the system, my personal use case would be handled by increasing the count, even if it's not detecting the count dynamically. It looks like it'd be simple to add channel count to the configuration file, which could be done with relatively few changes to the code and no new libraries. Thoughts?

daid:
Something was wrong with my github notification settings, so I missed a whole bunch of things.

PR #576, opened 2018-10-16
Restore the ability to have tutorials repeat

NinjaSpectre:
Allows for dedicated tutorial stations (again).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -22,7 +22,6 @@
 #include "httpScriptAccess.h"
 #include "preferenceManager.h"
 #include "networkRecorder.h"
-#include "tutorialGame.h"
 
 #include "hardware/hardwareController.h"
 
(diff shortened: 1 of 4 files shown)


daid:
I'm confused by this patch, while it does clean a few things up, it changes the behaviour of the "tutorial" option. It was intented to boot the game straight into the tutorial, and repeat itself. Now I see this behaviour has moved to the tutorial button with your patch...

NinjaSpectre:
The problem (somewhat) is that a Tutorial Menu has been added in between.  Agreed that the program should start up at the Tutorial Menu in this case, however should the setting also be expanded?  Should a number value given in the settings correspond to a specific tutorial to start?  Currently as I have it I left the choice of which tutorial to repeat as the one selected and started from the Tutorial Menu.

NinjaSpectre:
Any further updates needed on my side?  I would squash this but unfortunately I am not currently set up correctly to do so.

Issue #462, opened 2017-06-10
customizing Artifact behavior


amir-arad:
I see how I can customize the appearance of an artifact, but I don't find a way to define what happens when it is picked up. seems to me like it can be a great feature to add a pickup function to an artifact, that receives as an argument the colliding object.

This function can also serve as a filter (so that other space objects will be able to pick up the artifact). 

If that sounds like a welcomed addition, I would be happy to work on a PR.

Issue #23, opened 2015-02-27
Collision detection is off for really large ships


vslotman:
When I add a really large ship to the game (for example a Battlestation) and let AI-ships attack it, the AI-ships jump back and forth very fast when they are close to the Battlestation. 
And it might by by design, but a Battleship cannot use its beams when its target is too close.

daid:
The battlestation is currently unfinished, there are quite some AI issues when you add this one to the game, it's more a showpiece then a gamepiece right now.

PR #586, opened 2018-10-31
allow to configure a callback when an Artifact is picked up

czenker:
Because it was mentioned in the forums recently and it is also on my wishlist I had a look how it was possible to have a callback function if an `Artifact` is picked up.

This pull request depends on [SeriousProton:#23](https://github.com/daid/SeriousProton/pull/23). But if both are approved I think there are lot more places to utilize this pattern – like when a ship is destroyed, docks or undocks a station etc.

Here is how it could be used:

    function init()
        PlayerSpaceship():
            setTemplate("Piranha"):
            setFaction("Human Navy"):
            setCallSign("Thief")
        Artifact():
            setCallSign("Stack of Gold"):
            setPickUpCallback(function(foo, bar)
                print(foo:getCallSign() .. " picked up " .. bar:getCallSign())
            end)
    end

### DIFF
--- a/src/spaceObjects/artifact.cpp
+++ b/src/spaceObjects/artifact.cpp
@@ -71,9 +71,7 @@ void Artifact::collide(Collisionable* target, float force)
     {
         if (on_pickup_callback.isSet())
         {
-            on_pickup_callback.addObjectArgument(player);
-            on_pickup_callback.addObjectArgument(this);
-            on_pickup_callback.call();
+            on_pickup_callback.call(player, P<Artifact>(this));
         }
         destroy();
     }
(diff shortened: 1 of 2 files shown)


daid:
I think you also patched SeriousProton, but didn't submit a patch for it.

Also, with C++11, you can make the parameters and call in a single function, dynamic:
https://github.com/daid/SeriousProton2/blob/master/include/sp2/script/callback.h#L20

I'm not sure how good the SeriousProton code adapts to this method. (I haven't developed on that engine for quite a while, and put all my learnings from it in SeriousProton2)

amir-arad:
fixes #462

czenker:
> I think you also patched SeriousProton, but didn't submit a patch for it.

The pull request for Serious Proton is here: https://github.com/daid/SeriousProton/pull/23
 
> Also, with C++11, you can make the parameters and call in a single function, dynamic:
> https://github.com/daid/SeriousProton2/blob/master/include/sp2/script/callback.h#L20

That sounds like the better approach because then there is no more state in the ScriptSimpleCallback. Not sure if I get frustrated before I figure out how to make it work, but I will give that approach a try.

daid:
If you don't get the vararg template to work, don't worry, I know it's a bit of black magic. I will accept the patch with out.

czenker:
I have updated both pull requests.

daid:
Great work!

PR #588, opened 2018-11-03
AI Missile Behaviour : prevent allies to fire on themselves

simchanu29:
I have tweaked a bit the calculateFiringSolution function in ai.cpp to prevent ships to fire missiles on their allies. It has been tested with success on Ubuntu 16.04 with 50ish ships firing HVLI.

I'm not sure why my pull request removes and add so much lines. Maybe an issue with the file rights.


simchanu29:
Well, it doesn't disable friendly fire, it's a change in the AI code that deter AIs from firing missiles if an ally is on the path to it's target.

I don't think not deterring AIs from firing inadvertently on allies is a good idea, that's why I don't see the use of an option disabling it. Though, having an option to set friendly fire would be nice indeed, and yes, I definitely think friendly fire is a good idea. It's quite funny when your crew blow up a friendly ship by mistake.

That gave me an idea for AIs, crews could be experienced or inexperienced, Maybe rookie ships could make the mistake to fire on allies and not veteran ships. That would be for another PR, but I think it would be fairly easy to add.

JordanLongstaff:
So now there are no changes...do you want to close this PR and start another one? Or did you push a commit not knowing it would propagate to this?

simchanu29:
Ah, I my mistake. I've not used often Pull Requests. Yes I reverted a commit on master to track daid's master branch and commited my changes on a develop branch. Theses changes have been propagated.

I will start another PR to keep my changes on my fork.

PR #589, opened 2018-11-05
AI Missile Behaviour : prevent allies to fire on themselves 

simchanu29:
[PR recreated following some mistakes]

I have tweaked a bit the calculateFiringSolution function in ai.cpp to prevent ships to fire missiles on their allies. It has been tested with success on Ubuntu 16.04 with 50ish ships firing HVLI.

I'm not sure why my pull request removes and add so much lines. Maybe an issue with the file rights.


simchanu29:
[From previous PR | JordanLongstaff ]

Seems pretty cool, but maybe this should be controlled by a server option? Like, perhaps on the server setup screen there should be a "Friendly Fire" button that can be turned on or off to respectively allow or disallow ships firing missiles on allies. Would this be a good idea? Would there be a point of ever allowing friendly fire?

simchanu29:
[From previous PR | Simchanu29 ]

Well, it doesn't disable friendly fire, it's a change in the AI code that deter AIs from firing missiles if an ally is on the path to it's target.

I don't think not deterring AIs from firing inadvertently on allies is a good idea, that's why I don't see the use of an option disabling it. Though, having an option to set friendly fire would be nice indeed, and yes, I definitely think friendly fire is a good idea. It's quite funny when your crew blow up a friendly ship by mistake.

That gave me an idea for AIs, crews could be experienced or inexperienced, Maybe rookie ships could make the mistake to fire on allies and not veteran ships. That would be for another PR, but I think it would be fairly easy to add.

daid:
Most likely, the diff is so large because you changed line-ends.

simchanu29:
I have fixed the line-ends.

daid:
This patch does assume that the missiles file in a straight line from the source to the target. As this is not always the case (especially with the volley-ai) it might incorrectly decide not to fire...

And I do see cases where it would still hit friendlies. However, I do think this patch will help with the very obvious friendly-fire scenarios.

I also notice that the added code does not account for the size of the friendly, Not sure how easy that is to add, as it becomes quite more complex quite fast...

nallath:
So uh. This PR was kinda abandoned. Is it still worth it to pursue this? From what I see the consensus is that the behavior is slightly better than before (so hey, that's a win, right?).

simchanu29:
Oh, i totally forgot this PR. I will look into it, probably this weekend, to improve and test this behaviour :)
It should be near completion and only need a few test to insure the behaviour is better.

simchanu29:
You were right about broadside AI, this fix does not work good for them. I changed the triangle of fire safety from a search around the tube heading to a search around the angle from the ship to the target. It works better.

Overall it cover some cases with missile ships in line no longer firing on the next ally in the line trying to fire on enemies further ahead. But there is still a lot of cases that are not taken into account because the firing ship would need to know the trajectory of its missile and the the planned position os its allies at each position of the missile. I fear it might be too computational heavy to be worth.

PR #591, opened 2018-11-10
upgrade to sfml 2.5

amir-arad:
second attempt (renders #578 obsolete) after shocking discovery of great vagrant build scripts
SFML upgrade only

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,7 +377,7 @@ include_directories(${SERIOUS_PROTON_DIR}/src)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
 ## Setup SFML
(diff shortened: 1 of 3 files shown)


daid:
And there is also no need to update the find_package to 2.5, as the 2.3 line will find a 2.3 version or newer. So it would also work with 2.5 without that. And as the release builds that I do are still on a lower version, this will break my build at the moment.

gwaland:
I was able to verify that I could compile both with sfml 2.3 and 2.5 with the cmake file set as the following: 
find_package(SFML 2.3 REQUIRED system window graphics network audio)
At least on my windows build I did not see any graphics issues on either build using the rest of this patch.

amir-arad:
my idea was that the official EE will upgrade to 2.5, as compiling a new SFML seems pretty feasible. 
But simply making it compatible is also an improvement

daid:
At some point I most likely will, so the fixes are good. But I rather do that at my own phase, then forced by a pull request ;-)

PR #593, opened 2018-11-16
cargo, docks and drones

amir-arad:
this is a first draft for some new concepts I've been working on. 
### Docks and cargo (Inventory system)
docks are containers inside a ship that can hold cargo. 
cargo can be many things: auxiliary reactor, trade goods, ammunition, special buffs or bonuses, go-fetch mission objectives etc. It can fit the latest pick-up feature nicely.
### Drones
Drones are small spaceships that can be remotely controlled by an operator sitting in the mothership's bridge. they are fragile and have no reactor of their own, but they can dock *inside* the mothership (become cargo inside the ship's docks), there they can undergo maintenance (recharge, cool down etc.) and ultimately be launched to space again.
### Player Stations (see under *Extra*)
 two new player stations: the dockmaster and the drones operator. 
drone operator can connect to any drone in range and drive it remotely. 
dockmaster manages the inventory of the ship - move cargo between docks, launch drones, and recharge them (or - use them as batteries for the mothership's energy consumption!)

There's really much more work to be done to make this a final proposal. I know there are many details I haven't yet gotten into, but i wanted to get it out so that i can get some opinions and hopefully good pointers. Currently the only ship I've added docks and drones to is the "Atlantis" corvette


amir-arad:
Rught now there are two kinds of docks:launching and energy.
Launching docks can launch the cargo they hold, and energy docks can connect their cargo to the energy grid of the ship. 
I have thoughts about the dockmaster configuring some pipeline in the docks

Xansta:
This looks like an excellent addition!

src/shipCargo.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    template_name = ship_template->getName();
+    callsign = "DRN-" + gameGlobalInfo->getNextShipCallsign();
+    setEnergy(ship_template->energy_storage_amount);
+    hull_strength = ship_template->hull;
+}
+
+ShipCargo::ShipCargo(P<SpaceShip> ship) : ShipCargo()
+{
+    setEnergy(ship->getEnergy());
```
NinjaSpectre:
```suggestion
    template_name = ship->getTypeName();
    callsign = ship->getCallSign();
    setEnergy(ship->getEnergy());
```

NinjaSpectre: Without `template_name` being set there is a crash when selecting the dock after the drone has docked with the carrier

amir-arad: right. thanks!

amir-arad:
@czenker I'd love to hear your opinion

amir-arad:
@czenker thanks for the kind feedback

you've raised good points. the idea of a big storage bay for virtually unlimited storage is probably something that's going to happen sooner or later. Other suggestions are very interesting and I'll definitely look into them. the numbers (amount of energy, energy consumptions etc.) are indeed very rough yet.

I want to address some points, hopefully not to come off as defensive - just sharing my thoughts.

The drones are (currently) simply `PlayerSpaceShip`s that are not selectable in the main player stations menu. The goal is to increase the number of different "seats" in a single bridge. however, giving the drones some short-term autonomy ("go there", "attack that" commands like the GM has) can significantly increase drone operator's ability to manage multiple drones simultaneously. Incidentally, I want to add ability for a GM to manually operate a CPU ship for a time, so I might look at ways to combine the two.

Instead of replacing probes with drones, I was thinking about turning probes into "hotspots" for the drones connection. so that the ship can leave a probe and two drones to defend a spot, or lay a path of probes to allow drones to approach a target (perhaps a supply-train to recharge and restock the ship from a distant space station)

I hear what you say about the drone management taking over the dockmaster focus, but For me the new game vibe here is the war front logistics, during combat the dock master is mainly in charge of drones delivery and re-entry, and re-charging the ship's energy reserve  (and in the future, torpedoes ammo). I'd love to hear ideas of how to split the responsibilities during battle with another officer. 

I'm planning on significantly simplifying the dockmaster screen, though. however, drones can't really be stacked as counters, as they are individually identifiable, with their own unique properties (model, callsign, health, heat, energy etc.). In addition, I'm planning on having more cargo types that are not drones, which will have heat, health, and energy properties. so that other incoming cargo (i.e loot dropped from destroyed ships) will require repairing, or cooling down in the thermic dock , and possibly might be charged or discharged in the energy dock. the next interesting cargo types I think about are an auxiliary reactor, power cells, ammo box, and different power-ups (super speed, shield recharge etc.) and of course, the goods you have in lively epsilon.

What I can't figure out is a good API for cargo item scripting. I want to allow defining templates for things that can be either cargo or space object, and functions to translate one to the other.

daid:

Issue #543, opened 2018-06-25
Remote GM


k9lego:
How hard would it be to program a remote GM console for headless servers?

daid:
The GM station uses information and functions that aren't replicated across the network. Which is why it's server only, it's not designed like that, so it would be a lot of work.

amir-arad:
what's the scope of the difference?
say, what GM function would be trivial to do remotely? say I was thinking about quickly adding co-GM stations, what could co-GMs do?

daid:
My guess, only the GM comms functions would work right now. Enemy AI, tweaking objects, creating objects, all done server side only. Viewing the whole "world" on a client can be done, sans AI information.

tdelc:
I think that a co-GM station, even only for the communication would be a
very great thing.

k9lego:
I agree with that. GM would be more effective if one person made the story and one managed the enemies.

amir-arad:
I'd postpone the security thing, it's not a blocker IMO. as it is, players either get auto-connecting stations, or are free to choose, so what's the harm of them choosing a co-GM screen instead of any other "wrong" screen?

amir-arad:
re-visiting this discussion, as i expect to need co-GM stations in the foreseeable future.
I would appreciate a review on this idea:
expressing GM actions in  LUA (very small scripts) so that the local GM screen will run them locally, and a remote screen will send them to the server using a single designated GM_ACTION multiplayer command.
putting execution aside, is that approach practical? is there expected to be an issue with drag-and-drop?

PR #596, opened 2018-11-29
(mostly) Refactor gm screen

amir-arad:
This is a refactor of the GM screen - extract inner components `GuiObjectCreationView` and `GuiGlobalMessageEntryView` into their own files (and cut circular dependency  of `GMScreen` and `ObjectCreation` using callback)
In addition, the last commit introduced a GM actions executor singleton. its purpose is to allow remote / multiple GM stations (#543) , by sending GM commands to the server instead of running them locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,8 @@ thumbs.db
 *.swp
 
 options.ini
+
+
+.vscode
+logs
+*.log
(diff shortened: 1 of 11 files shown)


daid:
Ah, now I see what you try to do with the remote GM. I still think there will be a few issues where a remote GM won't have all the features as the local GM. But that might not be a huge issue.

daid:
Good first step :)

JordanLongstaff:
There are new files here. Please don't forget to add them to the CBP file.

PR #597, opened 2018-11-29
Enable GM station(s) from clients

### LINKED ISSUE #543: Remote GM

amir-arad:
enable Gm station from clients
- tweaks and AI infographics does not work on non-server GM stations
fixes #543


src/GMActions.cpp
```diff
@@ -45,6 +48,71 @@ void GameMasterActions::onReceiveClientCommand(int32_t client_id, sf::Packet& pa
...
+            int index;
+            packet >> index;
+            int selectedItemsLeft;
+            packet >> selectedItemsLeft;
+            PVector<SpaceObject> selection;
+            while (selectedItemsLeft--) {
+                int selectedItemId;
+                packet >> selectedItemId;
+                P<SpaceObject> object;
+                if (game_server)
```
daid: this should always be running on the server, so the if (game_server) is alway true

src/gameGlobalInfo.h
```diff
@@ -76,6 +76,8 @@ class GameGlobalInfo : public MultiplayerObject, public Updatable
 
     //List of script functions that can be called from the GM interface (Server only!)
     std::list<GMScriptCallback> gm_callback_functions;
+    //List of names of gm_callback_functions scripts (replicated to clients)
```
daid: I understand why this list is here, but now we have duplicate data, name is in both std::vectors. Would be cleaner to remove the name from the GMScriptCallback (which might not need to exist then and can be just a ScriptSimpleCallback object?)

JordanLongstaff: http://www.cplusplus.com/reference/algorithm/remove_if/

amir-arad: nevermind, I just RTFM-ed.

JordanLongstaff:
I got build errors with this. I think you need to declare `class CpuShip;` in GMActions.h.

Also, you might consider changing CMD_SEND_COMM_TO_PLAYER_SHIP to 0x000A (hex for 10).

JordanLongstaff:
Actually, no, you're just missing the imports of cpuShip.h and wormHole.h.

JordanLongstaff:
Oh, no...I guess you actually do have those imports. Weird that I didn't see them.

Anyway, I tried clicking the "Start Server" button and nothing happened, so I clicked it again and got an assertion error in gameGlobalInfo.cpp at line 10.

JordanLongstaff:
Turns out that I was having some kind of network issue, actually. For whatever reason, EmptyEpsilon wasn't able to open port 35666.

amir-arad:
> Actually, no, you're just missing the imports of cpuShip.h and wormHole.h.

@JordanLongstaff I went over it a few times now and I dont see anything missing. in `GMActions.h` there is no use of CpuShip or Wormhole so no need to declare or import. and in `GMActions.cpp` both `.h` files are included. See [here](https://github.com/daid/EmptyEpsilon/blob/1043a6d1ee41d58dd4ab7e3f8cf259997006a94a/src/GMActions.cpp#L5-L6)

amir-arad:
any chances of this getting merged?

amir-arad:
Hi @daid I heard you might have missed some github notifications so i'm taking the chance of bugging you.
I'm lazily working on other stuff that is based on this PR (lower level GM control over cpu ships). it would be simpler for me to know if this PR is/isn't gonna make it in. I respect specific rejections or fix requests, as well as principal "feature not worth the complexity".

amir-arad:
ping

daid:
Oops? :-)

Code wise, looks quite fine. Only thing I'm wondering about is how well the GM to ship chat works with multiple GMs connected. As that code is already fragile with just a single GM screen...

amir-arad:
@daid could you elaborate or suggest scenarios?
I did some sanity tests with 2 GM stations and 1 player ship but it was a while back. seemed to be working smoothly except for the UX for the GMs.
I was thinking about a switch for selectively enabling chats with specific ships in each GM station, so that any ship only talks with one GM at any given time. alternatively a GM comms station that only does chats. but that was for internal GM crew management purposes. I would love to hear more about the engineering concerns.

daid:
Just a concern, nothing concrete jumps to mind, just something that I wondered how well you checked.

amir-arad:
I didn't test it in "production" yet. I run 0-2 games a year, and this year was dry so far :(

k9lego:
@amir-arad Daid mentioned on the USN that he's concerned about security issues relating to this. I could test this in a production game on the 15th if you could get this secure.

daid:
By security I mean, it allows any client to do GM actions. And it allows and client to execute lua code on the server.

This could be fixable by making the GM client password protected, and setting a flag on the PlayerInfo object as soon as the correct password was entered, and checking this flag before and GM related function is executed on the server.

amir-arad:
I've used this feature in a 40 hours larp so I can say it works in production. 

My upcoming game is a few weeks from now so i won't be available anytime soon to add the security feature for the sake of merging.

k9lego:
Sounds good!

Kilted-Klingon:
I would just like to add my 2 cents here in this thread -- I also would find it very useful to have the option to choose a GM station from a client in order to spread the GM responsibilities among a team.  I have tested one of @amir-arad versions, and it seems to work fine.  However, I would definitely want to have some sort of security password enabled through the scenario script.

Kilted-Klingon:
@amir-arad , @daid , and @k9lego , I volunteer to help test capabilities.  Now that I can build both Linux and Windows versions from the GitHub repositories, I certainly don't mind helping out.

amir-arad:
work continues on #965

PR #600, opened 2018-12-17
extract ShipFilter from AutoConnectScreen

amir-arad:
a small refactor to extract a potentially reusable ship filter from auto-connect.

My next goal would be extending hardware.ini to support more than one ship (following [this](http://bridgesim.net/discussion/422/dmx-on-multiple-bridges-setup) thread)


amir-arad:
redundant by #601

PR #601, opened 2018-12-17
Multi ship hardware

amir-arad:
support configuring multiple ships in hardware.ini by using the ship filter from autoconnect


amir-arad:
sorry, haven't tested it yet. will re-open after testing.

PR #603, opened 2018-12-21
proposed fix to system degrade on low energy

amir-arad:


daid:
Not sure how the old logic was broken?

amir-arad:
currently, when the ship's energy is low, the power is *inverted* (`power = 1/power`) and multiplied by a factor of the energy. The inversion part struck me as a human error.

daid:
You're right :-) good catch!

Issue #584, opened 2018-10-30
Scenario 08 atlantis problem in mission state


Fank:
I rechecked the code and it looks like we stuck here, because we never triggered it:
https://github.com/daid/EmptyEpsilon/blob/1716e22743101136c289947ef8e267c520411c3e/scripts/scenario_08_atlantis.lua#L413-L419

Because we went through the wormhole, in to the territory with a lot of Kraylor and then we headed down/south for about 25-30U to get out of the enemy territory. Then continue to fly to B20 to reach the jump carrier. After we were next to it we could dock it at (B20), but it did nothing. Headed back to the HQ, still no option to do something.

daid:
That scenario indeed still has a bunch of bugs. I pushed the notes from my latest play session:
https://github.com/daid/EmptyEpsilon/commit/1716e22743101136c289947ef8e267c520411c3e

So there are more things still wrong with it.

Fank:
Some background information, we docked to the jump carrier for the first "jump" to B20 and undocked because of some stupidness and waited to return, after it was back we docked again and jump with the carrier to B20. Everthing worked fine here i wasn't sure if this info may helps.

oznogon:
Did PR #608 fix this?

Fank:
I haven't checked it yet.

PR #608, opened 2018-12-29
Fixed potential stuck place identified in issue #584...

Xansta:
...by broadening the trigger condition for bringing the jump ship back. Also set warp jammer ranges to 18000 to cover entire area surrounded by Kraylor since default range (~7000) insufficient. Added some reputation throughout. Corrected spelling and grammar inconsistencies.


Xansta:
It looks like it prevents you from jumping out (ie initiating a jump), but not jumping in or through. We played with a warp drive ship last night and the warp was stopped right at the edge of the warp jammer field. It could be that the code handles warp differently than jump. It could also be that the jump happens so fast (between tics) that the jammer cannot pick them up fast enough. 

For this scenario, increasing the jammer field range seems to have solved the problem.

PR #609, opened 2019-01-02
Better support for headless server in Defender Hunter Scenario

Xansta:
When there is no pause available (like when using a headless server), certain functions need nil checks:
   moving nebulae
   threaded pursuit
   artifact for wormhole in motion
Handle nil selection of random station in transport generation routine
Check for the presence of a spawned ship before proceeding with the first mission instead of just relying on the pause mechanism.


daid:
This also included your warp jammer range fix, but I agree with that as well :-)

Issue #587, opened 2018-11-02
More Event Callbacks for scripting


czenker:
After having the event callback for Artifacts introduced in #586 I would like to add more of them.

## 1. What are you missing for your missions?

I'm thinking of 

* ~`SpaceObject` destruction~ #612
* ~`SupplyDrop` pick up~ #611 
* ~`WormHole` when anything jumps~ #587
* ~`ScanProbe` being send~ #735
* ~`ScanProbe` self-destructs~ #735
* ~`Mine` explosion~ #713
* ~`Asteroid` destruction~ fe5eabe99681b938b845768b380c82eb2495014d

Anything else you are missing?

## 2. Naming things

I've been thinking on what naming scheme to use for the Lua API.

* `SpaceObject:setPickUpFunction(function() end)`
* `SpaceObject:setPickUpCallback(function() end)`
* `SpaceObject:onPickUp(function() end)`
* `SpaceObject:onPickUpFunction(function() end)`
* `SpaceObject:onPickUpCallback(function() end)`

Personal I would prefer the third version, because it is brief and all events are in one block in the documentation. Only issue I see is that `onPickUp` could be mistaken for a property rather than a function by a scripter.

What do you think?

daid:
The 3th format is also what I like the most. It reads the most natural.

And as most people look at examples on how to use things, I don't think the mistaking it for a property will be a huge risk.

The reason these callbacks didn't exists before, was because the ScriptSimpleCallback was a pretty late addition to the code. And the older callback mechanism had some serious flaws, so I didn't use it much.

amir-arad:
the 3rd format is used in EE gui elements for the handling of gui events. 
so if in space objects the same naming convention will have close but different semantics, it may be confusing.

I am missing global events for objects creation in this suggestion. something like `onNewSpaceShip(function() end)` where you can hook a function that registers all other relevant hooks for spaceships.

czenker:
I think I'll leave it at these three PRs for now. I don't find the other things mentioned in my opening post not as useful and currently have no plans on using those. But in case anyone finds them useful let me know.

> I am missing global events for objects creation in this suggestion. something like `onNewSpaceShip(function() end)` where you can hook a function that registers all other relevant hooks for spaceships.

[...]

PR #611, opened 2019-01-05
rename onPickUp callback for Artifact and add it for SupplyDrop

czenker:
I renamed the callback function for Artifact pickup to have a consistent naming as described in #587 and switched the parameter order. I found this more intuitive as lua functions usually have `self` as the _first_ argument. So this is a *breaking change*, but it should not have any real impact as the function was only introduced recently and is not used yet in any official missions.

This commit also adds the same functionality for SupplyDrops.
    
Usage:

    function init()
        PlayerSpaceship():
            setTemplate("Piranha"):
            setFaction("Human Navy"):
            setCallSign("Thief")
        Artifact():
            setCallSign("Stack of Gold"):
            onPickUp(function(foo, bar)
                print(bar:getCallSign() .. " picked up " .. foo:getCallSign())
            end)
        SupplyDrop():
            setCallSign("Ammo Crate"):
            setFaction("Human Navy"):
            onPickUp(function(foo, bar)
                print(bar:getCallSign() .. " picked up " .. foo:getCallSign())
            end)
    end

### DIFF
--- a/src/spaceObjects/artifact.cpp
+++ b/src/spaceObjects/artifact.cpp
@@ -19,8 +19,8 @@ REGISTER_SCRIPT_SUBCLASS(Artifact, SpaceObject)
     /// Set if this artifact can be picked up or not. When it is picked up, this artifact will be destroyed.
     REGISTER_SCRIPT_CLASS_FUNCTION(Artifact, allowPickup);
     /// Set a function that will be called if a player picks up the artifact.
-    /// First argument given to the function will be the playerSpaceShip, the second the artifact.
-    REGISTER_SCRIPT_CLASS_FUNCTION(Artifact, setPickUpCallback);
+    /// First argument given to the function will be the artifact, the second the player.
+    REGISTER_SCRIPT_CLASS_FUNCTION(Artifact, onPickUp);
 }
 
 REGISTER_MULTIPLAYER_CLASS(Artifact, "Artifact");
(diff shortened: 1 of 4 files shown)


Xansta:
I tried using this in a script two different ways and I got this error:
??[convert<scriptsimplecallback>::param] upvalue 1 of function is not a table
I did it with an inline function like the usage sample and I tried it in a separate function
```
	debris1 = Artifact():setPosition(908020, 150504):setModel("ammo_box"):allowPickup(true):setScanningParameters(2,1):onPickUp(pickupDebris1)
...
function pickupDebris1(debris, pGrab)
	pGrab.debris1 = true
end
```
Suggestions?

czenker:
> Suggestions?

Hi Xansta.

I know the fix to this issue. Its nothing specific to this callback, but seems to be a general issue with functions that are used as parameters for the Lua API (like Comms Scripts). The fix is simple and strange: You need to have a reference to the global scope inside your function. So if you write it like this it should work:

    function pickupDebris1(debris, pGrab)
        string.format("")
        pGrab.debris1 = true
    end

This is a somehow weird fix, but according to the comments SeriousProton needs that global scope in order to check if the script that function belongs to has already been destroyed. Here is the code https://github.com/daid/SeriousProton/blob/444440d8a0fec3a77c9236911f652acad5d4bc8e/src/scriptInterface.cpp#L643-L649 
So this looks lika a bug, but I do not see an easy fix there.

daid:
There isn't an easy fix, due to how the scripting environment is set up.

PR #612, opened 2019-01-05
add eventListeners for shipTemplateBasedObjects and warpJammers being damaged or destroyed

czenker:
EventListeners for when a station/ship/player/warp jammer take damage or are destroyed by taking damage. #587 

`onDestruction` will intentionally not be triggered in cases where the ship is not destroyed by taking damage. This could be scripts calling `destroy` directly, the GM console or some quirks like black holes. The black hole behavior should probably be fixed and there might be other edge cases.

Would be happy for feedback if this behavior is expected by you or if a change is required. `onDestructionByDamage` would be probably be a more accurate name, but is very cumbersome.

Also currently there is no instigator given if the damage is caused by collision with an asteroid. This could be fixed in the future also.

Usage:

    function init()
        local kraylor, station, ship, player
    
        kraylor = CpuShip():setFaction("Kraylor"):setTemplate("Odin"):setCallSign("Odin"):setPosition(1000, 0):orderRoaming()
    
        local source = function(instigator)
            if type(instigator) == "table" and type(instigator.getCallSign) == "function" and instigator:getCallSign() ~= "" then
                return " by " .. instigator:getCallSign()
            elseif type(instigator) == "table" and type(instigator.type) == "string" then
                return " by a " .. instigator.type
            else
                return ""
            end
        end
    
        local onTakingDamage = function(self, instigator)
            print(self:getCallSign() .. " is taking damage" .. source(instigator))
        end
        local onDestruction = function(self, instigator)
            print(self:getCallSign() .. " was destroyed" .. source(instigator))
        end
    
        SpaceStation():
        setTemplate("Medium Station"):
        setCallSign("Dummy"):
        setFaction("Human Navy"):
        onTakingDamage(onTakingDamage):
        onDestruction(onDestruction)
    
        CpuShip():
        setTemplate("Adder MK4"):
        setCallSign("Ship"):
        setPosition(0, 1000):
        setFaction("Human Navy"):
        onTakingDamage(onTakingDamage):
        onDestruction(onDestruction):orderIdle()
    
        PlayerSpaceship():
        setTemplate("Flavia P.Falcon"):
        setCallSign("Player"):
        setPosition(0, -1000):
        setFaction("Human Navy"):
        onTakingDamage(onTakingDamage):
        onDestruction(onDestruction)
    
        WarpJammer():
        setCallSign("Jammer"):
        setPosition(0, -2000):
        setFaction("Human Navy"):
        onTakingDamage(onTakingDamage):
        onDestruction(onDestruction)
    
    end

### DIFF
--- a/src/spaceObjects/shipTemplateBasedObject.cpp
+++ b/src/spaceObjects/shipTemplateBasedObject.cpp
@@ -60,6 +60,12 @@ REGISTER_SCRIPT_SUBCLASS_NO_CREATE(ShipTemplateBasedObject, SpaceObject)
     REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, setRearShield);
     /// [Depricated]
     REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, setRearShieldMax);
+    /// Set a function that will be called if the object is taking damage.
+    /// First argument given to the function will be the object taking damage, the second the instigator SpaceObject (or nil).
+    REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, onTakingDamage);
+    /// Set a function that will be called if the object is destroyed by taking damage.
+    /// First argument given to the function will be the object taking damage, the second the instigator SpaceObject that gave the final blow (or nil).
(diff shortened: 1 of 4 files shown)


amir-arad:
I like this. Didnt plan on it but scripting power is great

daid:
Not getting a callback on a blackhole sounds like a bug indeed, but minor.

PR #616, opened 2019-01-16
Switch android to gradle project

miketweaver:
I've been working on getting the android build working. The first step to bring this project into 2019 is to switch the android builds to gradle builds. (SFML has done the same)

Things to note:
- TargetSDK has switched to 26
- For my testing I have dropped `armeabi` , `x86`, and `mips`. I am only building for `armeabi-v7a`

Here are the commands I used to create a build environment on Ubuntu 18.04:
https://gist.github.com/miketweaver/cc8165e1e889fd20bfb37192d248cfa2

Currently the build is **FAILING** due to some compilation errors. I assume this is do to the switch from GCC to CLANG. (GCC has been removed from Android NDKs since September)


miketweaver:
I could use some help with the compilation errors:
```
  /home/user/android/EmptyEpsilon/src/hardware/serialDriver.cpp:10:67: error: expected function body after function declarator
      extern int ioctl (int __fd, unsigned long int __request, ...) __THROW;
                                                                    ^
  /home/user/android/EmptyEpsilon/src/hardware/serialDriver.cpp:11:55: error: expected function body after function declarator
      extern int tcsendbreak (int __fd, int __duration) __THROW;
                                                        ^
  /home/user/android/EmptyEpsilon/src/hardware/serialDriver.cpp:427:5: error: use of undeclared identifier 'tcsendbreak'
      tcsendbreak(handle, 0);
      ^
  3 errors generated.
  make: *** [/home/user/android/EmptyEpsilon/android/app/build/intermediates/ndkBuild/debug/obj/local/armeabi-v7a/objs-debug/emptyepsilon//home/user/android/EmptyEpsilon/src/hardware/serialDriver.o] Error 1
```

daid:
Ah, the spawn of Satan called gradle.

The compile fix is quite easy, the code for the serial ports will never work on android, so just conditionally not compile it on android
See: https://github.com/daid/SeriousProton2/blob/master/src/io/serialport.cpp

miketweaver:
Ok. Thanks to @oznogon, I realized I had copied some of the resource files into the wrong location.

I fixed the instructions on the GIST, and it now compiles and loads:
image

This will take some time to test and ensure full functionality. I also would like to move the resource copy to the `build.gradle` file before this gets murged

daid:
Note that I will merge this once you get a proper test, but I won't be able to actually build it. Due to how to gradle works, my build server runs out of disk space when I use gradle.

I am experimenting building without gradle, but there I get a segfault without a stacktrace at the moment.

miketweaver:
Well, the primary goal was to build something for you @daid, so that we could put Android back into the regular release cycle using a current NDK. If gradle won't work for you (need me to ship you an SSD?), I'll need to switch to something that will.

Any suggestions on where to take this now that both
- android update project
- and (Ant)

have been deprecated from the current NDK & SDKs?

----

At the same time, after some additional testing I get errors loading this on older APIs (less than Android 7.0)

EE:
```
01-17 23:33:59.570 24286 24286 E AndroidRuntime: Caused by: java.lang.IllegalArgumentException: Unable to load native library: /data/app/eu.daid.emptyepsilon-2/lib/arm/libsfml-activity-d.so
```

I get the same error when I build the SFML example, so it's something to do with the build process for these older devices.

SFML:
```
01-18 10:43:57.752 20069 20069 E AndroidRuntime: Caused by: java.lang.IllegalArgumentException: Unable to load native library: /data/app/org.sfmldev.android-1/lib/arm/libsfml-activity-d.so
```

I'll see if I can clear this issue up.

daid:
Oh, I have the diskspace, gradle is just claiming space on the smaller partition. It's also dog slow (the build machine isn't that fast to begin with)

Note that you can still download older NDK versions, I'm using the r17b NDK version for my current experiments, as the current released r18 dropped support for my phone.

Xansta:
I'll be so happy when a build for Android is back. I'll be able to put by tablet back into service as a station for any visiting guests. A couple of them brought their tablets remembering that we used to play with tablets, but I had to tell them that we could not at the moment use tablets. They did not go away without playing - we had plenty of computers to go around.

Xansta

miketweaver:
I'm fairly certain that the issues I'm having with running on an older android API are a SFML issue, not an EE issue. I have the same problem with the SFML example project when targeting the latest SDK required by Google Play. 

https://github.com/SFML/SFML/pull/1545#issuecomment-45999370

I believe the issue lies with the SFML android dependencies targeting a different SDK than what we target now. I'll need to fix the SFML android dependencies before we can get EE working on the current NDK with older devices.

At the same time @daid, r18 did drop support for android 15 (4.0.3). If you need a newer device and are willing to resume Android support, let me know and I'm sure I can get a hold of one for you.

For the gradle stuff, have you tried moving the gradle tempdir?
`export GRADLE_OPTS=-Djava.io.tmpdir=/path/to/tmpdir`
or
`-Djava.io.tmpdir=/somewhere/else`

daid:
On the good news, my old phone died, and I checked, my replacement is running 4.3 (API level 18)

I'm willing to spend some time on getting my build server to run with gradle. But I couldn't find myself to find the time to figure out how to get gradle working, and at the same time fixing the EE Android build issues.

(Publishing on Google Play isn't really a priority for me)

daid:
As I have a new build up and running with android, without the need for gradle. I'm closing this.
See: https://github.com/daid/EmptyEpsilon/tree/master/android

PR #617, opened 2019-01-22
Option to toggle captions in the window screen.

aBlueShadow:
While the window screen is probably rarely used, for obvious reasons (needs decent resources, and a spare laptop usually doesn't look good as a window), I think it has the potential to increase the immersion of being on a starship.

However, till now it is not really a window, showing heading and callsigns.
With this PR, you can choose whether you want those captions in the ship window screen or not.
By now you can only toggle both headings and callsigns together, with space dust always on, as I think those are the most obvious options.
However, the code is already prepared to have any combination of those three, so later additions would be pretty easy.

### DIFF
--- a/src/menus/shipSelectionScreen.cpp
+++ b/src/menus/shipSelectionScreen.cpp
@@ -77,31 +77,33 @@ ShipSelectionScreen::ShipSelectionScreen()
         window_button->setValue(false);
         topdown_button->setValue(false);
         cinematic_view_button->setValue(false);
+        window_options_label->hide();window_angle->hide();window_caption_button->hide();
     });
     game_master_button->setSize(GuiElement::GuiSizeMax, 50);
 
     // Ship window button and angle slider
-    window_button_row = new GuiAutoLayout(stations_layout, "", GuiAutoLayout::LayoutHorizontalLeftToRight);
-    window_button_row->setSize(GuiElement::GuiSizeMax, 50);
-    window_button = new GuiToggleButton(window_button_row, "WINDOW_BUTTON", "Ship window", [this](bool value) {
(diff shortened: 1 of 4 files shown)


daid:
IMHO, this patch should be much simpler, the "window" screen should never show callsigns or headings, no need for that as an option (less=more)

aBlueShadow:
Oops - Okay, in this case I should probaby just scrap this PR and make a new one, as this one mainly consists of those options.
I was'nt sure If the current behaviour was intended or used by someone, and multiple mainscreens won't work as they would be simply mirrored. So I wanted to be on the safe side. But I guess this is not really necessary, because if you want to know if a specific ship is outside the mainscreen at a specific angly, you have the radar view for that.

aBlueShadow:
@daid : as mentioned by @PlanetariumWSD in #621, the shipwindow screen is used as additional mainscreen by some, and having labels make sense there. So can we use this PR instead as there is a usecase? Alternatively, I could do a lighter PR, that just uses an options.ini switch. Or a hotkey, that toggles between labeling options in-game.

daid:
But a first person main screen is available already, if you press "F" on the main screen...

aBlueShadow:
Sure, but the emphasis on my sentence lies on "additional". 

On #621, @PlanetariumWSD was mentioning that they run EE in a planetarium, using multiple shifted window screens as a mainscreen to cover the dome. The mainscreen cannot be used to achieve this, as you can only rotate 90-degree-wise and, more importantly, the output would be simply mirrored.

Of course this is a pretty specific usecase, but I gues there could be other usecases where someone wants to build a huge viewscreen. 
As mentioned, if you think it is too complex for what it does, I can implement it in a more subtle way, like adding a hotkey or an options.ini switch.
In the case you don't like one of the solutions I suggested, I would rather suggest to revert my PR. As I don't want to be responsible that a planetarium's performance is impacted because of it.

PR #619, opened 2019-01-23
allow scripts to set fine grained scanState

czenker:
Right now it is only possible to set the scan state to `full` or none at all through scripts. (#146) This pull request allows to set all ScanStates. 

Usage:

    local ship = CpuShip():setTemplate("Adder MK5"):setFaction("Independent"):setPosition(1000, 0)
    ship:setScanStateByFaction("Human Navy", "simple")
    PlayerSpaceship():setTemplate("Phobos M3P"):setFaction("Human Navy")
end


czenker:
@daid What do you think of this merge request?

I can think of lots of uses for scenarios, like a station calling for help and FoF-identifying the attackers. Having the fully scanned would be helping the crew too much.

daid:
As the other scan states only make sense on spaceships, i think the fine grade functions shpuod be there. The functions to check these states are also in the spaceship class.

czenker:
It took me a while to get back to it, but here is the reworked commit that only adds these functions to `SpaceShip`. :)

PR #620, opened 2019-01-23
Fix the mod option

czenker:
The `mod` option did not work as I was expecting it. Starting EE like

    ./EmptyEpsilon mod=my_mod

did not use any files from the mod. And even though scenarios from the mod where found they could not be started. The simple bugfix to restore that functionality is in comit f5cc8a5.

I think that this resulted in the mod feature not being used by anyone<sup>[citation needed]</sup>. So I thought of forcing the `script`, `resources` and `packs` directories to the mods too. This would be a breaking change, but I'm not sure if anyone would be affected by it. But I do not see why scripts and models should be stored in the same folder for mods. This is the second commit 29ad5ec.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -120,11 +120,11 @@ int main(int argc, char** argv)
         string mod = PreferencesManager::get("mod");
         if (getenv("HOME"))
         {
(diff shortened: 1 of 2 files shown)


daid:

PR #621, opened 2019-01-26
Remove headings and callsigns from windowScreen

aBlueShadow:
So this lets the ship window look more like a window, as it removes the callsigns and headings.
I kept the spacedust though, while not really realistic it maintains a sense of movement. But of course this could be removed as well if you disagree here.
This replaces #617 - As without the options it's just a oneliner, doing a new PR was much easier than trying to reduce that other one.

### DIFF
--- a/src/screens/windowScreen.cpp
+++ b/src/screens/windowScreen.cpp
@@ -12,7 +12,7 @@ WindowScreen::WindowScreen(float angle)
 : angle(angle)
 {
     viewport = new GuiViewport3D(this, "VIEWPORT");
-    viewport->showCallsigns()->showHeadings()->showSpacedust();
+    viewport->showSpacedust();
     viewport->setPosition(0, 0, ATopLeft)->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
 
     new GuiShipDestroyedPopup(this);


daid:
And indeed, the whole point of the spacedust is to create a sense of movement. It's very unrealistic. But it enhances the feeling of speed.

PlanetariumWSD:
Reiterating from #636 
Having optional numerical headings on ship windows can be beneficial to game play as a kind of high tech overlay instead of simple glass.
It may seem that multiple ship's windows are not used often but... when I was giving a paper on Empty Epsilon at a planetarium conference in St. Louis a couple years ago, a delegate from San Francisco asked 'besides being a soft skill educational cooperative space themed experience, what relevance it had in the dome environment.' (the reason we gather) When I said that you can create large cohesive pans all the way around the audience.. his eyebrow raised and I think this, among many other features really sets Empty Epsilon "ahead of the pack" when thinking about what's available. i.e. some domes are playing Artemis without this option.

aBlueShadow:
Okay, thats interesting. Havent thought this view would be used much for gameplay.
The motivation why I did that PR was not gameplay, but immersion. And immersion-wise, for a window that is actually used as a window, a hud would be detrimental. I.e in the ship lounge, you would probably enjoy the view, not have it obfuscated by numbers and callsigns.

Originally I made a PR that made the labels on the ship window configurable in the station selection screen(#617 ), I was asked to make it much simpler, and as noone objected to it, I did so.
So this question is, now that we know about other usecases for ship window, would It be okay for you daid, to actually use #617 after all? As a compromise I could implement it as an option in config.ini instead of the GUI toggle.

PR #632, opened 2019-02-10
2 small fixes

amir-arad:
- fix broken cmake build
 - make planets visible to relay

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -355,6 +355,7 @@ foreach(SP_SOURCE
 	src/lua/ltablib.c
 	src/lua/lgc.c
 	src/lua/lbitlib.c
+	src/json11/json11.cpp
 	)
 	list(APPEND SOURCES ${SERIOUS_PROTON_DIR}/${SP_SOURCE})
 endforeach()
--- a/src/hardware/devices/philipsHueDevice.cpp
+++ b/src/hardware/devices/philipsHueDevice.cpp
@@ -4,6 +4,7 @@
 #include "hardware/serialDriver.h"
 #include "logging.h"
 #include <unistd.h>
+#include <json11/json11.hpp>
 
 using namespace json11;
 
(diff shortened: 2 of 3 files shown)


NinjaSpectre:
I had to alter this code block in order to get a working compile.
https://github.com/daid/EmptyEpsilon/blob/9d0837efcee906991c8c305709acb9349d0d2f89/src/hardware/devices/philipsHueDevice.cpp#L143-L149 To the following:
```c++
int currentInt = 0;
light_count = 0;
std::map<std::basic_string<char>, json11::Json> jsonMap = hue_json.object_items();
for(std::map<std::basic_string<char>, json11::Json>::iterator it = jsonMap.begin(); it != jsonMap.end(); it++) {
    std::stringstream(it->first) >> currentInt;
    LOG(DEBUG) << "Got key from Hue API " << currentInt;
    if (currentInt >= light_count) light_count = currentInt;
}
```

Kilted-Klingon:
This is my first EE GitHub comment.  Thanks for your patience.

amir-arad made this mod based on my request for a change to the behavior in the Relay Console map.  In doing the change, he compiled a version separately and sent it to me to take a look at.  I've run the version he gave me, and it appears to run properly.  

To succinctly summarize the change, planets now appear on the Relay Console map, but there are two important aspects to keep in mind, as follows:

1:  It works best if all the planets in a scenario script are on the 'nominal' orbital plane.  It appears that if there's not enough planetary mass within a certain range of the plane then the planet doesn't show up on the Relay Console map.  This is easily mitigated by using 'setDistanceFromMovementPlane(0)' for each planet in the scenario script and then there's no problem (maybe that's the default anyway?).
2:  Each planet has to have some atmosphere color set.  This is not a problem with 'normal' planets that have atmosphere color and texture set, but what if you want a moon without atmosphere?  The solution is very simple.  Just 'setPlanetAtmosphereColor(0.2,0.2,0.2)' and this gives the moon a 'just right' gray coloring that shows up on the Relay Console just like a moon should.  Voila!
(If no color is set, then the planet/moon shows on the map as a kind of stealth black, which could have it's own useful purposes!)

Otherwise I think this is a highly useful modification to the base game and recommend it be merged into the master.

Kilted-Klingon:
Woooooo-hooooooo!   Thanks to both!  Very much appreciate the support!

PR #635, opened 2019-02-26
Audio edit for What the dickens

aBlueShadow:
This modifies the audio for the What The Dickens scenario:
- Apply filter effects
- Replace wav by ogg files


daid:
That saves a lot of data :-)

PR #650, opened 2019-05-30
Config for joystick axes

aBlueShadow:
This adds some configurability for the joystick axes. It is pretty simple, we still only have 4 axes and four actions, but now you can freely assign in options.ini which action will be triggered by which axis, or deactivate one if needed.

They are two minor drawbacks to keep this PR relatively simple:
- The deadzone that is currently in there for vertical combat maneuvers, will now also be used for horizontal ones.
- The handle in the combat maneuver box will not moved when the CM is made by joystick. This is already the case on singlepilot and tactical, and will now also be true on helm. But I think the moving joystick itself will be enough feedback anyway

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -163,6 +163,18 @@ int main(int argc, char** argv)
     colorConfig.load();
     hotkeys.load();
 
+    //set joystick defaults to formerly hardcoded settings
+    
+    if (PreferencesManager::get("joystick_combat_strafe").empty())
+        PreferencesManager::set("joystick_combat_strafe", "4");
+    if (PreferencesManager::get("joystick_combat_boost").empty())
+        PreferencesManager::set("joystick_combat_boost", "2");
+    if (PreferencesManager::get("joystick_rotate").empty())
+        PreferencesManager::set("joystick_rotate", "1");
+    if (PreferencesManager::get("joystick_impulse").empty())
+        PreferencesManager::set("joystick_impulse", "3");
+                
+    
     if (PreferencesManager::get("headless") == "")
     {
         //Setup the rendering layers.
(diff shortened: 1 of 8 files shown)


amir-arad:
FYI if you're already touching the joystick controls:
 daid/SeriousProton#29 allows registering for all joystick 0 events (all axises and all buttons supported by sfml) directly via the input handler by creating a `JoystickEventHandler`. the deadzone is the same for all axises. 
might leverage the configurability even more

aBlueShadow:
Interesting. Might try to have a closer look, at a first glance i am not sure how to use it exactly (and could not find an example implementation on your fork either).
That said, I do not intend to make a comprehensive joystick config. (In fact, I hesitated to do this PR for far over a year now, as a user mentioned to plan a major input rework and I did not want to interfere.)
Basically this is just supposed to be a quick fix for situations that could not be circumvented by system settings or other tools, like joysticks with less than 4 axes and weird signaling/drivers.
As buttons can be handled fine by tools like joyt2key (unlike the axes, as there are no combat maneuver hotkeys), the axes are first priority.

btw about the deadzone: If I am not mistaken, there was a builtin deadzone anyway, but vslotman, the contributor of the original joystick code enhanced it, as it was assumingly needed:

> // Add **some more** hysteresis, since y-axis can be hard to keep at 0

As combat maneuvers tend to be hectic and not very suble  anyway, I figured that it won't hurt if I use this enhancement also for x-movement

amir-arad:
I hear you.
For future reference: [gameMasterScreen](https://github.com/amir-arad/EmptyEpsilon/blob/04ee962f016b0fa15629fa11cadce3b4a9ea9d31/src/screens/gm/gameMasterScreen.h#L26) inherits from it, implementing both [handleJoystickAxis](https://github.com/amir-arad/EmptyEpsilon/blob/04ee962f016b0fa15629fa11cadce3b4a9ea9d31/src/screens/gm/gameMasterScreen.cpp#L499) and [handleJoystickButton](https://github.com/amir-arad/EmptyEpsilon/blob/04ee962f016b0fa15629fa11cadce3b4a9ea9d31/src/screens/gm/gameMasterScreen.cpp#L522). it's a bit rough around the edges but it works great.

amir-arad:
I've re-written this PR in #656

aBlueShadow:
Closed, as the other PR was merged, and I only left this one open for reference/fallback.

PR #654, opened 2019-06-22
Somewhat Subtle Change to Science Scanning.

JonathanILevi:
I adjusted how locking works for Science's scanning dialog.

        Before my change, after the Science Officer got the frequency to lock, there was a 2 second delay before scan was complete.  I adjusted it so that when the Science Officer gets the frequency to be more accurate the delay is shorter, and when it is less accurate the delay is longer.  When the Science Officer touches a slider the delay gets reset, giving the Officer a choice of whether to move it closer or wait for the slower delay.  If the Officer happens to get the frequency to perfectly match then the delay is only 1/2 second.

        The new system is very similar to how it was; many players will not notice the difference right away.  The new system is more accessible to new players or players who have more difficulties with precision, because they don’t need to be as accurate in order to scan, at the cost of a longer delay.  The new system also gives veteran players that small improvement they can work for, helping them continue to enjoy the game.

        To be more detailed:  I divided the old "locking" delay into two parts: honing, and locked.  “Honing” is when the display is getting more accurate.  “Locked” is after the wave has turned  white and is always a 1/2 second delay.  The frequency is technically always honing but resets whenever the player is holding a slider.  Although the sensor is always honing, it will take 44 days if the frequencies are as mismatched as they can get.  Through my testing, scanning takes roughly the same amount of total time as before my change.
        The curve I created was specifically designed to have a soft corner at 0.05 (the old lock distance) so as to be similar.  Less than 0.05 gets faster gradually.  More that 0.05 gets slower very quickly.  [Here](https://www.desmos.com/calculator/creecu5fqo) is the curve I used; x is closeness, y is delay time (in seconds).

Tl;dr: When the scanning lock is more accurate the delay is shorter, and when it is less accurate the delay is longer.

### DIFF
--- a/src/screenComponents/scanningDialog.cpp
+++ b/src/screenComponents/scanningDialog.cpp
@@ -3,8 +3,21 @@
 #include "playerInfo.h"
 #include "gui/gui2_panel.h"
 #include "gui/gui2_label.h"
-#include "gui/gui2_slider.h"
 #include "gui/gui2_button.h"
+
+
+bool FrequencySlider::onMouseDown(sf::Vector2f p) {
+    GuiSlider::onMouseDown(p);
+    mouseDown = true;
+    return true;
+}
+void FrequencySlider::onMouseUp(sf::Vector2f p) {
+    GuiSlider::onMouseUp(p);
+    mouseDown = false;
+    mouseHadClick = true;
+}
+
+
 
 GuiScanningDialog::GuiScanningDialog(GuiContainer* owner, string id)
 : GuiElement(owner, id)
@@ -29,7 +42,7 @@ GuiScanningDialog::GuiScanningDialog(GuiContainer* owner, string id)
     
     for(int n=0; n<max_sliders; n++)
     {
-        sliders[n] = new GuiSlider(box, id + "_SLIDER_" + string(n), 0.0, 1.0, 0.0, nullptr);
+        sliders[n] = new FrequencySlider(box, id + "_SLIDER_" + string(n), 0.0, 1.0, 0.0, nullptr);
         sliders[n]->setPosition(0, 200 + n * 70, ATopCenter)->setSize(450, 50);
     }
     cancel_button = new GuiButton(box, id + "_CANCEL", "Cancel", []() {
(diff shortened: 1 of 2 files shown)


JonathanILevi:
@daid Do you have any thoughts on this?

daid:
My concern with this is that players won't notice that they have to be more accurate for faster results and thus actually wait longer all the time, as the "be more accurate" is non-obvious, and could potentially feel quite random when it is suddenly faster then other times.

JonathanILevi:
Thanks for your reply.  You make a good point, Daid.  After some musing, I do not expect it will be a problem, here is why.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I think it is rather intuitive that when the waves are closer it would take less time to hone in on the lock.  I also think that many players (such as myself) originally figured it did.  Regardless of a player’s expectation, I do not think the change is fighting against intuition.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The shape of the curve keeps it from being a real problem. If you are close enough to see the wave honing in, 90% of the time, the delay will be under 4 seconds and 50% of the time, the delay will be under 2 seconds.  This is because the curve (as described) has that soft corner.  In that 10% of the time when you notice the wave moving but it has quite a long delay, the delay is not that extreme only getting 10 seconds.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Both the speed of the honing and the mismatch of the waves is quite easy to see.  Because of this, I think the correlation will not be hard to spot.  It might take some new players a couple hours of play before they realize that the accuracy of the waves controls the speed of the auto-honing.  New players expect to not know everything and the sensation when they figure it out will be quite rewarding as opposed to disappointing.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I would love to test this feature and see what some players think of it and also what the response is for players who have not had it explained.  I think the only way to do this is to release it and hear back some feedback.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I guess to sum it up, I think your concern is valid although minor.  I think the reward and effect of the change on players outweighs the difficulties.  It can give new players the ability to find improvements that they can be proud of and gives veteran players the ability to continue to improve.

JonathanILevi:
Would making a setting by where you set scanning complexity for "experimental scanning" be a good idea?  Releasing this feature idea under that "experimental" flag.

I think it would allow for valuable experimenting as well as time to tweak it.  I think there are a couple tweaks that can be found to remove the concern of uninformed players not understanding it.

Later you can choose to either replace the current with new system (with or without some tweaks) or to just leave it as some sort of setting.

PR #655, opened 2019-07-01
fix joystick rotation

amir-arad:
currently the stick only increments ship's `targetRotation` when the axis value changes. so holding the stick steady at an angle does not translate to a continuous rotation. 
To continuously rotate using the stick, the pilot needs to jiggle the stick to create a stream of joystick events.

in this PR the joystick directly changes ship turn speed. feels much better IMO.

### DIFF
--- a/src/spaceObjects/playerSpaceship.cpp
+++ b/src/spaceObjects/playerSpaceship.cpp
@@ -1041,6 +1041,7 @@ void PlayerSpaceship::onReceiveClientCommand(int32_t client_id, sf::Packet& pack
     switch(command)
     {
     case CMD_TARGET_ROTATION:
+        turnSpeed = 0;
         packet >> target_rotation;
         break;
     case CMD_TURN_SPEED:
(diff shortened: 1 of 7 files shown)


src/spaceObjects/spaceship.cpp
```diff
@@ -466,7 +468,12 @@ void SpaceShip::update(float delta)
             warp_request = 0.0;
     }
 
-    float rotationDiff = sf::angleDifference(getRotation(), target_rotation);
+    float rotationDiff;
+    if (fabs(turnSpeed) < 0.0005f) {
+        rotationDiff = sf::angleDifference(getRotation(), target_rotation);
+    } else {
+        rotationDiff = turnSpeed;
+    }
```
daid: Shouldn't we update the target_rotation here somewhere, else when the turnSpeed goes to 0.0 the ship will rotate back to the target_rotation.

src/spaceObjects/playerSpaceship.cpp
```diff
@@ -1042,6 +1043,10 @@ void PlayerSpaceship::onReceiveClientCommand(int32_t client_id, sf::Packet& pack
     case CMD_TARGET_ROTATION:
         packet >> target_rotation;
```
daid: I think it would be good to set the turnSpeed to zero here. Meaning you can get back control by normal means if a joystick control was disconnected.

src/spaceObjects/spaceship.cpp
amir-arad: currently the only place that sets value to `turnSpeed` is the execution of `CMD_TURN_SPEED`, which also resets `target_rotation` to the current ship's rotation. setting it here seems redundant but mostly harmless.

daid: You're right, because the turnSpeed is only updated that way, and thus can only be set to 0 that way. Didn't think of that. That part still feels a bit iffy tough.

PR #656, opened 2019-07-01
rewrite joystick input

### LINKED ISSUE #413: REQUEST Configurable Keyboard/joystick inputs
It would be nice to have a setup screen (and config file) to allow mapping of inputs to actions. 
Recognizing that there are digital and analog actions (Throttle and movement type functions being analog).  It could allow analog functions to be assigned to keys, joystick buttons and/or Joystick Axis functions (multiple input mapping allowing up to 2 keys or 1 key and 1 axis would be optimal)   while only allowing keys and joystick buttons to be assigned to actions not identified as analog.
example: Impulse throttle assigned to a joystick axis and the W/S keys for increase/decrease. Fire tube 1 could be  joystick button 1 and keyboard key 1 (SFML is rather broken for accessing numpad keys last I checked)

amir-arad:
- options.ini configures joystick axis actions, similar to hotkeys.
- options.ini configures joystick buttons, using existing hotkeys properties (joystick buttons can act as any hotkey)
- joystick axes can control power and coolant, missile aim, and scan sliders.

this PR includes #655

fixes #413


src/spaceObjects/spaceship.cpp
```diff
@@ -116,8 +116,10 @@ SpaceShip::SpaceShip(string multiplayerClassName, float multiplayer_significant_
     impulse_acceleration = 20.0;
     energy_level = 1000;
     max_energy_level = 1000;
+    turnSpeed = 0.0f;
 
     registerMemberReplication(&target_rotation, 1.5);
+    registerMemberReplication(&turnSpeed, 1.5);
```
daid: I think 1.5 seconds between turnSpeed updates towards clients will give a few issues with joystick controls looking very jittery on other clients. 0.1 would be a better number. (note that this a minimum update delay, if the turnSpeed is not changed it will not send it at all)

src/screenComponents/scanningDialog.h
```diff
@@ -12,7 +13,7 @@ class GuiButton;
 class GuiScanningDialog : public GuiElement
 {
 private:
-    static constexpr int max_sliders = 4;
+    static constexpr int max_sliders = EScanningComplexity::SC_MAX;
```
daid: SC_MAX isn't the same thing as max_sliders. They are the same value by luck, not by design.

src/spaceObjects/spaceship.cpp
amir-arad: addressed. thanks

src/screenComponents/scanningDialog.h
amir-arad: addressed

daid:
Just noticed something, you where missing "return" statements in the onJoystickAxis functions, which is an warning in C/C++, but IMHO, should be an error. As this will invoke undefined behavior. Chances are it worked fine in testing, but it wasn't proper code. Missed it in my review.

Issue #628, opened 2019-02-05
Display Enhancement: show the password in upper and lower case as entered


pixelclef:
There's a noticeable mental disconnection between typing:
`password`
and seeing
`PASSWORD`

... especially when the password check is case sensitive!

amir-arad:
IMO the font in the entire game is rendered as uppercase. suggest a new font?

JonathanILevi:
Is a reasonable solution to just make passwords not case sensitive?
@daid

pixelclef:
> IMO the font in the entire game is rendered as uppercase. suggest a new font?

Understanding the typographical intent, I would recommend Eurostile. Close alternative would be Eurocine.

daid:
I kinda like the font, even tough it's all caps all the time. Note that the font was changed by someone else as a pull request, by someone who did the whole styling. So making the password case-insensitive would be a 'better' solution.

JonathanILevi:
This issue should be fixed?

PR #657, opened 2019-07-02
Possible fix for #628.

JonathanILevi:
Since the font used in Empty Epsilon always shows uppercase, I made it so the password is always uppercase/case-insensitive.

Is this an appropriate fix for #628?


Xansta:
This looks like it addresses the part of the problem that would confuse the user when typing lower case and seeing upper case. A different font would be a bigger change

JonathanILevi:
Oh, leme fix that.

JonathanILevi:
I changed it to use the `upper` method.  Much nicer now.

PR #660, opened 2019-07-07
Engineering quick-tweak function.

JonathanILevi:
I added the ability to click on the power level and coolant level "progress" bars in engineering to make quick adjustments to the systems.

Often times in engineering a lot of quick, rough adjustments are needed.  Clicking though each system to adjust the sliders is tedious and cannot be done quick.  Also keeping a mental understanding of the systems states is difficult when all the adjustments are done separate from the visual feedback.  Some might like this to add difficulty but I prefer the difficulty to be the game and communication with the other officers rather than the interface method with the computer.  When interfacing with the game is smooth and direct, so that mental thoughts translate subconsciously into the game, allowing me to do seemingly amazing things, have been some of the most satisfying experiences games have given me.

I know I will use this adjustment for myself and I hope you will accept it, when it is polished, upstream.  The code is perhaps a little rough still, but I would love to hear your thoughts on the concept.

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,6 +191,7 @@ set(SOURCES
 	src/gui/gui2_arrowbutton.cpp
 	src/gui/gui2_entrylist.cpp
 	src/gui/gui2_progressbar.cpp
+	src/gui/gui2_progressslider.cpp
 	src/gui/gui2_scrolltext.cpp
 	src/gui/gui2_advancedscrolltext.cpp
 	src/gui/gui2_button.cpp
--- a/src/gui/gui2_progressbar.cpp
+++ b/src/gui/gui2_progressbar.cpp
@@ -1,13 +1,13 @@
 #include "gui2_progressbar.h"
 
-GuiProgressbar::GuiProgressbar(GuiContainer* owner, string id, float min, float max, float value)
-: GuiElement(owner, id), min(min), max(max), value(value), color(sf::Color(255, 255, 255, 64)), drawBackground(true)
+GuiProgressbar::GuiProgressbar(GuiContainer* owner, string id, float min_value, float max_value, float value)
+: GuiElement(owner, id), min_value(min_value), max_value(max_value), value(value), color(sf::Color(255, 255, 255, 64)), drawBackground(true)
 {
 }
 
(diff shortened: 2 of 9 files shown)


amir-arad:
screenshots would be nice

daid:
Wondering how this looks myself as well, if it looks the same but just with added UI interaction, then the code could be a whole lot simpler, as the new GUI control can be a subclass of slider or progressbar then, only addeding the graphics or input handling that it needs that the base does not have.

JonathanILevi:
Engineering looks exactly the same as it did before, only adding an interaction.

I originally had the `ProgressSlider` a subclass of the `Progressbar` and then I switched to a subclass of the Slider and then I just made it a subclass `GuiElement` directly.

To make it a subclass of `Progressbar` some private member variables need to be changed to protected and a hole slew of slider code needs to be copied over, and `onMouseDrag` adjusted for the different proportions.

Being a subclass of `Slider` requires less copied code but the `ProgressSlider` does not have `SnapPoint`s (currently at least).

What would have the least code duplication would be to add a `BasicSlider` as the common parent of `Slider` and `ProgressSlider`.  Or we can also do multiple inheritance, if you prefer that.

Do have a preference on how it gets done?

JonathanILevi:
Also for a quick change for my sanity.  `Progressbar` uses `min` and `max` while `Slider` uses `min_value` and `max_value`.  Can I change one or the other so that they both use either `min`/`max` or `min_value`/`max_value`?

daid:
You can do the sanity change, no issue with that. A single base class I have no issue with, multiple inheritance is just a pain, so I wouldn't do that. Also, making things protected instead of private is also no issue at all, especially if that saves code duplication.

daid:
Oh, and the change itself makes perfect sense. There is already the power/coolant screen that has all the sliders directly accessible as well.

JonathanILevi:
`ProgressSlider` now uses the new `BasicSlider` as its parent.
The code `ProgressSlider` is quite short.
I think the code for this feature is polished enough, thoughts?

JonathanILevi:
@daid, what do you think?

daid:
Looks pretty fine now, there is always room for improvement, but that I can say about all my code as well :-)

PR #665, opened 2019-07-23
make maximum coolant available to a crew configurable

czenker:
I think adding different coolant levels might make some missions more interesting:

* @TolotosRhodan created a [mission](https://github.com/TolotosRhodan/EE-MotU) where you can dispose of some coolant in order to create a temporary nebula to escape like a ninja
* decreasing coolant would force a crew to play more strategic and turn systems off more often
* scenarios could add a mechanic where coolant has to be refilled regularly or can be earned as reward
* "sabotage" the players by draining their coolant and have them improvise

EmptyEpsilon_015
EmptyEpsilon_016

---

Im currently having a problem with linking. The compiler runs fine, but the linker seems not to be able to find the reference to `max_coolant_per_system`. I spend over an hour trying to figure out what the problem is, but I am out of ideas.

`playerSpaceship.cpp:(.text+0x5876): undefined reference to 
 PlayerSpaceship::max_coolant_per_system'`

Maybe someone can give it a try. It could be an issue with my setup or it is just my missing C++ knowledge or some weird foo... :-/ 

@daid to the rescue!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Added
+
+- Mission scripts can modify the total coolant available to engineering
+
+### Changed
+
 - Nothing
 
 ## [2018-02-15]
--- a/src/screens/crew6/engineeringScreen.cpp
+++ b/src/screens/crew6/engineeringScreen.cpp
@@ -117,6 +117,8 @@ EngineeringScreen::EngineeringScreen(GuiContainer* owner, ECrewPosition crew_pos
     power_label->setVertical()->setAlignment(ACenterLeft)->setPosition(20, 20, ATopLeft)->setSize(30, 360);
     coolant_label = new GuiLabel(box, "COOLANT_LABEL", "Coolant", 30);
     coolant_label->setVertical()->setAlignment(ACenterLeft)->setPosition(110, 20, ATopLeft)->setSize(30, 360);
+    total_coolant_label = new GuiLabel(box, "TOTAL_COOLANT_LABEL", "Total Coolant", 30);
+    total_coolant_label->setVertical()->setAlignment(ACenterLeft)->setPosition(200, 20, ATopLeft)->setSize(30, 360);
 
     power_slider = new GuiSlider(box, "POWER_SLIDER", 3.0, 0.0, 1.0, [this](float value) {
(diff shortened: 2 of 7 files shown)


czenker:
I found the reason for the linker issue: https://stackoverflow.com/questions/16957458/static-const-in-c-class-undefined-reference/16957554 It had to do with the way `std::min` works.

daid:
Yeah, that linking issue is really annoying. It's fixed in the C++17 or C++19 standard (not sure which)

I like the feature. A scenario with a coolant leak sounds like fun, puts some pressure on the players.
Not sure about the name and location of the "total coolant" label. It's more of a global information thing for the engineer, now located at a system specific slider.

Maybe "Available coolant: 100%" would be a better label, and on where to put it, not sure, the screen is already quite full... and I think the label should also be placed on the power/coolant screen as well.

Showing the amount of used coolant doesn't do much I think. As the total used is 0 or 100% of available due to the automatic distribution system.

czenker:
Good suggestions. I moved the display to the top left on the engineering screen. This really makes it look less clumsy.

EmptyEpsilon_018

Also I added it to the Power Management Screen. I'm not 100% happy with the looks but I think it will do. To make the coolant display less lost I also added the energy display – I think it makes sense to also have it there.

EmptyEpsilon_020

And finally this is the test script I used for playing around.

    -- Name: Hello World
    -- Description: Testing stuff
    -- Type: Mission
    
    require "utils.lua"
    function init()
        local p = PlayerSpaceship():setTemplate("Flavia P.Falcon"):setFaction("Human Navy")
    
        p:addCustomButton("engineering", "up", "coolant +2", function()
            p:setMaxCoolant(p:getMaxCoolant() + 2)
            print(p:getMaxCoolant())
        end)
        p:addCustomButton("engineering", "down", "coolant -2", function()
            p:setMaxCoolant(p:getMaxCoolant() - 2)
            print(p:getMaxCoolant())
        end)
    
        local auto_coolant = true;
    
        p:addCustomButton("engineering", "coolant", "toggle autocoolant", function()
            p:setAutoCoolant(auto_coolant)
            auto_coolant = not auto_coolant
        end)
    end

daid:
Yeah, much better I think. I also like the addition to the power management screen.

PR #667, opened 2019-07-25
toggle between simple and complex radar signatures

czenker:
This commit adds a toggle to the server creation screen that can simplify the radar signatures.

The default mode mixes gravitational, bio signs and electronic signatures into all of the three radar signatures. This is not intuitive for new players. To make the science screen more accessible the simple mode only shows one dimension for each radar signature:
 * red shows gravitation
 * green shows bio signs
 * blue shows electronic signatures

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -31,6 +31,7 @@ GameGlobalInfo::GameGlobalInfo()
     long_range_radar_range = 30000;
     use_beam_shield_frequencies = true;
     use_system_damage = true;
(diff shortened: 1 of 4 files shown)


czenker:
I will close this PR – I came to realize that this does not really add to the game. Interpreting the signatures is not sooo difficult if you got the hang of it. 

* strong oscillating lines are electronic signatures
* electrical signatures are best visible on the red band
* gravity affects all bands

And relying on the signatures to beat the game is not something that a beginners crew would need to do anyways...

PR #671, opened 2019-08-09
move order panel to bottom right on GM screen

czenker:
I moved the order panel on the GM screen to the bottom right, so that custom functions can be called with CpuShips being selected. This makes `getGMSelection` muuuuch more valuable. :smile:

EmptyEpsilon_021


daid:
Makes sense :-)

PR #672, opened 2019-08-13
add modifier keys to the GM screen

czenker:
- `Shift` adds objects to the current selection
- `Ctrl` only selects stations and ships
- `Alt` only selects objects from the same faction as the faction selector

### DIFF
--- a/src/screens/gm/gameMasterScreen.cpp
+++ b/src/screens/gm/gameMasterScreen.cpp
@@ -445,6 +445,7 @@ void GameMasterScreen::onMouseUp(sf::Vector2f position)
     case CD_BoxSelect:
         {
             bool shift_down = InputHandler::keyboardIsDown(sf::Keyboard::LShift) || InputHandler::keyboardIsDown(sf::Keyboard::RShift);
+            //Using sf::Keyboard::isKeyPressed, as CTRL does not seem to generate keydown/key up events in SFML.
             bool ctrl_down = sf::Keyboard::isKeyPressed(sf::Keyboard::LControl) || sf::Keyboard::isKeyPressed(sf::Keyboard::RControl);
             bool alt_down = InputHandler::keyboardIsDown(sf::Keyboard::LAlt) || InputHandler::keyboardIsDown(sf::Keyboard::RAlt);
             PVector<Collisionable> objects = CollisionManager::queryArea(drag_start_position, position);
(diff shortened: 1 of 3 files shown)


src/screens/gm/gameMasterScreen.cpp
```diff
@@ -444,14 +444,32 @@ void GameMasterScreen::onMouseUp(sf::Vector2f position)
         break;
     case CD_BoxSelect:
         {
+            bool shift_down = InputHandler::keyboardIsDown(sf::Keyboard::LShift) || InputHandler::keyboardIsDown(sf::Keyboard::RShift);
+            bool ctrl_down = sf::Keyboard::isKeyPressed(sf::Keyboard::LControl) || sf::Keyboard::isKeyPressed(sf::Keyboard::RControl);
+            bool alt_down = InputHandler::keyboardIsDown(sf::Keyboard::LAlt) || InputHandler::keyboardIsDown(sf::Keyboard::RAlt);
```
czenker: @daid Control key strokes seem not to trigger a `sf::Event::KeyEvent` on their own. I did not dare adapt the `InputHandler::keyboardIsDown` logic because I have no idea what the side effects could be.

daid: Using sf::keyboard is fine in this case. Would be nice to place a comment in the code why that is done, instead of only on the pull request. That not someone else comes by and tries to "fix" the inconsistency.

SFMLs input handling isn't super great. So I'm not surprised.

src/screens/gm/gameMasterScreen.cpp
```diff
@@ -444,14 +444,32 @@ void GameMasterScreen::onMouseUp(sf::Vector2f position)
         break;
     case CD_BoxSelect:
         {
+            bool shift_down = InputHandler::keyboardIsDown(sf::Keyboard::LShift) || InputHandler::keyboardIsDown(sf::Keyboard::RShift);
+            bool ctrl_down = sf::Keyboard::isKeyPressed(sf::Keyboard::LControl) || sf::Keyboard::isKeyPressed(sf::Keyboard::RControl);
```
daid:
```suggestion
            //Using sf::Keyboard::isKeyPressed, as CTRL does not seem to generate keydown/key up events in SFML.
            bool ctrl_down = sf::Keyboard::isKeyPressed(sf::Keyboard::LControl) || sf::Keyboard::isKeyPressed(sf::Keyboard::RControl);
```

PR #673, opened 2019-08-13
fix: documentation was not rendered

czenker:
Removing the cbp file in f1544194fcb26d33b315da018eac0544d6cba58d broke the rendering of the script reference.


czenker:
One other issue I found is that the `script_reference.html` seems not to be created anymore during the build!?

daid:
During the android build system testing I noticed that the script reference was always build, even if it was up to date, I tried to fix that, but I guess that resulted in it never being build any more...

PR #674, opened 2019-08-13
add message overlay for game master screen

czenker:
Added the possibility to show a message on the GM screen as it is possible on all stations.

I found it limiting to have no space to display longer texts to the GM during a scenario or inform them that the players have achieved a significant milestone.

It looks identical to the overlay on the stations (probably because I copied most of the code from there :smiley: )

EmptyEpsilon_022


daid:
I seem to mis a function to add a message? Forgot to commit something?

czenker:
Yeah – I missed those...

PR #676, opened 2019-08-16
fix #649

### LINKED ISSUE #649: [BUG] Black main screen
There is a bug in the main screen : in some situations, the viewport disappear and all the screen is black.
How to do it : 
- create more than one playerspaceship. The problem don't appear with the first person
- choose the main screen for a playerspaceship (not the first) and choose first person view ("F")
- destroy all object in front of the playerspaceship
- the screen go black

I think that the viewport does not appear if there is not object to show. So, with the first person view, it can be nothing to show. The strange thing is that it does not the case for the first player.

Do you have some advice to correct it ?

amir-arad:
fix #649
ensure at least one depth iteration in viewport,
as some mandatory GL instructions are in the render_lists loop.

### DIFF
--- a/src/screenComponents/viewport3d.cpp
+++ b/src/screenComponents/viewport3d.cpp
@@ -149,7 +149,10 @@ void GuiViewport3D::onDraw(sf::RenderTarget& window)
         float depth;
     };
     std::vector<std::vector<RenderInfo>> render_lists;
-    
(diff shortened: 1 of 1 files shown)


amir-arad:
Apologies in advance, I don't know what exactly was the issue. 
as with `3f4524b`, I just black-boxed it. some prep work in the 3d rendering outer loop is important to the texture rendering, so I've made sutre there is at least one ("empty') iteration. 
will gladly take feedback, though sadly I'm slow to answer.

PR #679, opened 2019-08-29
Missile size

nallath:
Because it's been quite some time since I last did something with empty epsilon, I'm already posting this bit as a draft pull request to get some feedback. 

The idea of this is to enable different sizes for the various missile weapons. For now, I've limited them to small/medium/large. For the moment, they essentially just influence the damage done by the missiles (50% dmg for small, 100% for medium and 200% for large). In case of EMP and Nuke, it also scales the blast radius by the same factor (Although it might make a large nuke _waaay_ to powerfull. On the other hand, having some mini-nukes in there could be interesting) 

The main reason for adding this in the first place is to allow for a much better differentiation between ships, as fighters or small ships often don't have missiles and if they do, they tend to have very small storage due to balance reasons. 

I'm not quite sure if I should also add the option to have multiple different tube sizes per ship, as this would also mean that you would need different storage types as well (which might clutter the interface). On the other hand, it would allow for interesting ship types again (a bomber that has a single medium tube and 2 small ones).


daid:
Quite interresting, I think the scaling of the nukes and EMPs is a bit overkill. Both a damage and radius boost makes them super weapons...

However, with one more tweak, that could be offset. What if the homing and turning speeds of missiles is inversely effected?

amir-arad:
I had similar thoughts, but also with extending the range of missiles. WDYT?

nallath:
We could also scale the damage and blast radius of the nukes a bit different of course. 

I'm really in favor of also making the big missiles really slow (and the small ones fast) since it suddenly gives missiles a specialized role (eg; Anti capital vs Anti fighters). But this would to me be another reason to have mixed tubes possible on a ship. With mixed tubes, you could have a capital ship with a few broadside heavy missiles and a scattering of small point defense missiles.

daid:
It might be nice to have templated missles, just like spaceships and stations. However, the missle storage gets a bit more complex then. And the script interface will need some backwards compatibility or it will break someones missions...

nallath:
That's kinda why I just went with "size". That way it won't break backward compatible missions (because the default is just medium).

nallath:
So, after a bit of struggeling with the logging and figuring out why the templates changes didn't get set in the code (mumbles something about the install putting files where I didn't find them), i've ran into an issue that I don't know how it's caused.

So for some reason, it always sets the size to 1, even if I use the template:setTubeSize(0, 0) (which should set it to small and thus 0). It's probably something super obvious I'm missing, but I can't really find what it is. @daid Any idea what thing I'm missing?

daid:
because the 2nd parameter of setTubeSize is an enum, and your script conversion function from lua to that enum takes a string, not a number.

czenker:
Hey nallath.

Great idea to tackle the lack of missile diversity. It would be great to diversify ships by their missile loadout also.

> It might be nice to have templated missles, just like spaceships and stations.

I prefer the idea of missile templates. It allows for more strategic approaches - especially for "standard" missiles. Do you take the bigger, but slower ones or the weaker, but faster ones on your ship.

I also thought of having some kind of [Cruise Disruptor Missile](https://freelancer.fandom.com/wiki/Cruise_Disruptor) in the game as seen in Freelancer. With missile templates (and a callback on hit) that could be easily done.

>  However, the missle storage gets a bit more complex then.

That is true. Also the AI would need to be adapted to select a sensible missile type for the occasion. But I think this could be doable.

> And the script interface will need some backwards compatibility or it will break someones missions...

On the other hand: Changing the tube size of existing ship templates might as well imbalance someones mission.

nallath:
First off; DERP. I can't believe I missed that stupid mistake with the LUA bindings. 

As for adding actual templates for the missiles, I would like to add that at some point, but it's a way bigger change. It's been quite some time since I last touched the code before this, soo I wanted to do a slightly easier project to get myself up to speed again. 

That being said, I do think that the missile templates and the missile size could work together. The template defines what a missile does (Guided, repeating, max_damge_after_time, explosive, damage type (dmg vs electrical), etc). Once you have the basic properties of a missile, we could think about making all the existing ones a template and then we could try and come up with more types.

Alternatively, we could also just think about some new missile types and not bother with templates at all. I've got some ideas that would be pretty hard to do with a template but should be doable with "just" c++ code (Cluster missile, defensive platform (dump a small ship with no engines that has a rotating laser), fake mines, EMP mines, etc)

Triqueon:
I like this a lot. I do have one thing I wonder about: as it is, it's quite feasible to dodge missiles with combat maneuvers. The large missiles turning slower will make this even easier. While that's not so much a problem for HVLIs (which don't turn) or Homing missiles, it has the potential of making large nukes/EMPs  rather pointless. Would it be possible to have nukes that reach the end of their range just explode regardless of having hit their target or not? Having them just disappear if they hit nothing makes them... well, meh. This might be another PR though.

nallath:
Right now there is only a single ship that has large nukes (mostly because we have a ton of small & medium ships, but hardly any large ships).

Having nukes explode at their end of life makes sense to me, since it greatly increases the danger of using them. An issue with that is that we should also update the AI to actually try to prevent being in a nuke range (maybe a behavior like with the mines?). It will also enable some strategic options by timing the location where it will explode. 

If this is merged I will have a look at the explode at end of life code.

src/spaceObjects/missiles/missileWeapon.cpp
```diff
@@ -1,12 +1,13 @@
...
+#include "spaceObjects/explosionEffect.h"
 
-MissileWeapon::MissileWeapon(string multiplayerName, const MissileWeaponData& data)
-: SpaceObject(10, multiplayerName), data(data)
+MissileWeapon::MissileWeapon(string multiplayer_name, const MissileWeaponData& data)
+: SpaceObject(10, multiplayer_name), data(data)
 {
     target_id = -1;
     target_angle = 0;
+    category_modifier = 1;
```
daid: I think the category_modifier needs to be replicated to the clients as well, for proper rendering simulation on the client side.

daid:
Exploding at the end of it's lifetime is relatively easy.

Another question is the damage scaling, for the larger missiles, would it be interesting to scale the damage if you are closer to the explosion?

nallath:
The damage already scales, right? Nukes have a damage_at_center and damage_at_edge property (160 and 30 respectively)

daid:
Oh, I missed that :-)

nallath:
@Triqueon Since it was pretty easy to do, i've added a quick prototype that adds explosion at end of life: https://github.com/daid/EmptyEpsilon/pull/689

PR #683, opened 2019-09-12
Additional Hacking Minigame / Hacking Difficulty

Triqueon:
Since I read somewhere that the Minesweeper in use for hacking was always only intended as a placeholder, and Minesweeper doesn't feel very hacky to me, I implemented my own optional additional minigame, and an additional setting for hacking difficulty, which basically controls board size.

General Features in this PR:
* New Hacking Minigame, Lights On: Turn on all lights in a nxn grid, everytime you turn on a light, all adjacent lights toggle their state
* Option to choose whether to use MineSweeper, LightsOn, or a random game when hacking in Server Creation screen
* Option to set Hacking Difficulty in Server Creation Screen. MineSweeper starts with a 6x6 Grid with 6 Bombs, and adds 2 bombs and 2 grid size in each direction with each difficulty. Lights On starts with a 3x3 Grid and adds 2 grid size in each direction with each difficulty

Disclaimer: This is my first C++ project in quite a while. If I've missed any coding conventions or you see any memory issues, please let me know.


amir-arad:
some screenshots would be nice :)

nallath:
Oh, for the next time, you can also batch commit suggestions. That way you won't have a million small change commits.

Triqueon:
Screenshots:
[Server Creation Screen with new selectors for Hacking Difficulty and Games](pascalstein.de/images/ServerCreation.jpg)
[Fiendish Difficulty Minesweeper, partially solved](pascalstein.de/images/FiendishMines.jpg)
[Fiendish Difficulty Lights On, partially solved](pascalstein.de/images/FiendishLights.jpg)
[Easy Lights On](pascalstein.de/images/SimpleLights.jpg)

src/screenComponents/mineSweeper.h
```diff
@@ -0,0 +1,35 @@
...
+#include "miniGame.h"
+#include "gui/gui2_togglebutton.h"
+#include <vector>
+#include <memory>
+
+
+class MineSweeper : public MiniGame {
+  public:
+    MineSweeper(GuiPanel* owner, GuiHackingDialog* parent, int difficulty);
+    virtual void reset();
```
daid: As these virtual functions override the base class, it would be good to mark them as "override":
https://en.cppreference.com/w/cpp/language/override

src/screenComponents/mineSweeper.h
```diff
@@ -0,0 +1,35 @@
...
+ *  Original implementation by https://github.com/daid
+ */
+
+#ifndef MINESWEEPER_H
+#define MINESWEEPER_H
+
+#include "miniGame.h"
+#include "gui/gui2_togglebutton.h"
+#include <vector>
+#include <memory>
```
daid: I don't see anything from vector or memory being used in this header file?

src/screenComponents/hackingDialog.h
```diff
@@ -19,35 +21,25 @@ class GuiHackingDialog : public GuiOverlay
...
-    {
-    public:
-        GuiToggleButton* button;
-        bool bomb;
-    };
-    FieldItem field_item[hacking_field_size][hacking_field_size];
-    
+
+    GuiPanel* minigame_box;
+    MiniGame* game;
```
daid: If you make the minigame a std::shared_ptr, you don't need to keep track of deleting it, and these is then no risk of memory leaks:
https://en.cppreference.com/w/cpp/memory/shared_ptr

(this isn't needed for the gui elements, as they manage their own lifetime and get deleted with their parent)

src/screenComponents/mineSweeper.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "miniGame.h"
+#include "hackingDialog.h"
+#include "gui/gui2_togglebutton.h"
+#include "gui/gui2_label.h"
+#include "gui/gui2_progressbar.h"
+#include "gui/gui2_panel.h"
+
+MineSweeper::MineSweeper(GuiPanel* owner, GuiHackingDialog* parent, int difficulty)
+: MiniGame(owner, parent, difficulty) {
+    for(int x=0; x<difficulty; x++)
```
daid: I understand the need for a single difficulty parameter. But I think it would be good to initialize a "hacking_field_size" and "bomb_count" from that, like in the old code. It reads pretty difficult now with "difficulty" being used for various things.

src/screenComponents/miniGame.h
```diff
@@ -0,0 +1,40 @@
...
+#include <vector>
+#include "hackingDialog.h"
+
+class GuiButton;
+class GuiToggleButton;
+class GuiProgressbar;
+class GuiLabel;
+class GuiPanel;
+
+class MiniGame
```
daid: Would be good to add a base class of sf::NonCopyable:
https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1NonCopyable.php

src/screenComponents/miniGame.h
```diff
@@ -0,0 +1,40 @@
...
+{
+  public:
+    MiniGame(GuiPanel* owner, GuiHackingDialog* parent, int difficulty);
+    virtual ~MiniGame();
+
+    virtual float getProgress();
```
daid: If any of these are "must be implemented by subclass", it would be better to make them true abstract virtual functions:
https://en.wikibooks.org/wiki/C%2B%2B_Programming/Classes/Abstract_Classes

src/screenComponents/lightsOut.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+bool LightsOut::LightsOutToggleButton::toggle()
+{
+    bool value = !getValue();
+    setValue(value);
+    return value;
+}
+
+LightsOut::LightsOutToggleButton::LightsOutToggleButton(GuiContainer* owner, string id, string text, func_t func) : GuiToggleButton(owner, id, text, func)
```
daid: Following the rest of the code style, the constructor is defined first, so before the toggle function.

src/screenComponents/hackingDialog.cpp
```diff
@@ -128,115 +133,41 @@ bool GuiHackingDialog::onMouseDown(sf::Vector2f position)
...
-            item.button->setText("");
-            item.button->setValue(false);
-            item.button->disable();
-        }
+void GuiHackingDialog::getNewGame() {
+    delete game;
+    int difficulty = 2;
+    EHackingGames games = HG_All;
+    if (gameGlobalInfo) {
+      difficulty = gameGlobalInfo->hacking_difficulty + 1;
```
daid: Magic + 1?

src/screenComponents/miniGame.h
Triqueon: I'm a bit conflicted on this, and open to suggestions. I had these as pure virtual functions originally. I decided it was nicer to have the hacking dialog have a dummy Minigamepanel when first opened, and thus needed to actually instantiate the class MiniGame, so these had to be implemented.

The more I think about it, the more I'm just going to go with an Empty panel instead of the dummy minigame, and make these abstract/pure virtual again.

src/screenComponents/hackingDialog.cpp
Triqueon: Yeah. Problem is: the selectors on the Server creation screen are 0-based, as far as I can tell, and a difficulty of 0 doesn't make much sense for the way the games are defined (Both have an offset, but LightsOut comes out as a 1x1 grid with difficulty 0, which is... sort of trivial, and MineSweeper gets 4x4, which isn't much better). I'm not altogether happy with the magic difficulty generation for each separate minigame either, so I'll probably move that, and the associated magic numbers to the individual game implementations.

src/screenComponents/miniGame.h
Triqueon: And, after looking at it, that doesn't work either, as the Reset Button would otherwise call reset() on a null pointer. There needs to be something in the game variable, and I think presenting an empty board is better to guide the user towards making a choice on which system to hack than a filled, but deactivated board. So, I think I will leave this as is.

src/screenComponents/mineSweeper.h
Triqueon: Agreed, done.

src/screenComponents/mineSweeper.h
Triqueon: Oops. Leftovers from initially implementing the entire structure here before it was moved to the MiniGame class. Good catch, thank you.

src/screenComponents/hackingDialog.h
Triqueon: Agreed. Done.

src/screenComponents/mineSweeper.cpp
Triqueon: +1 for readability. Did that here and for the LightsOut.

src/screenComponents/miniGame.h
Triqueon: Done

src/screenComponents/lightsOut.cpp
Triqueon: Moved the constructor up above the toggle definition.

src/screenComponents/hackingDialog.cpp
Triqueon: Moved the adjustment to the definition of the grid_size variables. That way, the original difficulty is still accessible if it's ever needed, and if someone wants, it's easy to make bomb counts different from grid size or similar.

src/screenComponents/miniGame.h
daid: Ok. Good. I was just wondering :-)

daid:
Note that I squashed and merged, this made all 91 commits into a single commit into the master branch. One of the reasons is else my CI system would build all 91 commits.

nallath:
Hmm, wouldn't it just be better to let the CI build the head? Now you're losing a lot of data / convenience from squashing the commits.

daid:
Losing lots of data? Just some history of how the complete patch was made. It's the end result that really matters.
Losing convenience? How? It's a lot more convenient squashed I would say, as looking at the commit history it's easier to see what the changes where.

And my CI builds all commits to see which exact commit breaks the build. (But my server went down like an hour ago...)

nallath:
Well, there is a reason atomic commits are seen as a best practice. If things are going to be squashed anyway, that's esentially as if only the head gets tested (since now you also lose the info what exact change broke it). 
It's not really a big issue now, but it might cause some issues in the future.

daid:
My project, my rules :-) And I like my contributors not being burdened with setting up atomic commits. Especially if I do not do that myself. Remember, it's a small project, with only a few poeple working on it.

So we don't need complex rules like atomic commits, rebasing before pull requests and protecting master branches like you would do in bigger projects with larger teams.

aBlueShadow:
@Triqueon 
There where reports that you can now win the minesweeper game when you also clicked on known bomb location - wich I just confirmed.
As I don't recall this issue from previous version, I guess it was somehow introduced by this PR. Maybe you could take a look?

Triqueon:
@aBlueShadow : Thanks. Looks like I tried to be to clever. https://github.com/daid/EmptyEpsilon/pull/727 has a fix.

PR #685, opened 2019-09-16
BeamEffect access for LUA scripts

Triqueon:
For some reason, BeamEffects can't be scripted, unlike explosions. These lines change that.

### DIFF
--- a/src/spaceObjects/beamEffect.cpp
+++ b/src/spaceObjects/beamEffect.cpp
@@ -4,6 +4,14 @@
 #include "mesh.h"
 #include "main.h"
 
+/// BeamEffect is a beam weapon fire effect that will fade after 1 seond
+/// Example: BeamEffect():setSource(player):setTarget(enemy_ship)
+REGISTER_SCRIPT_SUBCLASS(BeamEffect, SpaceObject)
+{
+    REGISTER_SCRIPT_CLASS_FUNCTION(BeamEffect, setSource);
+    REGISTER_SCRIPT_CLASS_FUNCTION(BeamEffect, setTarget);
+}
+
 REGISTER_MULTIPLAYER_CLASS(BeamEffect, "BeamEffect");
 BeamEffect::BeamEffect()
 : SpaceObject(1000, "BeamEffect")


daid:
Reason is simple, I never needed it.

You might want to expose "beam_texture" "beam_fire_sound" and "beam_fire_sound_power" as well to scripts then?

aBlueShadow:
How do I adress the source and target? The example leads to the assumption you just need the object handle, but that throws the lua error 

> bad argument #2 to 'setSource' (number expected, got no value)

looking further in the code, the c++ function also needs a vector argument, but that one is not registered in the script function

Triqueon:
I got this to work without problems using the object handle. The example is, admittedly, poorly chosen and incomplete, I'll work up another PR tomorrow to adress the texture and sound setters mentioned by daid, and complete it. The following line of code is taken from the scenario for which I needed this and works for me:

`      BeamEffect():setSource(artifacts[i * 2], 0, 0, 0):setTarget(central_station, 0, 0)
`

aBlueShadow:
Okay then I was almost right (overlooked the fact that setTarget only needs Vector2f instead of Vector3f)
Anyway, I have to correct myself, the script did not throw me out of a loop, it freezes the whole game, without any error message. Tested on two different machines, two different (debian based) distros. I tried it on a script where I already have a timer, so I could test it on a single instance. Will now setup a minimal script to test if it will reproduce.

aBlueShadow:
Okay, here is a minimal example, can you reproduce the issue? Instead of firing a beam effect every 5 seconds, it freezes after the first time it should fire here.

`-- Name: Beam Test
-- Description: 
-- Type: Basic

function init()

player = PlayerSpaceship():setFaction("Human Navy"):setTemplate("Atlantis") 
dummyobject = Artifact():setPosition(500, 0)
dummyobject:setModel("SensorBuoyMKIII")
beam_timer=0

end 

function update(delta) 

beam_timer=beam_timer+delta
    if beam_timer > 5 then 
        BeamEffect():setSource(player, 0, 0, 0):setTarget(dummyobject, 0, 0)
        print "beam" -- debug - the script gets this far, but freezes anyway
        beam_timer=0      
    end    
end
`

Triqueon:
That example works on my machine. One Beam every 5 seconds. I ran it for several minutes with no problems. I did compile mine for windows, maybe that makes a difference?

Triqueon:
While looking at exposing the mentioned attributes to the scripting engine, I discovered that some members have no defaults, but are usually set via either the BeamWeapon or BeamTemplate classes. This might lead to some odd behavior, including SEGFAULTs or the like, depending on your compile flags. Head over to https://github.com/daid/EmptyEpsilon/pull/686 and see if that solves your problem.

aBlueShadow:
Yeah, now it works! So the missing defaults seemed to cause the issue.

czenker:
Oh wow. How did I miss this Merge Request? It is finally time for a script to mine asteroids. :)

Thanks @Triqueon for implementing it.

PR #686, opened 2019-09-17
Feature/beam effects settings for lua

Triqueon:
Exposes the beamEffects properties to lua and sets some defaults which are usually set via BeamTemplate or BeamWeapon, but uninitialized when a BeamEffect is produced by the script.

Question on the side: the default laser fire sound is sfx/laser_fire.wav. There is a similar-sounding laser.wav in the main resources directory that I could find no usage of. Is that leftover from something, or is my grep-fu just bad?


daid:
The laser.wav could very well be a leftover of something, not sure.

PR #687, opened 2019-09-17
read beam settings in lua scripts

aBlueShadow:
Right now, you can only set the beam frequency and target systems in lua, but not read them. This PR closes that gap, including a function to also get the name of the target system.


daid:
I think some of these should be in the spaceship base class, not in the playerspaceship, as CPU ships can also have frequencies.

aBlueShadow:
Yes I thought about that, but as the corresponding set-commands were in playerspaceship, I decided to make them symmetrical to be on the safe side. But of course I can change that.

daid:
Not sure why they are in the playership, maybe the are contributed at well, and I didn't notice it?

aBlueShadow:
Maybe, probably a long time ago then, as commandSetBeamFrequency seems to be pretty old. 
The functions in this PR should be fine now though, as the target functions are player specific

daid:
Ah, if its linked to the "commandXXX" function, that makes sense to be at the player ship, as those functions are functions that are used to communicate from clients to the server. And thus are only on player ships.

PR #688, opened 2019-09-18
Ability to turn beam fire ring off per script

aBlueShadow:
This is basically a small addition to #686
The effect features are great, however, "setDuration" is not much useful, as bigger values also make the fire ring (much) bigger and therefore looking out of place.
That's why I added a new method "setRing", wich can turn the fire ring off. This will also create additional use cases for the beam effect, like tractor beams, communication beams, mining lasers, etc.

### DIFF
--- a/src/spaceObjects/beamEffect.cpp
+++ b/src/spaceObjects/beamEffect.cpp
@@ -30,6 +30,7 @@ BeamEffect::BeamEffect()
     beam_fire_sound = "sfx/laser_fire.wav";
     beam_fire_sound_power = 1;
     fire_ring = true;
+    registerMemberReplication(&lifetime);
     registerMemberReplication(&sourceId);
     registerMemberReplication(&target_id);
     registerMemberReplication(&sourceOffset);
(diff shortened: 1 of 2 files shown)


src/spaceObjects/beamEffect.cpp
```diff
@@ -28,6 +29,7 @@ BeamEffect::BeamEffect()
     beam_texture = "beam_orange.png";
     beam_fire_sound = "sfx/laser_fire.wav";
     beam_fire_sound_power = 1;
+    fire_ring = true;
```
daid: fire_ring requires a memberReplication as well then, as the clients need to know it for proper rendering.

aBlueShadow: Ah, missed that, tested only in a single Instance. While re-testing, I noticed that lifetime is also only known in the server. Will fix both.

daid: As the lifetime is simulated on the client as well, and updated on the server every tick, it's better to specify an update interval as 2nd parameter to the registerMemberReplication, like "100" for 100ms between updates. Else it will send out data every frame.

aBlueShadow: I see. But are you sure that parameter is measured in milliseconds? On other occurances, it seem to vary between 0.1 and 1.0, which looks like it is seconds. So 100ms would be 0.1.

daid: My mistake. Been working with other code that does things in ms a lot.

aBlueShadow: Okay. So it should now in order then, I guess.

PR #693, opened 2019-09-28
Customizable self-destruct duration based on an options.ini value.

DrCamel9000:
Added the ability to customize the self-destruct duration based on an options.ini value (self_destruct_countdown).

The default is 10 seconds as per the original system.


daid:
Kinda surprised, what's the usecase of this? A LARP where people have more time to escape?

DrCamel9000:
> Kinda surprised, what's the usecase of this? A LARP where people have more time to escape?

Yes, exactly that.

A friend has been putting together a physical room with escape room elements integrated with Empty Epsilon and one of the scenarios he wants to do would require triggering the self-destruct sequence and then exiting the physical space, for which a bit more time is required.

Issue #605, opened 2018-12-22
Warp slider on UI not updated when changed through HTTP api


Tuhis:
Hi,

**Problem:**
It seems that UI slider for warp is not updated when the value is changed through HTTP api. Related function call is PlayerSpaceship:commandWarp(int8_t target) if I'm correct.

**Proposed action:**
There is also other similar problems or UI controls that cannot be changed through api. Would it be ok if I do some changes and make every UI control to listen for spaceship's state (and so forth save state of every UI control in spaceship's state)? Motivation behind this is a project where we're building custom hardware consoles for controlling the ship.

czenker:
What is your opinion on that, @daid? Would you consider an UI not showing correct data consider a bug?

I also experienced this behavior and would like to fix it. But I am not sure if it has any bad side effects. I am not sure what it does to performance to add more into `onDraw` and if lag could become an issue. A slider that jumps while you drag it might not be the best experience. ;) From what I remember EE was never supposed to be played outside a local network, so response times might not be an issue!?

oznogon:
This seems resolved, just tested in the 20200115 build and the slider updates

PR #694, opened 2019-09-29
Adding functionality to HTTP api and binding UI to ship state

Tuhis:
This pull request adds some new functions to HTTP api to be able the ship through it. At this point I've gone through weapons and engineering stations.

Added functions:
- commandNextTarget
- commandPreviousTarget
- commandFireTubeAtCurrentTarget
- commandCalibrateShields
- commandSetShieldFrequencySelection
- commandSetNextShieldFrequencySelection
- commandSetPreviousShieldFrequencySelection
- commandConfirmSelfDestruct
- commandSetAimLock
- commandSetAimAngle
- commandSelectWeapon
- commandSelectSystem

Also, I've bound UI state to ship's state where it wasn't yet. Related issue: https://github.com/daid/EmptyEpsilon/issues/605

CMakeLists.txt and FindSFML.cmake modifications are something that I had to do to get build working for me... I'm not sure if they should be merged or not.


daid:
I have a few issues with this pull request:

- Unrelated build system file changes
- It moves a few things server side, that should be client side in my opinion. Like system selection for the engineering station. Else if multiple people look at engineering, the selected system can change while you are adjusting things.
- Not sure why you put spaces around the ```->``` operator, I've never seen that before...

So while there are some good fixes in there, as a whole it's a bit too much.

Tuhis:
Spaces around the ”->” operator are mistake, shouldn’t be there. Also I can get rid of build system changes.

Selected system and other similar things are moved to server side to be able to control the ship through the api. Otherwise the requests must be sent to correct EE client which is a bit too tough requirement in my opinion. Is there some situation when multiple people would be looking at a single station (e.g. engineering)?

aBlueShadow:
> Selected system and other similar things are moved to server side to be able to control the ship through the api.
You don't need to remote control the station to control the ship, you can just do it directly by things like commandSetSystemPowerRequest. As system selection is a client thing, you can just implement it in your webapp or whatever you are using for the http api. Or you have just a bund of sliders, like in the power control-screen to allocate power and coolant directly.
I would put it that way: By using the http API you are not controlling a station, you are (re)making one.

daid:
> Is there some situation when multiple people would be looking at a single station (e.g. engineering)?

Yes, when there are for example 3 people controlling a ship. Multiple people might take the engineering station.

I'm always open for more APIs, and there are a few things in here that I like. But I don't like moving things server side that should be client side. This can also cause odd UI jittering with the manual targeting for example.

And, for example gui2_rotationdial.cpp suddenly knows about the player spaceship now, and that's not right for a "generic" control. Even tough it is used in only 1 spot.

PR #695, opened 2019-09-30
Custom captions for windowscreen via options.ini

aBlueShadow:
This is a compromise between #617  and #621,  
It turned out that for some scenarios, a window screen with callsigns and headings is needed. For example, @PlanetariumWSD uses it to have multiple angled ship windows to combine into a huge first-person mainscreen that is projected onto a planetarium dome.
This PR is much simpler than #617 and uses an options.ini value instead of a menu.

### DIFF
--- a/src/preferenceManager.cpp
+++ b/src/preferenceManager.cpp
@@ -45,7 +45,7 @@ void PreferencesManager::save(string filename)
         fprintf(f, "# Empty Epsilon Settings\n# This file will be overwritten by EE.\n\n");
         fprintf(f, "# Include the following line to enable an experimental http server:\n# httpserver=8080\n\n");
         fprintf(f, "# For possible hotkey values check: http://www.sfml-dev.org/documentation/2.3.1/classsf_1_1Keyboard.php#acb4cacd7cc5802dec45724cf3314a142\n\n");
-        fprintf(f, "# Values for ship_window_caption are: spacedust:1, headings:2, callsigns:4 \n# Add them for any combination. Example: ship_window_caption=7 for all three\n\n");
+        fprintf(f, "# Values for ship_window_flags are: spacedust:1, headings:2, callsigns:4 \n# Add them for any combination. Example: ship_window_caption=7 for all three\n\n");
         std::vector<string> keys;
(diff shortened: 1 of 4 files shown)


src/screens/windowScreen.cpp
```diff
@@ -8,11 +8,16 @@
...
 : angle(angle)
 {
     viewport = new GuiViewport3D(this, "VIEWPORT");
-    viewport->showSpacedust();
+    if (flag_callsigns & caption)
+      viewport->showCallsigns();
+    if (flag_headings & caption)
+      viewport->showHeadings();
+    if (flag_spacedust & caption)
+      viewport-> showSpacedust();
```
daid: extra space after ->

I rather see the check the other way around "caption & flag_spacedust"

src/preferenceManager.cpp
```diff
@@ -45,6 +45,7 @@ void PreferencesManager::save(string filename)
         fprintf(f, "# Empty Epsilon Settings\n# This file will be overwritten by EE.\n\n");
         fprintf(f, "# Include the following line to enable an experimental http server:\n# httpserver=8080\n\n");
         fprintf(f, "# For possible hotkey values check: http://www.sfml-dev.org/documentation/2.3.1/classsf_1_1Keyboard.php#acb4cacd7cc5802dec45724cf3314a142\n\n");
+        fprintf(f, "# Values for ship_window_caption are: spacedust:1, headings:2, callsigns:4 \n# Add them for any combination. Example: ship_window_caption=7 for all three\n\n");
```
daid: I would call it "window_flags" instead of "window_caption", as a "caption" is a "a title or brief explanation accompanying an illustration, cartoon, or poster"

aBlueShadow: I had another meaning of caption in my mind, that of "A piece of text appearing on screen as subtitle or other part of a film or broadcast." - or videogame, for that matter, as some games has a "caption" option. But you are right, a more general name would be better as it is not exactly the same as a subtitle and also does not cover the space dust

src/screens/windowScreen.cpp
aBlueShadow: Okay, changed both and also replaced all "caption"s by "flag"s

Issue #624, opened 2019-02-05
Feature Request: Remember Last IP Address


pixelclef:
Having to enter the same IP address over and over in a sequence of games - or even when trying to connect to one game - is a little frustrating. Can this be remembered and prefilled?

PR #697, opened 2019-10-09
Issue #624 - Remember last connection

techfreek:
In the event of rejoining a server, or getting disconnected, this will pre-populate one of the servers in the list with the last server you connected to. I didn't add a check to see if the server is still valid, just cache the IP.


daid:
Simple enough. Saves manual IP entry for some people and has little effect on others.

PR #700, opened 2019-10-12
Issue 629 - Don't timeout connections while waiting for auth

techfreek:
Previously, during server auth, there are no packets sent back and forth between the client and server, which causes the client to timeout after 20 seconds (assuming this is in case the server has shutdown/crashed/etc). As reported in issue 629, this may not be long enough. So combined with https://github.com/daid/SeriousProton/pull/37, this will send keep alives to the server (which will re-request auth) to ensure the connection stays alive. This implementation has the client act as the initiator of the keep alive, not the server. This prevents adding extra load to the server.


techfreek:
Based on feedback in the SeriousProton PR, I'm closing this

PR #701, opened 2019-10-12
Remove deprecated sf::Text::setColor calls

techfreek:
[sf::Text::setColor](https://www.sfml-dev.org/documentation/2.5.0/classsf_1_1Text.php#afd1742fca1adb6b0ea98357250ffb634) was deprecated and causes compilation warnings (not failures). This replaces calls setFillColor and setOutlineColor instead to avoid the warnings

This seemed like the sanest approach, but it does duplicate lines, so I'm miffed. Figured I'd open as draft to get ideas.


daid:
This will make EE incompatible with debian stretch which is on SFML2.3 (the replacement functions where introduced in 2.4)

I think our setup is running on stretch, so I rather not break that (as I'm currently away from our setup, and won't have time to update it)

techfreek:
I figured was fearing something like that might happen. I guess either we live with the deprecation warnings, or I could add compile time checks for the version number: https://github.com/SFML/SFML/blob/master/include/SFML/Config.hpp#L32-L33

daid:
We'll live with the warning for now, but keep this pull request active for once I get to check&update our system.

PR #708, opened 2019-12-09
Correct spelling of "coming" in a couple of scenarios

dorward:


daid:
Dyslexia strikes again!

PR #714, opened 2019-12-29
Fix Hue lights unresponsiveness when using multiple bulbs, and add Transition between states

k9lego:
**Changes**
This also adds caching for light state (transitiontime, brightness, saturation, hue) and only makes a request to the Hue hub when the lights change status. This is to allow lights to be more responsive when a change is actually needed. 
It also adds a 4th channel for Hue Transitions, which is useful to fade multiple lights from one state to another so the Hue update delay per light is unnoticeable. Without this, using multiple lights is uncomfortable.
**Stability**
This has been tested on Ubuntu 18.04 LTS, I built it from sources. I did a whole 90 minute game with friends and it worked as expected. I also tested it outside of this. It also worked on another Debian-based Distribution. I have not tested this on Windows, but these changes shouldn't affect the Windows build any differently.

Please let me know if any of my code is not to convention.

### DIFF
--- a/src/hardware/devices/philipsHueDevice.cpp
+++ b/src/hardware/devices/philipsHueDevice.cpp
@@ -171,22 +171,23 @@ bool PhilipsHueDevice::configure(std::unordered_map<string, string> settings)
 
 void PhilipsHueDevice::setChannelData(int channel, float value)
 {
-    int light_idx = channel / 3;
+    int light_idx = channel / 4;
     if (light_idx < 0 || light_idx >= light_count)
         return;
 
     sf::Lock lock(mutex);
-    switch(channel % 3)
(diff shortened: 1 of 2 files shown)


daid:
Looks good. I do not have hue hardware to test on. So i trust you that this works.

k9lego:
I just re-built this for Linux and it worked, I guess the Windows build failed? Looks like a Int-Float mismatch. I'll open another pull request when I figure out what went wrong.

daid:
Only the intermediate commit failed to build. My CI system builds every single commit, not just the latest one. You missed some header changes in the first commit and those where added in the 2nd commit. The latest commit has all lights green.

k9lego:
Alrighty!

PR #716, opened 2020-01-07
Log planet radius to game state log

oznogon:
- Add a getPlanetRadius function to return the value, and also exposes it for scripting.
- Recognize Planets in the game state logger and log their radius.
- Render the radius in the game state viewer. This renders planets that are not on the same plane as the players (ie. cosmetic Planet objects, like the moon in the Empty scenario) that do _not_ appear in crew radars.

screenshot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Added
 
 - `SpaceShip:setScanState()` and `setScanStateByFaction()` scripting functions.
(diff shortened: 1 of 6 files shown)


daid:
Looks good. Some people might also be interested in the collision radius, instead of the rendered radius?

oznogon:
@daid makes sense; I can follow up with that

PR #722, opened 2020-01-11
Allow setting custom master registry server URLs

oznogon:
Adds two options, `registry_registration_url` and
`registry_list_url`, that allow the URLs to the registry server
to be configured. Defaults to the existing behavior
(`http://daid.eu/ee/list.php` and `register.php)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 - _Shoreline_ scenario refactored to fix errors.
 - Add planet radius to GM screen scripting exports.
+- Added `registry_registration_url` and `registry_list_url` to optionally use a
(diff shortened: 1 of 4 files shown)


oznogon:
Worth noting that the default `daid.eu` server returns 404s, and I haven't yet mimiced the responses and behaviors of a real server well enough to test whether this would work. The registry server response formats and POST data seem straightforward: https://gist.github.com/oznogon/c6c80a0e3432669b6e85857c117f9ad4

daid:
The code of the masterserver is at https://github.com/daid/SeriousProton/tree/master/masterserver

daid:
Not sure why it is down. Cannot check right now on the server

PR #725, opened 2020-01-14
Move signal names to playerSpaceship and let scripts set them

### LINKED ISSUE #534: [Request] Add specific label for Scan Dialog
Is it possible to customize the scan dialog with some label instead of "Radio profile" or "Raspberry jam prevention" ?

Something like obj:setScanLabel(text)
If this is null, then EE use one of the random label. If not, it use the custom label.

oznogon:
Resolves #534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `SpaceShip:setScanState()` and `setScanStateByFaction()` scripting functions.
 - `Planet:getPlanetRadius()` and `Planet:getCollisionSize()` scripting functions.
+- `PlayerSpaceship:setSignalLabels()` scripting function.
 - Log planet radius to the game state log and render it in the viewer.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


daid:
Wouldn't it make more sense to optionally get labels from the target?
Also, registerMemberReplication is missing, so in this state it will not work for clients

oznogon:
> Wouldn't it make more sense to optionally get labels from the target?

My thinking was that this was a player ship-specific change (the ships have these cosmetic "features" like "jam prevention", rather than the target), and it seemed generally like something relevant only to GMs and mission writers who want to control the text. But I can move it to spaceship and request it from there instead, it'll solve the same problem and be more flexible.

> registerMemberReplication is missing, so in this state it will not work for clients

Weird, I thought I'd tried it on a non-server client, but in that case it wouldn't have made sense. Thanks for catching that, I'll try to straighten that out.

Issue #14, opened 2015-02-20
Sensor fingerprints for space objects


IvanSanchez:
This is an idea from playing too much Allegiance and UFO Afterlight, and from working with real radar&lidar systems.

The sensors system is quite naïve: a spaceObject is either seen or not seen, and "sensors" is just one monolithic system. In real life, targets are acquired by aggregating input from several sensor systems: radar, lidar, AIS, FoF, cameras, and what-have-you.

Also, some objects (e.g. planets, stations) are more difficult to detect than others (fighter ships). This is the sensor fingerprint.

I don't want to reinvent the Science UI but rather set foundations for this.

So every space object should have several attributes:

- Mass, measured in tons. Massive objects will show up in gravimetric sensors (and, yes, gravimetry is a real thing https://en.wikipedia.org/wiki/Gravimetry ).
- Reflectivity, measured in percent. AKA "if I throw some radiation at the thing, how much will I get back?". Rough surfaces (planets) have high reflectivity, and stealth ships have low reflectivity. Active sensors (such as radar) operate under this concept.
- Radiance, measured in watts. AKA "how much energy is the thing passively emitting away". Shielded spaceships should have low radiance. Things like warping and firing beams would momentarily increase radiance. Passive sensors (such as video cameras) operate under this concept.
- Psi, AKA "intelligent lifeforms", measured in sapiens (1 sapien = the brain of a standard homo sapiens). Futuristic alien sensors might be able to detect Psi, because UFO Afterlife had "Psi Vision" http://www.ufo-afterlight.com/images/scr_10.jpg . And maybe some aliens are attracted/repelled by Psi.

If we want to go crazy, radiance and reflectivity can be split into several parts of the EM spectrum: microwave, visible and x-ray.

I just want my science officer to yell "Contacts in the gravimetric sensor, activating microwave emitters!" :-)

nallath:
Ooh! Good point. I do think we need to have a simple mode (which is the current mode) and the advanced mode, which pretty much does all of the above.

This also enables a bit more 'silent hunter' type of play, where you try to be as invisible as possible (flying slower, disabling certain systems, etc). 

I also think that adding stuff like probes is an excellent way of doing more with sensors.

daid:
[...]

PR #726, opened 2020-01-14
Change ship signatures based on system states

oznogon:
Ship signatures can change based on their heat, power level, and
usage.

- Systems that generate heat increase the biological band; coolant
  offsets this effect even if the system maintains heat.
- Systems that use energy increase the electrical band.
- Ships that are jumping or warping spike the gravity band.

This allows Science to predict when a ship is going to warp/jump,
and in stealth missions can give away the positions of ships that
run hot.

CPU ships do not manage heat/energy but still exhibit jump/warp
spikes.

Example of a jump spike before and after:

jumping
jumping2

Progress toward #14

### DIFF
--- a/src/screenComponents/rawScannerDataRadarOverlay.cpp
+++ b/src/screenComponents/rawScannerDataRadarOverlay.cpp
@@ -62,6 +62,46 @@ void RawScannerDataRadarOverlay::onDraw(sf::RenderTarget& window)
         // Get the object's radar signature.
         RawRadarSignatureInfo info = obj->getRadarSignatureInfo();
 
+        // If the object is a SpaceShip, adjust the signature dynamically based
+        // on its current state and activity.
+        P<SpaceShip> this_ship = obj;
+
+        if (this_ship)
+        {
+            RawRadarSignatureInfo signature_delta;
+
+            // For each ship system ...
+            for(int n = 0; n < SYS_COUNT; n++)
+            {
+                ESystem this_system = static_cast<ESystem>(n);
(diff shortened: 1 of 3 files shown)


tdelc:
I like this PR, bit do you test it with old computers ? I have huge lags
with my old PCs for the science station and I am afraid it could be worse
with your feature.

daid:
Dont think it will effect performance at all. But it does break setting custom signatures from scripting i think.

oznogon:
Ah, good point @daid — I'll revise.

daid:
I also think this causes unneeded network traffic. As the signature is calculated on clients now as well as send to clients.

oznogon:
If I understand correctly, it should be straightforward to move the loop into an `if (game_server)` condition to fix the network redundancy?

oznogon:
Updated the branch to move the logic to the RawScannerDataRadarOverlay component and changed the behavior on further testing.

- Coolant use can no longer result in negative bio band values.
- Jump drive grav spikes start small and grow larger closer to the jump.
- Jump drive recharge state causes an electrical spike that drops
  the closer the system gets to being fully recharged, and scales to the
  system's power level.
- A ship's electrical signature can be reduced by underpowering its systems.

The baseline radar signature is also slightly lower now to add some contrast between an idle and active ship.

daid:
I was thinking this logic could also be moved to the "getRadarSignatureInfo" function, which could have been made virtual then and overloaded in the spaceship. However, that would mean the set and get functions would return different data, which is odd. So I've accepted this solution.

PR #727, opened 2020-01-14
Fixed game completion condition for Minesweeper

Triqueon:
This fixes an issue where the Minesweeper hacking is successful even if you click on a bomb if more than 50% of the tiles are revealed.


daid:
Thanks for the fast fix!

PR #732, opened 2020-01-21
Edit SpaceObject scripting docs

oznogon:
- Document, expand existing docs, and add examples to SpaceObject scripting functions.
- Reorder functions in the script registry for consistency (grouping by type, setters before getters)
- Clarify the comments and output of the compile_script_docs script

See [oznogon/EmptyEpsilon-test-scenarios/scenario_00_spaceObject.lua](https://github.com/oznogon/EmptyEpsilon-test-scenarios/blob/master/scenario_00_spaceObject.lua) for in-scenario usage to try to confirm the examples.

### DIFF
--- a/compile_script_docs.py
+++ b/compile_script_docs.py
@@ -1,7 +1,12 @@
 # Python script that generates the documentation for script functions.
-# This parses the codeblocks for to find all the source files, then parses all the source files for script macros.
-# Finally it outputs all the documentation to stdout so it can be stored on disk.
-# This script should run in both python2 and python3
+#
+# This parses the codeblocks to find all the source files, then parses
+# all of the source files for script macros.
+#
+# It then outputs all the documentation (comments starting with `///`)
+# as HTML to stdout so it can be stored on disk.
+#
+# This script should run in both Python 2 and 3.
 import re
 import os
 import sys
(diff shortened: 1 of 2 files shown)


oznogon:
Some of the functions don't have many examples in the wild, particularly takeDamage. I'm not sure how to set the energy damage frequency or ESystem target, for instance.

daid:
See for how the parameters of takeDamage are handled: https://github.com/daid/EmptyEpsilon/blob/master/src/spaceObjects/spaceObject.cpp#L405

```obj:takeDamage(50, "kinetic", 0, 0, 0, "warp");```
Will damage the warp drive I think.
The beam weapon frequencies are a bit odd. As the displayed frequency is: "400 + (frequency * 20)"
https://github.com/daid/EmptyEpsilon/blob/5515686db4ffa942793816fb9ac5b6a0850e5352/src/spaceObjects/spaceship.cpp#L1324
So ```obj:takeDamage(50, "energy", 0, 0, 4);``` should try to do 50 energy beam damage on frequency 480.

(note, didn't test this)

**Comment by daid** [comment]
Still, great work!

**Comment by oznogon** [comment]
Thanks — I think I trip up on the DamageInfo constructor's defaults (-1 for frequency and SYS_None for system_target). ie. if I don't want to set a frequency but do want to set the target system, I _think_ I'd expect to pass:

```
obj:takeDamage(50, "kinetic", 0, 0, -1, "warp");
```

daid:
I think frequencies are ignored for anything that is not energy damage anyhow.

Issue #548, opened 2018-07-17
Cloaking Implementation Idea


RayvenMaker:
Here's an idea: what if cloaking was more of a radar jammer? Most of the rules that EVE ship simulator  #547 lays out would still apply, but it might not come across as a glitch.`

If a ship is currently is "cloaked", it would appear either as a large, faint blur on the radar display or as a single ship that randomly glitches around the general area that the ship might be. In Helms, Weapons, and Science, if someone attempts to select this ship, it'll show an error saying that this ship is "cloaked". However, in Science, there is an option to run a deep scan (the hard, two-step kind of scan) on that "cloaked" ship. This will give the "jamming frequency" that the cloaked ship is using. Then Science tells this frequency to Relay, who can broadcast a destructive frequency that cancels out the ship's jamming frequency. Once that's done, the ship is completely visible... for as long as Relay keeps on broadcasting that decloaking frequency. Both the process of broadcasting jamming frequencies and broadcasting anti-jamming frequencies would require power and would prevent Relay from doing anything except communicating with other ships.

If you're trying to cloak your own ship, you simply set your jamming frequency in Relay, and have Relay broadcast that frequency out. Of course, you can only cloak if you have a ship capable of cloaking. 

This would give another job to Relay (which would be nice), as well as show an obvious clue to the players that the ship is actually cloaked, and that it isn't just glitching out. This also retains the tactical advantage of cloaking, because although it's obvious that you're there, the enemy has no idea exactly what, where, and who you are. Honestly, I'd love to see this implemented in Empty Epsilon.

// RayvenMaker

Issue #16, opened 2015-02-24
Missile lock-on ability and indicator


IvanSanchez:
Allow missile weapons to lock onto targets with a high enough sensor fingerprint (see #14), or on targets scanned by the Sci console. It should be impossible to lock onto "cloaked" ships, or ships too far away, or ships not scanned by Sci. 

If the weapons console is targeting a enemy vessel, display an indicator as to whether a missile can or cannot lock onto that target.

Locking the missile weapons onto a target should take a few seconds (because the sensor fingerprint is being synchronized with the missile sensor pack and technobabble technobabble technobabble technobabble). Blinky "lock" indicator = good.

PR #742, opened 2020-02-03
[PROPOSAL/POC] Detailed signal visualization on Science

oznogon:
Adds toggle to crew6 science (intentionally only this station, for now) to show circular and numerical visualizations of radar signal values. This also exposes VisualAsteroids, and objects with no radar trace but a radar signal value.

ezgif com-optimize

Moves the dynamic radar signal visualization previously implemented in rawRadarScannerDataOverlay to a new getter in Spaceship, `getDynamicRadarSignatureInfo`. This makes these dynamic values available to other functions without duplication. Using a separate getter should avoid the potential discrepancy of overloading the spaceObject's `getRadarSignatureInfo` getter and getting different values than the setter's input.

Exposes these component dynamic values to scripts via the `getDynamicRadarSignatureGravity`, `...Electrical`, and `...Biological` functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - `ScanProbe:onExpiration`, `ScanProbe:onDestruction`, and
   `PlayerSpaceship:onProbeLaunch` callback scripting functions.
 - Scan object and cycle selected object hotkeys added to Science.
+- `Spaceship:getDynamicRadarSignatureGravity`, `...Electrical`, and
+  `...Biological` scripting functions.
+- Detailed radar signal data visualizations as a toggleable view on the Science
+  station's radar.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,19 @@
 
 ### Added
 
-- `ScanProbe:onExpiration`, `ScanProbe:onDestruction`, and
+- `ScanProbe:onExpiration`, `...onDestruction`, and
   `PlayerSpaceship:onProbeLaunch` callback scripting functions.
 - Scan object and cycle selected object hotkeys added to Science.
 - `Spaceship:getDynamicRadarSignatureGravity`, `...Electrical`, and
   `...Biological` scripting functions.
+- `SpaceObject:setRadarSignatureGravity`, `...Electrical`, and `...Biological`
+  scripting functions.
+- `SpaceObject:setVisibility` and `...isVisible` scripting functions, allowing
+  SpaceObject visibility to be toggled.
 - Detailed radar signal data visualizations as a toggleable view on the Science
   station's radar.
+- GM tweak features for radar signatures.
+- Scan object, cycle selected object, and signal detail hotkeys added to Science.
 
 ### Changed
 
--- a/src/gui/gui2_button.cpp
+++ b/src/gui/gui2_button.cpp
@@ -6,14 +6,18 @@
 GuiButton::GuiButton(GuiContainer* owner, string id, string text, func_t func)
 : GuiElement(owner, id), text(text), func(func)
 {
+    override_color = false;
+    override_text_color = false;
     text_size = 30;
 }
 
 void GuiButton::onDraw(sf::RenderTarget& window)
 {
-    sf::Color color = selectColor(colorConfig.button.background);
-    sf::Color text_color = selectColor(colorConfig.button.forground);
-    
+    if (!override_color)
+        color = selectColor(colorConfig.button.background);
+    if (!override_text_color)
+        text_color = selectColor(colorConfig.button.forground);
+
     if (!enabled)
         drawStretched(window, rect, "gui/ButtonBackground.disabled", color);
     else if (active)
@@ -71,9 +75,11 @@ void GuiButton::onMouseUp(sf::Vector2f position)
     }
 }
 
-string GuiButton::getText() const
+GuiButton* GuiButton::setColor(sf::Color color)
 {
-    return text;
+    this->override_color = true;
+    this->color = color;
+    return this;
 }
 
 GuiButton* GuiButton::setText(string text)
@@ -88,12 +94,24 @@ GuiButton* GuiButton::setTextSize(float size)
     return this;
 }
 
+GuiButton* GuiButton::setTextColor(sf::Color color)
+{
+    this->override_text_color = true;
+    this->text_color = color;
+    return this;
+}
+
 GuiButton* GuiButton::setIcon(string icon_name, EGuiAlign icon_alignment, float rotation)
 {
     this->icon_name = icon_name;
     this->icon_alignment = icon_alignment;
     this->icon_rotation = rotation;
     return this;
+}
+
+string GuiButton::getText() const
+{
+    return text;
 }
 
 string GuiButton::getIcon() const
--- a/src/gui/gui2_button.h
+++ b/src/gui/gui2_button.h
@@ -11,6 +11,10 @@ public:
 protected:
     string text;
     float text_size;
+    bool override_color;
+    sf::Color color;
+    bool override_text_color;
+    sf::Color text_color;
     EGuiAlign text_alignment;
     func_t func;
     string icon_name;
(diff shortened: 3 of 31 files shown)


daid:
Note sure if I like the extra visual "clutter", and if it adds something or takes something away.

But, maybe the numbers next to the ships work better as 3 bars (like the health bars in the GM screen)

Also, what happens with ships behind nebula? Are these "visible" now?

daid:
Another thought, maybe the science screen should be able to switch between multiple view modes, like for example "visual" "gravity" "organic" and only show 1 thing at a time, instead of trying to show all the information at once. Give the player some choice and uncertainty (I'm looking at gravity, but might miss an object which doesn't emit a strong gravity signal)

oznogon:
I've roughly implemented visual/grav/elec/bio lenses, and to cut down on clutter I've limited the specific numbers to display only on the currently selected scan target (and can now also be viewed on any scannable object; they don't require a successful scan to view).

Unfortunately I also have a rather bad Access Violation crash bug whenever:

- the visual lens is disabled
- any of the grav, elec, or bio lenses are enabled
- any object within Science radar range gets sucked to the center of a black hole

```
EmptyEpsilon.exe caused an Access Violation at location 005026A8 in module EmptyEpsilon.exe Reading from location 00000000.

Registers:
eax=00000000 ebx=462d0a7c ecx=0147dfb0 edx=00000000 esi=c5ae8dbd edi=0147dff0
eip=005026a8 esp=0147ddf0 ebp=0147e008 iopl=0         nv up ei pl nz na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00010206

AddrPC   Params
005026A8 11BC0590 11BC0380 FFFFFFFF  EmptyEpsilon.exe!GuiRadarView::drawSignalDetails
004FBCB0 016799F8 00000000 44960000  EmptyEpsilon.exe!GuiRadarView::onDraw
0054F771 00000000 00000000 44960000  EmptyEpsilon.exe!GuiContainer::drawElements
0054F7A8 00000000 00000000 44960000  EmptyEpsilon.exe!GuiContainer::drawElements
0054F7A8 00000000 00000000 44960000  EmptyEpsilon.exe!GuiContainer::drawElements
0053CA47 016799F8 00000001 0147E70C  EmptyEpsilon.exe!GuiCanvas::render
005C4044 016799F8 00000001 0147E76C  EmptyEpsilon.exe!RenderLayer::render
005C3FE5 016799F8 C51412BF 0147E7C8  EmptyEpsilon.exe!RenderLayer::render
005C3FE5 016799F8 FFFFFFFF 0147E7D8  EmptyEpsilon.exe!RenderLayer::render
005C3FE5 016799F8 000000A7 11B1FBA8  EmptyEpsilon.exe!RenderLayer::render
005B612C 016799F8 07058E38 683835A0  EmptyEpsilon.exe!PostProcessor::render
005B612C 016799F8 00000014 00000014  EmptyEpsilon.exe!PostProcessor::render
00612B27 3C904923 007A6FA4 007A6720  EmptyEpsilon.exe!WindowManager::render
006124D9 0147FB98 0147FBC8 0076233E  EmptyEpsilon.exe!Engine::runMainLoop
00404087 01FDD538 00000010 00000003  EmptyEpsilon.exe!main
004013E3 00215000 76796340 0147FFDC  EmptyEpsilon.exe!__tmainCRTStartup  [./mingw-w64-crt/crt/crtexe.c @ 334]
76796359 00215000 37B162EA 00000000  KERNEL32.DLL!@BaseThreadInitThunk@12
77B07B74 FFFFFFFF 77B28F0F 00000000  ntdll.dll!__RtlUserThreadStart
77B07B44 004014E0 00215000 00000000  ntdll.dll!__RtlUserThreadStart@8
```

This doesn't happen if the visual lens is enabled, if all lenses are disabled, or if signal details are disabled. I've run out of brain cells for the day to try to figure it out.

oznogon:
Realized I was effectively duplicating chunks of drawObjects, so I just moved the signal details drawing into there. Still need to clean things up and test, but the crashes have since stopped.

An interesting side effect in testing is that, since every spaceObject has a default radar signature, this picks up everything — including beam firing, beam impacts, missiles, and explosions.

out

oznogon:
-   Changed the circle-drawing behavior from separate transparent circles for each band, to mixing the colors and combining their effects when more than one band is selected. This also reduces the number of draw calls. Signal circles are now opaque.

    radar-colors
-   Changed or added radar signature values to several space objects, including explosions, so they show up on the detailed signal views. For example, EMP explosions have large electrical signatures, Homing missiles have small bio (heat) signatures, HVLI slugs have small gravity (density) signatures, default supply drops modify their signatures based on their contents, etc.

    radar-signatures
-   Impulse engine usage increases bio (heat) signatures, and active shields increase electrical signatures.
-   Signal details now toggle their visibility and state more consistently, including when Science switches from Radar to Database views.
-   Briefly attempted to tie radar trace visibility to signal strength, but this needs more work.

There's still potential here for a "cloaking device" or slow/quiet running to allow ships to hit 0 or negative signal values as stealthy radar-avoidant ships and objects, or to hook missile tracking to signals in a manner that could let ships spoof a signal/drop a "flare", or cut power or duck into a nebula, out of desperation to hide from a missile. But I think any more detailed work toward those things (ie. a lot of the ideas in #14) depends on whether a system like this is right/good for EE.

oznogon:
Some of the gameplay mechanics this has enabled in brief 3-player testing:

-   Science can now spot when other ships are engaged in combat by looking for fluctuations in different signature bands on the long-range scanner.
-   By setting a specific signature value on a scannable object in a scenario, I can use mission text or dialogue to require Science to identify an objective based on the numbers in its signature ("scan only the mine with a signature of 400/700/800; any other mine will explode when scanned"; "pick up only the energy supply drop, but leave the weapons drops behind for the resistance. We can't tell you how to identify the energy drop, you'll have to improvise")
-   Science can guess when another player ship's shields have been raised or lowered by sudden fluctuations in its electrical signature. Potential here to adjust signatures further based on whether weapon tubes are un/loaded.

oznogon:
-   Objects with negative values on all bands do not render radar traces on visual mode.
-   Signal bands in detail view are recolored on feedback to make them more mnemonic: red for electrical (radiation), green for gravity, blue for biological.
-   On the ops screen, the buttons are moved to the top left to avoid colliding with comms and waypoints.

oznogon:
-   Signal color combinations are more proportional, and intensity scales to signal value.
-   Radius scales to the signal value only if the value is > 1.0.
-   Fewer circle calls on objects with no signal or radius.

oznogon:
- Objects with 0 or negative signal values cannot be targeted or selected.

There's no current way a ship can reach all zero or negative signal values in play without invoking scripting functions to set its baseline values to below 0.

If a ship is targeted or selected when its signals drop to 0 or less, it remains selected/targeted until unselected.

oznogon:
This covers pretty much everything I had wanted to do with this system. As noted there are still lots of ways to extend or apply this system further to accomplish #14, #16, and #548.

daid:
I don't think the fully saturated colors look very nice in the circles.
And maybe we can give the G/E/B buttons the same colors as the signals? As that would provide some visual indication what they mean. Currently they are single letter buttons and people have no clue what they will do without pressing them.

oznogon:
-   Desaturated the circle colors.
-   Added setter functions to GuiButton in order to assign the signal lens buttons custom colors. I'm not sure if there's a better way to do this.
-   Fixed a regression in targeting low-signal targetable objects.
-   Enabled detailed signal view on probe radar.

daid:
As for the button colors, seems like a decent solution. Another option would be to expand the "colorConfig" and have the option for buttons to use a different colorconfig entry (that's sort of how I did it in SeriousProton2. Where widgets can have different styles assigned to them and those define the colors and other things)
But I fully understand if that becomes too complex too fast.

daid:
How do the new colors look now?

oznogon:
colors

oznogon:
- Add hotkeys for toggling signal details (<kbd>W</kbd>) and lenses (<kbd>E</kbd>, <kbd>G</kbd>, <kbd>B</kbd>).

oznogon:
- Object visibility is separated from radar signals and can be toggled via scripting (`setVisibility(bool)`, `bool isVisible()`). This allows ships to cloak without dampened signals, which means Science could see an invisible object's radar signature. This also toggles visibility in 3D rendered viewports.
- Detailed signal numbers stabilize when scanned or are within short-range sensor range (5U).
- GMs can tweak objects' baseline radar signatures in a new Radar tweak menu. The on-open values on the sliders is broken at the moment.

oznogon:
I think I've got the proof of concept together. I'll revise and decompose this into components and refile.

PR #743, opened 2020-02-10
Combat maneuver hotkeys

aBlueShadow:
This adds hotkeys for combat maneuvering (which were already present as a placeholder).

_Unfortunately D is already predefined, and S could be accidentally pressed, otherwise defaulting them to WAD would have been nice._

You may wonder why I didn't use the onHotkey events. I did at first, and then realized I also need a way to recognize whether the keys were released, to stop the maneuver. So I had to manually check anyway, and checking via onHotkey became redundant.

### DIFF
--- a/src/screenComponents/combatManeuver.cpp
+++ b/src/screenComponents/combatManeuver.cpp
@@ -3,12 +3,133 @@
 #include "combatManeuver.h"
 #include "powerDamageIndicator.h"
 #include "snapSlider.h"
-
+#include "preferenceManager.h"
 #include "gui/gui2_progressbar.h"
 
 GuiCombatManeuver::GuiCombatManeuver(GuiContainer* owner, string id)
 : GuiElement(owner, id)
 {
+	static std::vector<std::pair<string, sf::Keyboard::Key> > sfml_key_names = {
+	    {"A", sf::Keyboard::A},
+	    {"B", sf::Keyboard::B},
+	    {"C", sf::Keyboard::C},
+	    {"D", sf::Keyboard::D},
+	    {"E", sf::Keyboard::E},
+	    {"F", sf::Keyboard::F},
+	    {"G", sf::Keyboard::G},
+	    {"H", sf::Keyboard::H},
+	    {"I", sf::Keyboard::I},
+	    {"J", sf::Keyboard::J},
+	    {"K", sf::Keyboard::K},
+	    {"L", sf::Keyboard::L},
+	    {"M", sf::Keyboard::M},
+	    {"N", sf::Keyboard::N},
+	    {"O", sf::Keyboard::O},
+	    {"P", sf::Keyboard::P},
+	    {"Q", sf::Keyboard::Q},
+	    {"R", sf::Keyboard::R},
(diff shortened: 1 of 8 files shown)


src/screenComponents/combatManeuver.cpp
```diff
@@ -3,12 +3,133 @@
...
 #include "powerDamageIndicator.h"
 #include "snapSlider.h"
-
+#include "preferenceManager.h"
 #include "gui/gui2_progressbar.h"
 
 GuiCombatManeuver::GuiCombatManeuver(GuiContainer* owner, string id)
 : GuiElement(owner, id)
 {
+	static std::vector<std::pair<string, sf::Keyboard::Key> > sfml_key_names = {
```
daid: This looks like a copy&paste from the hotkey code. To prevent duplication the hotkey class could have a static function (or 2) to convert keynames from/to string.

src/screenComponents/combatManeuver.h
```diff
@@ -11,6 +11,12 @@ class GuiCombatManeuver : public GuiElement
 private:
     GuiSnapSlider2D* slider;
     GuiProgressbar* charge_bar;
+	string combat_left_key;
```
daid: As these 3 strings are only used during construction, there is no need for them to be members of the class, but can be defined inside the function itself.

aBlueShadow: Yeah, saw that shortly after I submitted the PR. done.

src/screenComponents/combatManeuver.cpp
aBlueShadow: Yes it is. I thought about making that vector global in some way, but figured it could be a bit overkill for using it in just one file.
Will look into do a conversion function.

daid: It's never overkill to remove code duplication. (Not saying the whole hotkey part is very well designed anyhow)

aBlueShadow:
Turned out to be a bit more tricky than "just add some hotkeys", but possible side effects should be prevented now.

src/screenComponents/combatManeuver.cpp
```diff
@@ -29,6 +37,24 @@ GuiCombatManeuver::GuiCombatManeuver(GuiContainer* owner, string id)
...
+			strafe_amount=-1.0;
+		}
+		if (sf::Keyboard::isKeyPressed(combat_right_keycode)){
+			strafe_amount=1.0;
+		}
+		if (sf::Keyboard::isKeyPressed(combat_boost_keycode)){
+			boost_amount=1.0;
+	}
+		my_spaceship->commandCombatManeuverStrafe(strafe_amount);
+	my_spaceship->commandCombatManeuverBoost(boost_amount);
```
daid: You should only call the "command" functions if the strafe/boost amount changed, else you are generating a constant stream of network traffic (as the command functions generate network traffic)

aBlueShadow: I see. Fixed in efdcba0b1542522e313e424c023238c1a41fe82a

PR #744, opened 2020-02-11
Add scan object hotkey to Science

oznogon:


Xansta:
Are the keyboard shortcuts to scan and to cycle to the next object configurable in the options.ini file like the other keyboard shortcuts? Looks like the scan key defaults to "S" and the cycle to next object defaults to "C"

aBlueShadow:
Yes, the are configurable. That's the purpose of hotkeyConfig.cpp. and the newkey methods add a new configurable hotkey, optionally with a default.  (And in this case newCategory was needed as well, as there was no science category yet)

PR #746, opened 2020-02-13
Toggle virtual keyboard on all text fields

### LINKED ISSUE #392: No keyboard appearing in mobile
Cannot type any inputs into the mobile version anywhere on the client or server creation screens. (dunno if this is the right place to even report this) 

Using 6.0.1 on Nexus 5. My friend has the same problem; she's running 6.0.1 on a Note 7.

oznogon:
Attempt to resolve #392 by toggling SFML's virtual keyboard function whenever a text entry element has focus. This is maybe overkill?


daid:
I took a slightly different method:
https://github.com/daid/EmptyEpsilon/commit/4d3fdd2fe50566796b9196222c12535ff8b2d32a

As this will also hide the keyboard when you can no longer enter text. And it won't keep calling the show function repeatedly, which I'm not sure if is a good idea.

PR #753, opened 2020-02-22
Add red, green, blue colorConfig profiles for buttons

oznogon:
- Add WidgetColorSets for red, green, and blue buttons
- Add color codes for those sets to colors.ini
- Add `setColors()` and `getColors()` functions to GuiButton to configure a button's colorSet
- Set the self-destruct button's colorSet to Red

Supports re-implementing the signal details PR. Default behavior/colorSet is unchanged.

buttoncolor


daid:
Conflict, otherwise looks good

PR #756, opened 2020-02-23
Allow spaceObject visibility to be toggled

oznogon:
Add `setVisibility()` scripting function to spaceObjects to toggle their visibility on radar and in 3D rendering, and `isVisible()` (returns bool) to check visibility in scripting.

Invisible objects can't be targeted, and invisible ships don't leave a trail of ghost dots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - `ScanProbe:onExpiration`, `ScanProbe:onDestruction`, and
   `PlayerSpaceship:onProbeLaunch` callback scripting functions.
 - Scan object and cycle selected object hotkeys added to Science.
+- SpaceObjects can be invisible, hiding them from radar and 3D views.
+- `SpaceObject:setVisibility` and `isVisible` scripting functions.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


daid:
Not sure now clear it is to players they are dealing with an invisible ship and not a bug...

oznogon:
Setting visibility is exclusively via scripting and not in any stock ship templates or features, so explaining it is left to the scenario author or GM for now, pending something like the signal detail visualization PR.

With that said, there are still uses for intentionally invisible objects that players don't have any means of detecting, such as a custom script function for a playership to toggle a cloak, or for a scenario author to make a comms contact's object invisible/untargetable until a scenario condition is fulfilled (current workarounds include moving the object to/from distant coordinates in order to "hide" them).

aBlueShadow:
> And invisible ships don't leave a trail of ghost dots.

But wouldn't that be a great way to detect invisible ships by experienced players?
Detecting cloaked ships by detecting exhaust / warp signatures is also a common trope.

oznogon:
Ghost dots aren't drawn on all relevant stations, and rendering dots on an invisible object without a radar trace looks and feels more like a bug to me.

More broadly, this is the chicken/egg problem of rebuilding the signal details PR more atomically: https://github.com/daid/EmptyEpsilon/pull/742

I thought the value of adding this as a scripting-only option was worth it, but I'll close this and re-implement it after re-implementing signal details.

PR #760, opened 2020-02-24
Mirror model rendering

### LINKED ISSUE #294: Text on space stations textures are mirrored
The text is mirrored on the textures. Not sure if it's the original textures themselves or the way EE handles them. 
This has been around probably since the space stations were added. Not a huge deal, but, might make things look more polished.

oznogon:
Scale models across the Y axis, which also inverts winding order. To mitigate the inverted order, cull front faces instead of the default back faces and invert the skybox.

Fixes #294

### DIFF
--- a/src/mesh.cpp
+++ b/src/mesh.cpp
@@ -173,10 +173,10 @@ Mesh* Mesh::getMesh(string filename)
         {
             ret->vertices[n].position[0] = -vertices[indices[n].v].x;
             ret->vertices[n].position[1] = vertices[indices[n].v].z;
-            ret->vertices[n].position[2] = vertices[indices[n].v].y;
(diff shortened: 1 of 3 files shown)


oznogon:
I don't know enough about OpenGL programming to confirm whether this is a good fix, but it does appear to flip the models.

Before:

reverse-bad

After:

reverse-good2

daid:
I think fixing the mesh loader function would be a better solution, as OpenGL state changes are generally "expensive" (not that EE is very good at not doing a lot of state changes)
https://github.com/daid/EmptyEpsilon/blob/master/src/mesh.cpp#L174

oznogon:
@daid Tried it again by flipping coordinates in the mesh loaders and moving the front-facing orientation from the default counterclockwise to clockwise in viewport3d and rotatingModelView.

PR #763, opened 2020-02-27
Detailed signal visualization

oznogon:
- Visualize SpaceObject electrical/gravity/biological radar signal values on Science's RadarView.
- Allow SpaceObjects (including SpaceShips and PlayerShips) to be visually hidden and detectable only by their radar signal. Invisible targetable/selectable objects can be selected only if they have a significant signal value (default > 0.3 sum). Invisible objects still leave ghost dot trails on short-range radars.
- Add scripting functions to support signal detail retrieval and setting, signal threshold detection, and visibility.

visualization and hidden ship demo


k9lego:
I would love for an option on the main menu to restrict a ship to only radar/visual/both. Any chance the data on the screen when selected could be moved to the side bar for the Science? Imo it's good like that for GM but adds to the clutter on the science screen since the data values are so precise. This is amazingly cool btw!

oznogon:
@k9lego I'm not going to make any guarantees or recommendations on stability without further review or more testing. I don't see crashes, but I haven't seen other confirmed crash bugs on my systems either. I haven't even really tested the most recent set of changes yet on a proper networked client/server setup to ensure I've got replication sufficiently covered, and probably won't before March 15.

I've rewritten parts of this in the last day to:

- allow signal details to be disabled on a per-ship and per-template basis via scripting `ShipTemplate():setSignalRadar()`, `SpaceShip():setSignalRadar()`, and `SpaceShip():hasSignalRadar()`. Making specific lens availability more granularly enabled is another level of complexity I haven't gotten around to implementing.
- move and rewrite the calculation of specific signal values to a Science sidebar tab
- change the lens behavior to disable the Visual lens if any of signal lenses are selected, and vice versa

I also rewrote the Science sidebar tabs to always be visible when a target is selected, instead of appearing only on a fully scanned target, a clunky workaround in lieu of rewriting it to be smarter about which tabs appear at different scan levels. This ensures that at least noisy, imprecise signal detail values are always available.

I'm considering removing the "Signal Details" toggle button altogether because it now serves no purpose, and instead just show the lenses. Signal details are always reachable in the sidebar and there's no longer any way to mix Visual and signal lenses.

signals

k9lego:
That is hilariously amazing! I love how the buttons have the frequency on the left side. I get it, until Daid implements it into the codebase it's in the "wildwest" in terms of stability. I'll do what I can to get crash logs if it does happen, but I plan on doing my own testing as well! We'll have a backup plan if crashes occur, but it'll be thoroughly tested on the 15th I can assure you. I would really like to see this implemented into the game correctly, I feel as if it'd make it even more clear what these lines around Science mean to new-comers.

src/screens/crew6/scienceScreen.cpp
```diff
@@ -203,18 +219,123 @@ ScienceScreen::ScienceScreen(GuiContainer* owner, ECrewPosition crew_position)
...
+
+void ScienceScreen::setVisualDetailsToggle(bool value)
+{
+    if (my_spaceship->has_signal_radar && science_radar->getSignalDetails())
+    {
+        science_radar->setVisualObjects(value)->setSignalElectrical(false)->setSignalGravity(false)->setSignalBiological(false);
+        probe_radar->setVisualObjects(value)->setSignalElectrical(false)->setSignalGravity(false)->setSignalBiological(false);
+        signal_details_electrical_button->setValue(false);
+        signal_details_gravity_button->setValue(false);
+        signal_details_biological_button->setValue(false);
```
k9lego: I would suggest that you change
`        science_radar->setVisualObjects(value)->setSignalElectrical(false)->setSignalGravity(false)->setSignalBiological(false);
        probe_radar->setVisualObjects(value)->setSignalElectrical(false)->setSignalGravity(false)->setSignalBiological(false);
        signal_details_electrical_button->setValue(false);
        signal_details_gravity_button->setValue(false);
        signal_details_biological_button->setValue(false);`
To
`        science_radar->setVisualObjects(value)->setSignalElectrical(!value)->setSignalGravity(!value)->setSignalBiological(!value);
        probe_radar->setVisualObjects(value)->setSignalElectrical(!value)->setSignalGravity(!value)->setSignalBiological(!value);
        signal_details_electrical_button->setValue(!value);
        signal_details_gravity_button->setValue(!value);
        signal_details_biological_button->setValue(!value);`
This makes it so when you deselect V it will just select all 3 other buttons. While it still is possible to deselect all 3 sensors and see nothing, it provides a quick way to switch between seeing Visual and seeing all available sensor data.
https://gyazo.com/a68b277506a56e83d9baba71a31c0044

oznogon:
Looks like the radar lock feature does break this, so I'll close this PR and revisit later.

PR #768, opened 2020-03-06
Prevent access violation in nvoglv32.dll

NinjaSpectre:
**Observed Problem:**
Failing to have a corresponding `glDisableClientState()` for each `glEnableClientState()` somehow corrupts `glDeleteBuffers()`.  Even though the buffer appears to be cleared successfully, any openGL function call made after `glDeleteBuffers()` could result in an access violation in `nvoglv32.dll`.
This access violation may only affect Maxwell graphics processors.

**Implemented Solution:**
Immediately after `glDrawArrays()` use `glDisableClientState()` for each `glEnableClientState()`.


oznogon:
When rebased onto master and tested on a run-through of Basic with unaffected macOS 10.15 and Windows 10 systems, this did not introduce any regressions. However, testing was complicated by #769.

PR #771, opened 2020-03-09
Prepare system names for translation

aBlueShadow:
Choose to add context, as there are different ones for jump/warp.
Added the context to the other systems for consistency.

### DIFF
--- a/src/shipTemplate.cpp
+++ b/src/shipTemplate.cpp
@@ -1,3 +1,4 @@
+#include <i18n.h>
 #include "shipTemplate.h"
 #include "spaceObjects/spaceObject.h"
 #include "mesh.h"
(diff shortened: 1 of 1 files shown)


daid:
I was afraid of this patch. Didn't expect it so soon :)

As this one is one of those where it gets ugly. By directly translating in this function there are problems. For example, if the server and client are not running the same language, hacking will no longer work. And the script function `getBeamSystemTargetName` will suddenly return different results. And hardware.ini configuration will break. Just to name a few things (as there are more things)

As I was foreseeing this. I just added a `trMark()` function, which marks a string for translation, but does not do the actual translation yet. And then at the right locations where these strings are used we could wrap them in a call to `tr()`.
Another possible solution would be to have a `getLocalizedSystemName` function next to the getSystemName function. And use that when the string is requested for displaying goals instead of for other uses.


Maybe I should list 2 goals in the translation wiki:

1. Do not break scenario scripts when language is changed
2. Do not break features if clients use a different language then the server (it is accepted that comms message will not match your selected language)

aBlueShadow:
O wow, I mean even I have forseen problems (remember our talk about button texts this morning), but did not expect issues on that part, as I thought all systems are determined by their Esystem value.
As for getBeamSystemTargetName, IIRC i just added that function as a "nice to have" option with getBeamSystemTarget being the more important one.

daid:
Well, many parts of the code where pretty quickly written and organically grown instead of well thought out. So best not to make assumptions on how things work and check them.

daid:
https://github.com/daid/EmptyEpsilon/commit/5d9f41288efd99348fccfd3416ddcf6990cb0ea3

PR #773, opened 2020-03-09
FactionInfo localization & translation

lacombet:
Translation and localization of the "FactionInfo.lua" script in French, using Daid prerogatives :

https://github.com/daid/EmptyEpsilon/wiki/Translation&Localization


daid:
Ok, few things.

First, simple one. We don't need the `.mo` file. (This is a binary version of the `.po`, faster to load, and smaller on disk. But that isn't really useful for us)

Secondly, by translating the `setName()`, we break every single scenario. This is one of those instances where these strings have a dual role, for showing to the player and for internal use. But I also understand that `Human Navy` needs some form of translation. So I've added a function `setLocaleName` to the faction info. So there are different names for displaying and script/internal usage. This means the FactionInfo lines can be changed:
```
human = FactionInfo():setName("Human Navy")
```
To:
```
human = FactionInfo():setName("Human Navy"):setLocaleName(_("Human Navy"))
```

lacombet:
I removed the .mo file and fixed the names.

daid:
Awesome!

PR #774, opened 2020-03-10
Ship templates corvette localization & translation  (FR)

lacombet:
Localization and translation of "shipTemplates_Corvette.lua"

Not tested yet as I can't build.

@daid : Should I translate ship classes like I did ?


itugul:
Of course, I understand that everything is already ready, and I must decide to make adjustments to each code associated with the displayed "string".
Here are all the files where my changes were:

image

Here are the key changes:
(The rest of the files are what is not processed in this place.)
image

Text Translation Functions:

lacombet:
Needs a fix :ShipTemplate should have a setLocaleName function.

daid:
Added the set functions, so you can do the localization.
However, not using the values anywhere in the code yet. As that part is a bit of a mess...

lacombet:
Thanks, It works.
As you said, names still appear in english, but the rest of it is ok.

If the class translations or ok with you, it's good to go ;)

scripts/shipTemplates_Corvette.lua
```diff
@@ -232,68 +232,68 @@ template:setBeam(5, 30, 300, 4000.0, 1.5, 20)
 --[[----------------------Freighters----------------------]]
 
 for cnt=1,5 do
-    template = ShipTemplate():setName("Personnel Freighter " .. cnt):setClass("Corvette", "Freighter"):setModel("transport_1_" .. cnt)
+    template = ShipTemplate():setName("Personnel Freighter " .. cnt):setLocaleName(_("Personnel Freighter ") .. cnt):setClass(_("Corvette"), _("Freighter")):setModel("transport_1_" .. cnt)
```
daid: Instead of `_("Personnel Freighter ") .. cnt`
It would be better to use `_("Personnel Freighter %d").format(cnt)`, as that gives the translator the decision on where to put the number.

lacombet: Oups, my mistake.

lacombet:
My bad !

I fixed it ;)

scripts/shipTemplates_Corvette.lua
lacombet: My bad !

I fixed it ;)

PR #783, opened 2020-03-14
Add short range radar range preference

oznogon:
Add `server_config_short_range_radar_range` to PreferencesManager
and modify short-range radar view and screens accordingly.

Values are bounded to a range of 500 and 20000 (0.5U to 20U).

The default value remains 5000.

Nebula fog of war behavior now relies on this setting.

- If set to greater than 5000, helm, weapons, tactical, and
  single pilot use NebulaFogOfWar as science does.
- Friendly shared radar ranges on relay scale with the value.
- If close enough to a nebula, science can see through it up to
  the value.

This value also affects probe radar range.

This does _not_ add the setting to the server creation screen.
It is set only in options.ini (for now).

Examples with short-range radar set to 20U (20000) and
long-range radar set to 125U (125000):

Screen Shot 2020-03-14 at 1 53 55 PM
Screen Shot 2020-03-14 at 1 53 02 PM
Screen Shot 2020-03-14 at 1 52 57 PM

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -31,7 +31,8 @@ GameGlobalInfo::GameGlobalInfo()
     scanning_complexity = SC_Normal;
     hacking_difficulty = 2;
     hacking_games = HG_All;
-    long_range_radar_range = 30000;
+    short_range_radar_range = 5000.0;
+    long_range_radar_range = 30000.0;
     use_beam_shield_frequencies = true;
     use_system_damage = true;
     allow_main_screen_tactical_radar = true;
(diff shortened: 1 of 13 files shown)


Xansta:
Can the short range sensors be set larger than the long range sensors? Can they be set to the same value?

daid:
I'm actually thinking about making these ship level configuration parameters (long and short range radar) instead of server configuration.

oznogon:
OK @daid, thanks. I'll close this PR out and leave the branch up for reference

daid:
Added the patch today, that allows to set this per ship and per ship template.

PR #790, opened 2020-03-16
i18n for helms

aBlueShadow:
Prepare translation for helms screen.

### DIFF
--- a/src/screenComponents/jumpControls.cpp
+++ b/src/screenComponents/jumpControls.cpp
@@ -1,3 +1,4 @@
+#include <i18n.h>
 #include "playerInfo.h"
 #include "spaceObjects/playerSpaceship.h"
 #include "jumpControls.h"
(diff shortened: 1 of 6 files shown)


src/screens/crew6/helmsScreen.cpp
```diff
@@ -126,7 +127,7 @@ void HelmsScreen::onDraw(sf::RenderTarget& window)
         energy_display->setValue(string(int(my_spaceship->energy_level)));
         heading_display->setValue(string(my_spaceship->getHeading(), 1));
         float velocity = sf::length(my_spaceship->getVelocity()) / 1000 * 60;
-        velocity_display->setValue(string(velocity, 1) + DISTANCE_UNIT_1K + "/min");
+        velocity_display->setValue(string(velocity, 1) + DISTANCE_UNIT_1K + tr("unit", "/min"));
```
daid:
```
string(velocity, 1) + DISTANCE_UNIT_1K + tr("unit", "/min")
```
Would be better as:
```
tr("{value} {unit}/min").format({{"value", string(velocity, 1)}, {"unit": DISTANCE_UNIT_1K}})
```
I think. As that would add flexibility to the translator, and a bit more context where this is used.

src/screenComponents/jumpControls.cpp
```diff
@@ -18,10 +19,10 @@ GuiJumpControls::GuiJumpControls(GuiContainer* owner, string id)
     charge_bar->setPosition(0, -50, ABottomLeft)->setSize(50, GuiElement::GuiSizeMax);
     charge_bar->hide();
     
-    label = new GuiKeyValueDisplay(this, id + "_LABEL", 0.5, "Distance", "10.0");
+    label = new GuiKeyValueDisplay(this, id + "_LABEL", 0.5, tr("Distance"), "10.0");
```
daid: Might be good to have a context with this distance? Not all distances might translate the same in all languages?

aBlueShadow: You are right. I actually added the context in another part of the file, but missed that one.

src/screens/crew6/helmsScreen.cpp
aBlueShadow: Done. (I guess the **:** was a typo)

src/screenComponents/jumpControls.cpp
aBlueShadow: Done.

PR #793, opened 2020-03-18
Configurable engine sounds

oznogon:
Allow impulse engine sounds to be configured in ship templates, and impulse engine volume levels and playback to be configurable in the options screen.

- Add engine volume and playback options to the Options screen, mirroring the music playback options.
- Fix a bug where impulse engine sounds would continue to play after the ship was destroyed.
- Add `engine_enabled` and `engine_volume` preferences to the preferences file/`options.ini`.
- Add code to play engine sounds from crew stations as well as the main screen, if `engine_enabled` is set to `1` ("Enabled").
- Add the `setImpulseSoundFile()` function to ShipTemplate and ShipTemplateBasedObject, taking a string of a filename in a relative path to the `resources` directory.
- Add a new engine sound for the MP52 Hornet (`resources/sfx/engine_fighter.wav`).

This does not change any default behaviors, but does move the default impulse engine sound from `resources/engine.wav` to `resources/sfx/engine.wav`.

Options screen:

Screen Shot 2020-03-18 at 2 55 17 AM


src/menus/optionsMenu.cpp
```diff
@@ -69,25 +69,38 @@ OptionsMenu::OptionsMenu()
     sound_volume_slider->setPosition(50, top, ATopLeft)->setSize(300, 50);
 
     // Override overlay label.
-    sound_volume_overlay_label = new GuiLabel(sound_volume_slider, "SOUND_VOLUME_SLIDER_LABEL", "Sound Volume: " + string(int(soundManager->getMasterSoundVolume())) + "%", 30);
+    sound_volume_overlay_label = new GuiLabel(sound_volume_slider, "SOUND_VOLUME_SLIDER_LABEL", tr("Sound Volume: ") + string(int(soundManager->getMasterSoundVolume())) + "%", 30);
```
daid:
```
tr("Sound Volume: {volume}%").format({{"volume", string(...)}})
```
(Also at the other locations)

src/screens/crewStationScreen.cpp
```diff
@@ -144,6 +147,34 @@ void CrewStationScreen::update(float delta)
                 break;
             }
         }
+
+#ifndef __ANDROID__
+        // If engine sounds are enabled and we have impulse, loop the engine sound.
```
daid: Not so happy that this is duplicated code with the main screen.

src/menus/optionsMenu.cpp
oznogon: Updated

src/screens/crewStationScreen.cpp
oznogon: Me neither, but the music code is already duplicated between ScreenMainScreen and CrewStationScreen.

I'm having trouble figuring out the best place to put both of them — in a new background channel managed by SF's SoundManager, or in a ScreenComponent in EE, or by rewriting the Main Screen to inherit from CrewStationScreen (or creating a more generic screen class that both ScreenMainScreen and CrewStationScreen)

daid: I think a new class that just manages the engine sounds would be a good way. And then just create this new class from both different screens.

oznogon: Moved impulse sound to its own class under ScreenComponents in 85cd2ab.

src/screens/crewStationScreen.cpp
```diff
@@ -65,6 +66,9 @@ CrewStationScreen::CrewStationScreen()
         });
     }
 #endif
+
+    // Initialize and play the impulse engine sound.
+    impulse_sound = std::unique_ptr<ImpulseSound>( new ImpulseSound(PreferencesManager::get("impulse_sound_enabled") == "1") );
```
daid, line 71: This is missing the default of "2". So when you open a crew station or the main screen per default on a fresh install, and never went to the options screen the engine sounds will be off. But if you first went to the options screen, the sound will be mainscreen only.

src/screens/mainScreen.cpp
```diff
@@ -63,6 +64,9 @@ ScreenMainScreen::ScreenMainScreen()
         });
     }
 
+    // Initialize and play the impulse engine sound.
+    impulse_sound = std::unique_ptr<ImpulseSound>( new ImpulseSound(PreferencesManager::get("impulse_sound_enabled") != "0") );
```
daid, line 68: This is missing the default of "2". So when you open a crew station or the main screen per default on a fresh install, and never went to the options screen the engine sounds will be off. But if you first went to the options screen, the sound will be mainscreen only.

PR #796, opened 2020-03-18
weaponname localization

aBlueShadow:
This adds the function getLocaleMissileWeaponName, similar to getLocaleSystemName

### DIFF
--- a/src/scienceDatabase.cpp
+++ b/src/scienceDatabase.cpp
@@ -157,7 +157,7 @@ void fillDefaultDatabaseData()
         {
             if (ship_template->weapon_storage[n] > 0)
             {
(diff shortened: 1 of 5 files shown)


src/scienceDatabase.cpp
```diff
@@ -157,7 +157,7 @@ void fillDefaultDatabaseData()
         {
             if (ship_template->weapon_storage[n] > 0)
             {
-                entry->addKeyValue("Storage " + getMissileWeaponName(EMissileWeapons(n)), string(ship_template->weapon_storage[n]));
+                entry->addKeyValue("Storage " + getLocaleMissileWeaponName(EMissileWeapons(n)), string(ship_template->weapon_storage[n]));
```
daid: tr("Storage {weapon}").format({...})

aBlueShadow: fixed

PR #819, opened 2020-03-24
Move music playing code on crew and main screens to its own class

oznogon:


src/screens/mainScreen.h
```diff
@@ -22,12 +23,15 @@ class ScreenMainScreen : public GuiCanvas, public Updatable
     GuiRadarView* long_range_radar;
     bool first_person;
     GuiCommsOverlay* onscreen_comms;
+    Music* music;
```
daid: instead of using raw pointer `Music*` use a `std::unique_ptr<Music>` as right now, you leak memory when the main screen is destroyed, as the Music object is never destroyed. Instead of manually destroying it, you can just use `std::unique_ptr` to delete it for you.

You might notice I use raw pointers for the GUI elements, because those are owned by their parent GUI element, so they will get destroyed when the GUI is destroyed.

src/screens/crewStationScreen.h
```diff
@@ -34,6 +37,7 @@ class CrewStationScreen : public GuiCanvas, public Updatable
     GuiPanel* message_frame;
     GuiScrollText* message_text;
     GuiButton* message_close_button;
+    Music* music;
```
daid: Same as with the main screen.

src/screenComponents/music.h
```diff
@@ -0,0 +1,24 @@
...
+class Music
+{
+private:
+    P<ThreatLevelEstimate> threat_estimate;
+    string music_file;
+    bool music_enabled;
+    bool threat_set_enabled;
+public:
+    Music(bool enabled);
+
```
daid: Adding a destructor `~Music()` which stops the music ensures that the music stops playing as soon as this object is destroyed. Needing less "manual" calling of stop to stop the music when screens are exited.

src/screens/mainScreen.h
oznogon: coincidentally learned about shared_ptr and unique_ptr last night so well timed, thanks @daid

daid:
Specific reason you deleted this branch?

oznogon:
I need to figure out a different way to do this. I was modeling it after how I'd moved the impulse sounds to a class, but that has the same issues and `unique_ptr` wasn't cleaning up objects either (which led to #823), so I'll try to get impulse sounds done right first.

The branch is still there, fwiw: https://github.com/oznogon/EmptyEpsilon/tree/classify-music

PR #824, opened 2020-03-26
Revert "i18n for alert levels"

daid:
Reverts daid/EmptyEpsilon#812

### DIFF
--- a/src/spaceObjects/playerSpaceship.cpp
+++ b/src/spaceObjects/playerSpaceship.cpp
@@ -197,9 +197,9 @@ string alertLevelToString(EAlertLevel level)
     // Convert an EAlertLevel to a string.
     switch(level)
     {
-    case AL_RedAlert: return tr("alert","RED ALERT");
-    case AL_YellowAlert: return tr("alert","YELLOW ALERT");
-    case AL_Normal: return tr("alert","Normal");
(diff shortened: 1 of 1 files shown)


aBlueShadow:
What was the reason for this? I know it might look suspicious (similar to those functions that needed a locale counterpart), that's why I split it from the relay-PR. However IMO it is different in this case, as it only affects the button labels on the relay console, and alertLevelToString is not called from any other place. Or did I miss something?

daid:
It broke a script function on the player ship there is the `getAlertLevel` function which would have returned localized data. Missed that before I merged.

aBlueShadow:
Hm.. I looked into getAlertLevel back then and it looked like it only gets the EAlertLevel to me.
https://github.com/daid/EmptyEpsilon/blob/64523908a7d68132850dd4a5cd5eec7c3caf10b4/src/spaceObjects/playerSpaceship.h#L267

daid:
And that is converted by this function:
https://github.com/daid/EmptyEpsilon/blob/3646a96efb32b974e0e9c4bf5ea73911cb37cfa9/src/spaceObjects/playerSpaceship.hpp#L6

aBlueShadow:
oh I see. learning more every day

daid:
The whole lua script binding uses quite a bit of magic :)

PR #828, opened 2020-03-27
Add and restore icons from Wayback cache

oznogon:


daid:
Smart! Did not think of that

PR #830, opened 2020-03-27
Radar Lock for rotating radars to have ship pointing up.

JonathanILevi:
`RadarView` `view_position` now has a twin `view_rotation`.

`SpaceObject`'s render to radar functions now take a rotation argument as well.

Most of the draw functions in `RadarView` now  use the preexisting but updated `worldToScreen` function rather than calculating the position individually.  It may be appropriate to have `radar_screen_center` and `scale` be member variables since they are used all over the place so they do not need to be recalculated all the time.

The options for whether to radar lock are in the options menu.  I also added a save button to the options menu so that players are not annoyed that they have to reset the options every-time.

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -239,16 +239,8 @@ void GuiRadarView::drawSectorGrid(sf::RenderTarget& window)
         for(int sector_y = sector_y_min - 1; sector_y <= sector_y_max; sector_y++)
         {
             float y = sector_y * sector_size;
-            if (view_rotation==0)
-            {
-                sf::Vector2f pos = worldToScreen(sf::Vector2f(x,y));
-                drawText(window, sf::FloatRect(pos.x, pos.y, 30, 30), getSectorName(sf::Vector2f(sector_x * sector_size + sub_sector_size, sector_y * sector_size + sub_sector_size)), ATopLeft, 30, bold_font, color);
-            }
-            else
-            {
-                sf::Vector2f pos = worldToScreen(sf::Vector2f(x+(30/scale),y+(30/scale)));
-                drawText(window, sf::FloatRect(pos.x-10, pos.y-10, 20, 20), getSectorName(sf::Vector2f(sector_x * sector_size + sub_sector_size, sector_y * sector_size + sub_sector_size)), ACenter, 30, bold_font, color);
-             }
+            sf::Vector2f pos = worldToScreen(sf::Vector2f(x+(30/scale),y+(30/scale)));
+            drawText(window, sf::FloatRect(pos.x-10, pos.y-10, 20, 20), getSectorName(sf::Vector2f(sector_x * sector_size + sub_sector_size, sector_y * sector_size + sub_sector_size)), ACenter, 30, bold_font, color);
         }
     }
     sf::VertexArray lines_x(sf::Lines, 2 * (sector_x_max - sector_x_min + 1));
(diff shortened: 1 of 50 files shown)


src/screenComponents/radarView.cpp
```diff
@@ -218,41 +223,50 @@ void GuiRadarView::drawNoneFriendlyBlockedAreas(sf::RenderTarget& window)
...
     for(int sector_x = sector_x_min - 1; sector_x <= sector_x_max; sector_x++)
     {
-        float x = radar_screen_center.x + ((sector_x * sector_size) - view_position.x) * scale;
+        float x = sector_x * sector_size;
         for(int sector_y = sector_y_min - 1; sector_y <= sector_y_max; sector_y++)
         {
-            float y = radar_screen_center.y + ((sector_y * sector_size) - view_position.y) * scale;
-            drawText(window, sf::FloatRect(x, y, 30, 30), getSectorName(sf::Vector2f(sector_x * sector_size + sub_sector_size, sector_y * sector_size + sub_sector_size)), ATopLeft, 30, bold_font, color);
+            float y = sector_y * sector_size;
+            if (view_rotation==0)
```
daid: This part looks a bit odd, why the special case for view_rotation zero?

amir-arad:
I've implemented this feature in my branch,  but sadly it's beyond PR (too many un-merged features to the radar component, dating back years)

technically, my approach was more light-weight: I rotated the textures before and after drawing on them,  and sometimes specific vectors as well (waypoints, callsigns, things I wanted to draw "straight" but on a rotated position) , using these helper methods.

```c++
void GuiRadarView::rotateTexture(sf::RenderTexture& texture, float rotation){
    sf::View view(rect);
    view.setViewport(texture.getView().getViewport());
    view.setRotation(rotation);
    texture.setView(view);
}

void GuiRadarView::orient(){
    if (target_spaceship && auto_orient){
        rotateTexture(background_texture, target_spaceship->getRotation() + 90);
        rotateTexture(forground_texture, target_spaceship->getRotation() + 90);
    }
}

sf::Vector2f GuiRadarView::orientVector(sf::Vector2f input){
    if (target_spaceship && auto_orient){
        return getCenterPosition() + sf::rotateVector(input - getCenterPosition(), -target_spaceship->getRotation() -90);
    } else {
        return input;
    }
}

void GuiRadarView::deOrient(){
    if (target_spaceship && auto_orient){
        rotateTexture(background_texture, 0);
        rotateTexture(forground_texture, 0);
    }
}
```

IMO it's a much simpler approach. feel free to "steal" from my branch if you agree.

daid:
Much simpler, but most likely generates more rendering artifacts.

JonathanILevi:
@daid I believe you were looking at where the sector grid labels are drawn.

The original location of the labels is in the extreme corner touching the edge.  I did not what to effect the look of the unmodified radar but the labels don't really work mashed that far in the corner when they are rotated.

If you don't like it.  The sector labels could always be gapped from the corner, rotated with the view (confusing and hard to read), or if you have another idea?

JonathanILevi:
About @amir-arad 's solution.  I believe daid is right about distortion, which would be worse on low res screens (which many bridges, on a budget, use).

The solution with the fewest modified lines of code is not necessarily the best, nor cleanest, solution.

I believe rotating images is also slower than sending an additional argument to a function.

Rotating the draw surface is clever but hacky.

Sending a rotation argument to the draw functions is the direct, simple, mathematically-correct solution.  We tell the function where to draw, it makes sense to also tell it what rotation to draw at.  It also leaves the control of how to deal with the rotation to the object rather than either having `if` conditions in the radar code, or sending an additional argument anyway.

Daid may disagree, but that is my thought.

daid:
About the labels, where do you draw them now? Got a screenshot?

Speed wise, drawing to a texture and then drawing that is slower. As drawing to a texture and then directly using it usually invokes a pipeline stall. There is already one of these in the radar rendering when it masks things for nebulae. And the impact of that is notable on really low end hardware.

I think the solution in this pull request is clean and proper. Thing I might have done differently is moving all the parameters into a structure, so there are not that many arguments to pass. And it's a bit more flexible to add/remove parameters in the future. But that's not really something that needs to change before I would merge it.
I was really close to merging it, but I was wondering about the sector labels.

JonathanILevi:
Here is a picture:
[https://cdn.discordapp.com/attachments/468130409129967631/694265035542757456/radarUpSectorLabel.png](url)

PR #834, opened 2020-03-28
[WIP] Controllable main camera angle

oznogon:
Allow changing the main screen camera's height relative to its ship and its pitch.

Neither the input method nor the behavior feels right. The more I poke this, the more it seems like the camera should be an object of its own with its own functions, instead of being sort of universal and controlled by modifying main-scope variables.

Keys used are W/S (pitch up/down) and E/D (raise/lower), which don't align with the cinematic view controls of Up/Down and R/F respectively; those conflict with the existing main screen controls.


oznogon:
Just noticed that a change to move camera_fov to a higher scope and give it a setter. Can remove it later if it's a conflict, I need to step away for a while

oznogon:
Need to revisit this in more detail, but will leave the branch up

PR #837, opened 2020-03-30
Add password protection for Spectate station

oznogon:
- Adds GM control code to the server creation screen
- Adds server_config_gm_control_code setting to options.ini for
  headless servers


daid:
Note that I would have preferred the check server side (as now, with a modified client you can bypass the password or even get the password) but a way of preventing this is better then none at all.

PR #841, opened 2020-03-31
Ship-like space stations

oznogon:
Convert space station templates to use the ShipTemplate class, making them work more like ships. Stations built with ShipTemplate are scannable and hackable

Add the Station class of SpaceShip and revise ShipTemplate, ShipTemplateBasedObject, and SpaceShip to render Station-class ships like existing space stations.

Does not remove SpaceStation, which would disrupt many scenarios.


src/spaceObjects/shipTemplateBasedObject.cpp
```diff
@@ -5,11 +5,19 @@ REGISTER_SCRIPT_SUBCLASS_NO_CREATE(ShipTemplateBasedObject, SpaceObject)
...
     REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, setShipTemplate);
     /// Set the class name of this object. Normally the class name is copied from the template name (Ex "Cruiser") but you can override it with this function.
     REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, setTypeName);
     REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, getTypeName);
+    REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, getClass);
+    REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, getSubClass);
+    /// Sets the physics on this object.
+    /// Enabled, static
+    /// Example: setPhysics(true, true)
+    REGISTER_SCRIPT_CLASS_FUNCTION(ShipTemplateBasedObject, setPhysics);
```
daid: I think there are a bunch of thing here now that are overcomplicated. And not actually working. I suggest not having a dynamic `setPhysics` on `shipTemplateBasedObject`. As this will be difficult to correctly send to the clients.

src/spaceObjects/spaceship.cpp
```diff
@@ -216,6 +217,17 @@ SpaceShip::~SpaceShip()
 
 void SpaceShip::applyTemplateValues()
 {
+    class_name = ship_template->getClass();
+
+    // If this ship is a station, default to the appropriate comms script and
+    // callsign, and increase its avoidance radius.
+    if (class_name == "Station" && game_server)
```
daid: Instead of looking at the class, I think looking at the TemplateType is the proper way to go.

src/spaceObjects/spaceship.cpp
```diff
@@ -257,10 +271,14 @@ void SpaceShip::applyTemplateValues()
...
         }
     }
+
     //shipTemplate->has_cloaking;
     for(int n=0; n<MW_Count; n++)
         weapon_storage[n] = weapon_storage_max[n] = ship_template->weapon_storage[n];
 
+    physics_enabled = ship_template->physics_enabled;
+    static_physics = ship_template->static_physics;
+    setCollisionPhysics(physics_enabled, static_physics);
```
daid: I have the feeling I see too many calls to setCollisionPhysics at the moment?

oznogon:
Will find another way to do it, thanks for the review @daid

PR #847, opened 2020-04-01
Ship-like space stations

oznogon:
Re-parent the SpaceStation class to SpaceShip, inheriting its traits and template options.

Revise radar rendering to scale all SpaceShip objects, to maintain SpaceStation rendering behavior.

Move Defense Platform template from Corvettes to Stations.

Stations are now scannable; to maintain the previous radar appearance and FoF behavior, all stations start at SS_SimpleScan to all factions.


Kilted-Klingon:
Ok, so I take it space stations will not be making the evolutionary leap to be playable by a crew?

PR #849, opened 2020-04-02
Add autoconnect_address option

oznogon:
Enable specifying an address for autoconnect clients using the `autoconnect_address` option, instead of relying on autodiscovery.

Example usage:

```
./EmptyEpsilon autoconnect=9 autoconnect_address=192.168.0.15
```

autoconnects this instance to the Single Pilot station of the first PlayerShip on the server at 192.168.0.15.


JonathanILevi:
@oznogon is it possible to have that autoconnect not select a station but stay in the ship/station selection screen?

oznogon:
@JonathanILevi I don't know, don't think so atm, the autoconnect screen is separate from the ship selection screen. an option to connect to a server but not a ship station would require either further changes to autoconnect behavior or a separate option

PR #850, opened 2020-04-02
Toggle player ship capabilities

oznogon:
Allow player ship capabilities to be toggled. Disabled capabilities don't appear on ship screens.

Adds toggles to ship templates, the GM tweaks menu (Player 2 tab), and scripting.

Toggleable capabilities are:

- Combat maneuvers `setCanCombatManeuver(bool)`
- Docking `setCanDock(bool)`
- Scanning `setCanScan(bool)`
- Self destruct sequence `setCanSelfDestruct(bool)`
- Probe launching `setCanLaunchProbe(bool)`
- Hacking `setCanHack(bool)`

Example:

Screen Shot 2020-04-01 at 8 50 40 PM

Screen Shot 2020-04-01 at 8 50 49 PM

### DIFF
--- a/src/screens/crew1/singlePilotScreen.cpp
+++ b/src/screens/crew1/singlePilotScreen.cpp
@@ -74,7 +74,8 @@ SinglePilotScreen::SinglePilotScreen(GuiContainer* owner)
     );
 
     // Ship stats and combat maneuver at bottom right corner of left panel.
-    (new GuiCombatManeuver(left_panel, "COMBAT_MANEUVER"))->setPosition(-20, -180, ABottomRight)->setSize(200, 150);
+    combat_maneuver = new GuiCombatManeuver(left_panel, "COMBAT_MANEUVER");
+    combat_maneuver->setPosition(-20, -180, ABottomRight)->setSize(200, 150)->setVisible(my_spaceship && my_spaceship->getCanCombatManeuver());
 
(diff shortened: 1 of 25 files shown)


Kilted-Klingon:
Hey Oz; wow; this is very interesting and provides some great options for the scavenger hunt/race scenario I'm cooking up.  Meaning, all the competing ship crews begin in the starting block, but only have the very basic capabilities, and they have to go and find/scavenge for increased abilities as well as the "keys" to finish the race.  They have to scavenge to enable offensive/defensive capabilities against marauders in the "arena".  And you just provided a whole bunch more options for that scenario!  Too cool!  Thanks very much for doing this!  8-)

src/screens/crew1/singlePilotScreen.cpp
```diff
@@ -120,6 +121,9 @@ void SinglePilotScreen::onDraw(sf::RenderTarget& window)
 {
     if (my_spaceship)
     {
+        // Toggle ship capabilities.
+        combat_maneuver->setVisible(my_spaceship->getCanCombatManeuver());
```
daid, line 125: Instead of each station controlling visibility of this control, the control itself could manage its own visibility. Just like the docking button controls it's own labels and so on.

This does require that the guiElement has some kind of update function called even if it is not visible. As it currently does not receive anything if it is hidden. It requires a bit more code changes "up front" but after that is less error prone, and less details in the code.

PR #855, opened 2020-04-04
Add custom functions to log screen

aBlueShadow:


Kilted-Klingon:
Cool, tks.  Not sure what I'd use this for just yet, but I'm sure I'll think of something.... 8-)

aBlueShadow:
@Kilted-Klingon actually, I was thinking that you might be one of the first people that might find it useful :-) I mean, there is space for 16 custom slots, more than any other station. So there has to be a usecase - maybe even in combination with log texts

Kilted-Klingon:
16 slots... that is quite a bit of real-estate!  Yes, I'm definitely going to have to put that to use... Thanks again!  8-)

aBlueShadow:
You're welcome!
Unfortunately, it is not more than 15 slots now, due to #861, but I think that should still be plenty :-).

PR #857, opened 2020-04-04
Add mission clock to Relay

oznogon:
Clock counts up in seconds, same as ship's log entries.

Screen Shot 2020-04-04 at 1 14 42 PM

### DIFF
--- a/src/screens/crew6/relayScreen.cpp
+++ b/src/screens/crew6/relayScreen.cpp
@@ -146,6 +146,10 @@ RelayScreen::RelayScreen(GuiContainer* owner, bool allow_comms)
     info_reputation = new GuiKeyValueDisplay(option_buttons, "INFO_REPUTATION", 0.7, tr("Reputation:"), "");
     info_reputation->setSize(GuiElement::GuiSizeMax, 40);
 
+    // Mission clock display.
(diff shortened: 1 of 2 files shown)


daid:
Woke up, and I realized, this patch does not work correctly on clients. As the `getElapsedTime` from the engine is local to their client. And not the same as the server.

Kilted-Klingon:
Oops.
Since you're re-examining this function, were you thinking of providing any sort of control over it via scenario script?  For example, for it to be a countdown timer?  Or perhaps for it to be reset, as in, hide/search phases of CtF have different times, 5 mins for hiding/placing flags, and then 60 min for searching phase, and it could be useful to allow the script to reset the clock.  
Just thoughts.  Having the base functionality just the way you made it is definitely helpful regardless.

daid:
I think a timer/counter in gameGlobalInfo would be a proper solution, it can be send to the clients, and we can nicely start it at the scenario start.

oznogon:
Ah... sorry, I'm still confused but I see a little better now. PlayerSpaceship::addToShipLog uses the engine's getElapsedTime, but it's only called from scripting, so it's only called from the server. ships_log is then replicated.

PR #858, opened 2020-04-05
Black Holes destroy objects by causing damage

aBlueShadow:
Before deleting a ship at the singularity, the game now tries to destroy it by dealing a huge amount of damage first. That way onDestruction() will be called when a sturdy ship gets swallowed by a black hole.


oznogon:
Are there indestructible objects? I can't remember if anything is actually wired to can_be_destroyed https://github.com/daid/EmptyEpsilon/blob/9b55e6dc48eb0100746c14e5d1458962b2f1aaf2/src/spaceObjects/shipTemplateBasedObject.cpp#L120

daid:
player ships can be set to be indestructible for LARP reasons. Kinda difficult to explain that the ship you are on is destroyed without wiring up the building with explosives.

Kilted-Klingon:
Thanks very much Blue, Oz, and Daid for taking care of this!  Much appreciated!  8-)

PR #862, opened 2020-04-06
More overrides

oznogon:


daid:
Surprising that you are adding these, very good. But I noticed the sheer amount of them and didn't even want to bother.

SeriousProton2 is compiled with ` -Werror=suggest-override` so that overrides are a must. Maybe we'll get there with EE as well?

PR #865, opened 2020-04-07
fix ship selection screen when player ships are destroyed

czenker:
**How To reproduce**

* create three player ships
* destroy the first one
* go back to the mission selection screen

**What happens**

The name of the second player ship is displayed twice and the name of the third one is not visible.

ship-selection


Kilted-Klingon:
Thanks @czenker for documenting and fixing this issue!  This had been an occasional annoyance that we just kind of dealt with, but it will be nice to have it fixed.  Tks again!

czenker:
You're welcome. The problem seemed new to me, but then again I have not played with multiple ships too often - let alone being killed. :wink:

PR #866, opened 2020-04-08
Use server-replicated mission clock, add to GM screen

oznogon:
Generate mission clock on the server as gameGlobalInfo->elapsed_time,
distribute to clients, and reset it across scenarios.

Convert addToShipLog to use gameGlobalInfo->elapsed_time instead of
engine->getElapsedTime().

Add the mission clock to the GM screen.

Clockwise from top: GM screen mission clock at top right corner info layout; two clients, each with different ship's log entries at different times synchronized to the mission clock.

Note that this changes the behavior of the ship's log on servers spanning multiple scenarios, which would continue ticking across multiple scenarios run in succession from the same server.

Screen Shot 2020-04-07 at 12 46 00 PM


Kilted-Klingon:
Very nice!  Tks for your efforts on this!

daid:
Thanks for jumping on this so fast!

PR #870, opened 2020-04-08
Fallback to old width for Single Pilot

aBlueShadow:
This falls back to the old viewport tiling when at the single Pilot Station. So the viewport for that station will be as wide as it was prior to a2ed0b2ea36a1993624b1f197bc270eb79272238


daid:
Another thing I was thinking about was the background that the stations are adding. If that was handled a bit differently, we could have an small "overlap" between the right side main screen and the left side station screen. Giving a bit of the glass cockpit effect without taking over the whole screen.

I also see my patch just as a start of more options. Not as the death of the glass cockpit, just as a more generic solution.

PR #871, opened 2020-04-08
Unitize Mission clock string

aBlueShadow:
Add a ":" to the mission clock string on the GM screen, the same way it is in relay. Otherwise, there would be two, almost identical, strings in the po file.


oznogon:
the only problem is that makes this the only info block on that info_layout AutoLayout of the GM screen with a colon. not sure if that matters. alternatively, could add a colon to the text in the loop that populates when an object is selected.

alternatively, could probably just move the colon out of the `tr()` function and concatenate it: `tr("Mission Clock") + ":"`

aBlueShadow:
I see. I was wondering if there was a specific reason for the different format.
I think in that case, a concatenation is probably the best solution. That way, the colon can't be translated, but I don't think that would be really needed.

daid:
> alternatively, could probably just move the colon out of the tr() function in Relay and concatenate it: tr("Mission Clock") + ":"

Is fine.

PR #880, opened 2020-04-11
Add orbital movement to spaceObject

oznogon:
Move the Planet orbit mechanics to SpaceObject as
SpaceObject::setOrbit(target, time) and SpaceObject::unsetOrbit()
functions. Also expose these functions to scripting.

Add applyOrbit(delta) function, to be added to the update(delta)
functions of objects. Adds it to ShipTemplateBasedObject,
could be added to other SpaceObjects.

These aren't orbital physics — objects in an orbit are dragged
around by setPosition. Ships locked into an orbit can rotate
but can't move.

### DIFF
--- a/src/spaceObjects/planet.cpp
+++ b/src/spaceObjects/planet.cpp
@@ -103,7 +103,6 @@ REGISTER_SCRIPT_SUBCLASS(Planet, SpaceObject)
     REGISTER_SCRIPT_CLASS_FUNCTION(Planet, setPlanetCloudRadius);
     REGISTER_SCRIPT_CLASS_FUNCTION(Planet, setDistanceFromMovementPlane);
     REGISTER_SCRIPT_CLASS_FUNCTION(Planet, setAxialRotationTime);
-    REGISTER_SCRIPT_CLASS_FUNCTION(Planet, setOrbit);
 }
 
 REGISTER_MULTIPLAYER_CLASS(Planet, "Planet");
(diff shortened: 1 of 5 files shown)


Kilted-Klingon:
You're awesome Oz!!  Thanks so very much for doing this, truly.   It will simplify things greatly in our scenario scripts.  Before I had to set up tables to track and manage all the orbital items, but this change eliminates having to do all that.  And as I said in Discord, the real money-maker here is when we've got multiple levels in a single scenario script, this change greatly simplifies how to manage movement update functions when only 1 of several levels is active.  

Thanks very, very much!!

oznogon:
There are a lot of caveats with this approach and I won't be surprised or disappointed if it's not accepted, in favor of keeping Planet objects as a special class. A better solution IMO would interact more with game physics/use forces rather than setting position. But this is at least an option for scripting in specific situations.

Kilted-Klingon:
Understood, however, the setting position is a simple/rough approximation for what I was hoping for, and whereas I do agree some of the resultant arrangements could be on the absurd side (black holes orbiting asteroids), it does definitely work in my favor to simply the scripts and I really hope that daid will accept.  I was only asking for just the asteroids and nebula, but your change does also bring in the option of stations in orbit, as well as BHs in mutual orbits.  Truly, thanks again!

tdelc:
I did my own tests before. As I think (I didnt test your PR) there is one
really mistake with this method for the science officer : the relative
speed of objects will be wrong into his HUD.
However, this option is very interesting to a lot of scenarios.

Le dim. 12 avr. 2020 à 01:37, Kilted Klingon <notifications@github.com> a
écrit :

oznogon:
Yeah, and that's already true about planets if you make them targetable by science

Kilted-Klingon:
uh-oh..... ok, so does this mean no game-managed orbiting for asteroids and nebula?  8-(

PR #884, opened 2020-04-11
add translation markers for ships and small fixes

aBlueShadow:
- add translation markers for ship templates.
- now the ship class and ship name are both called dreadn**o**ught. I left the template file as shipTemplates_Dreadn**a**ught.lua for now, to not obfuscate the diff
- small fix that shows "no class" and "no subclass" correctly
- change "Cruiser" to "Karnack MK 2"  as that seems to be more consistent due to the description (the predecessor is even referred to as mark I) and both ships are extremely similar


aBlueShadow:
mixed up some branches. so I will close this PR and do a fresh one

PR #892, opened 2020-04-13
Refactor ship templates

oznogon:
Edit and refactor ship template files.

- Normalize file permissions and filenames
- Edit and expand some ship descriptions
- Standardize template formatting for improved readability
- Add some missing i18n markers
- Normalize subclasses; integrate unique subclass identifiers (ie. "sniper", "popper") into the ship's description
- Classify OLD template ships and move them into appropriate class files
- Give descriptions and locale names to some generic ship templates
- Move Ktlitan ships into their own file
- Remove generic transports ("Transport1x1", "Transport1x2", etc.) in favor of the specific equivalents ("Personnel Freighter 1", "Personnel Freighter 2", etc.) in the Corvettes file

Note that new locale names for ships won't appear by default; old non-locale `setName()`s were not modified in order to maintain scenario compatibility.


oznogon:
Due to the nature of some of these changes, this PR might be easier to review by commit than by files changed.

oznogon:
per @aBlueShadow fleshing out generic ships might conflict with people's use of them in scenarios, so I'll remove descriptions and locale names from them

if there's a strong opinion about removing the generic TransportNxN freighters, I can restore them as well, but they add a _lot_ of ships to the database that already exist as specific named corvettes

daid:
By renaming the files this pull request cannot be properly reviewed.

The old templates except for the Ktlitan should be deleted, or "hidden" from the game (maybe we should add some flag on them that they no longer show up in the database nor the GM screen selection) as the only reason to keep them is backwards compatibility. They generally have suitable replacements in the other templates (usually more then 1)

PR #895, opened 2020-04-14
Use ship locale names in the Science DB

oznogon:


aBlueShadow:
That will cause a problem, as the DB button will stop working. I made a similar PR in parallel, that should take that problem into account

PR #896, opened 2020-04-14
Show locale shipname

aBlueShadow:
This adds locale names to science DB GM screen and the science radar. It also sorts the DB entries by the locale name.

Probably #895 should be reverted before merging this one, as both basically tackle the same topic

### DIFF
--- a/src/menus/shipSelectionScreen.cpp
+++ b/src/menus/shipSelectionScreen.cpp
@@ -178,11 +178,12 @@ ShipSelectionScreen::ShipSelectionScreen()
     {
         GuiSelector* ship_template_selector = new GuiSelector(left_container, "CREATE_SHIP_SELECTOR", nullptr);
         // List only ships with templates designated for player use.
-        std::vector<string> template_names = ShipTemplate::getTemplateNameList(ShipTemplate::PlayerShip);
+        std::vector<std::pair<string, string> > template_names = ShipTemplate::getTemplateNameList(ShipTemplate::PlayerShip);
         std::sort(template_names.begin(), template_names.end());
 
-        for(string& template_name : template_names)
+        for(std::pair<string, string> template_pair : template_names)
         {
+            string& template_name = template_pair.second;
             P<ShipTemplate> ship_template = ShipTemplate::getTemplate(template_name);
             ship_template_selector->addEntry(template_name + " (" + ship_template->getClass() + ":" + ship_template->getSubClass() + ")", template_name);
         }
--- a/src/scienceDatabase.cpp
+++ b/src/scienceDatabase.cpp
@@ -91,13 +91,14 @@ void fillDefaultDatabaseData()
     P<ScienceDatabase> shipDatabase = new ScienceDatabase();
     shipDatabase->setName(tr("database", "Ships"));
 
-    std::vector<string> template_names = ShipTemplate::getTemplateNameList(ShipTemplate::Ship);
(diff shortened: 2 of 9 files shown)


aBlueShadow:
Or do you think I should adjust it first? I thought about doing a PR that reverts #895, but I did not want to appear to be disrespectful. So I would prefer if you or @oznogon would do that

oznogon:
just change https://github.com/daid/EmptyEpsilon/pull/895/commits/c5efee0150751b7d80772aaed61ec2dcefeedc56#diff-38973b868ebb8c4b94872b11f56e1c09R122 and note that https://github.com/daid/EmptyEpsilon/commit/387d53c9eafcbd74bef52dfe06448a22a805f052 already implemented https://github.com/daid/EmptyEpsilon/pull/896/files#diff-e9fb8ac3fb369af13b93fdd5d781d19cR79

daid:
Actually, for the database button on science, it might be good for database entries to have a "lookup key" next to the name. And then use that lookup key to find the entry, instead of the raw name.

Note that I have a giant patch to make the database send from the server to the clients. However, this patch currently does not work yet (crashes). A lot of the database code could use some... cleanup.

aBlueShadow:
@oznogon: okay, so I will do the first part as a separate PR then. As that would be faster than rebasing and re-submitting this PR here. (As your changes did not happen yet on my PR)

aBlueShadow:
> Actually, for the database button on science, it might be good for database entries to have a "lookup key" next to the name. And then use that lookup key to find the entry, instead of the raw name.
I am not sure if I understand. Isn't that basically what3124e8c7d34386db0f1186c0d60bc506647269f7 does? By introducing a localname property to the database, the untranslated entry will become the "lookup" key.

daid:
> > Actually, for the database button on science, it might be good for database entries to have a "lookup key" next to the name. And then use that lookup key to find the entry, instead of the raw name.
> 
> I am not sure if I understand. Isn't that basically what [3124e8c](https://github.com/daid/EmptyEpsilon/commit/3124e8c7d34386db0f1186c0d60bc506647269f7) does? By introducing a local_name property to the database, the untranslated entry will become the "lookup" key.

It sort of does, however, naming things properly is important. As it took me a while to figure out that that was why you did it that way.

aBlueShadow:
Hm.. generally speaking it was not that different from all those locale functions we introduced so far. Letting the original reference untouched by introducing a locale counterpart. Although this time the structure was quite different:

> As it took me a while to figure out that that was why you did it that way.
Then I am relieved that it is not just me :-)
As it also took me while to figure  out the structure of the science DB, and how I can introduce a locale counterpart in this case.
As my first attempt was was basically the same as c5efee0150751b7d80772aaed61ec2dcefeedc56 and it took me a few hours getting from there to here. So by "naming things properly" do you mean renaming some function or adding a comment in the code?

daid:
I opted to merge the "science database from scenario scripts" first, knowing it very likely would break this one. Sorry.

aBlueShadow:
I see. Keep in mind though, in case i did not get a rework ready in time, you should merge #898 before the next release, to prevent introducing a regression.
Cause as it seems, the database just got way more complex, so it could take a while.

aBlueShadow:
I forgot this was still open. Probably because the db button bug still exists. But this PR won't be the solution, and for maintain awareness of the issue, there is still #1080

PR #897, opened 2020-04-14
arrange elements in database view

czenker:
I rearranged text, KeyValuePairs, model and image on the database view so it looks nicer and uses the available space.

EmptyEpsilon_013
EmptyEpsilon_015
EmptyEpsilon_016
EmptyEpsilon_018
EmptyEpsilon_019
EmptyEpsilon_020
EmptyEpsilon_021

### DIFF
--- a/src/screenComponents/databaseView.cpp
+++ b/src/screenComponents/databaseView.cpp
@@ -33,7 +33,7 @@ DatabaseViewComponent::DatabaseViewComponent(GuiContainer* owner)
         }
         display(entry);
     });
-    item_list->setPosition(0, 0, ATopLeft)->setMargins(20, 20, 20, 130)->setSize(400, GuiElement::GuiSizeMax);
+    item_list->setPosition(0, 0, ATopLeft)->setMargins(20, 20, 20, 20)->setSize(navigation_width, GuiElement::GuiSizeMax);
     fillListBox();
 }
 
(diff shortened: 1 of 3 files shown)


daid:
Wont this conflict with tge station selector if you have multiple stations?

czenker:
I added some screenshots to show how it looks like. The key value pairs have not moved and the text avoids the area where the selector buttons would be. The only change in the text placement is that it moves _left_ of the buttons if there are no key-value pairs. But even then it does not overlap the buttons.

I also extended the navigation on the left (as seen on the ship screen). There was a bigger gap there before, but I am not aware of any element that would be placed there, right?

The other layout changes are for placing the image at a suitable place. I found it acceptable to have it partially covered by the semi-transparent station selector in some cases.

daid:
I see my mistake, I thought this was the science screen, which has an extra button to switch back to science radar.

src/screenComponents/databaseView.cpp
```diff
@@ -33,7 +33,7 @@ DatabaseViewComponent::DatabaseViewComponent(GuiContainer* owner)
         }
         display(entry);
     });
-    item_list->setPosition(0, 0, ATopLeft)->setMargins(20, 20, 20, 130)->setSize(400, GuiElement::GuiSizeMax);
```
daid, line 36: This extra room is for the switch back to science button.

src/screenComponents/databaseView.h
```diff
@@ -22,6 +22,8 @@ class DatabaseViewComponent : public GuiElement
     P<ScienceDatabase> selected_entry;
     GuiListbox* item_list;
     GuiElement* database_entry;
+
+    int navigation_width = 400;
```
daid, line 26: Can be a static constexpr

src/screenComponents/databaseView.cpp
czenker: You are right: The switch button on the science screen was the thing I forgot.

daid:
Also, any pull request containing pictures of kittens have a 93.4% improved chance of being accepted.

PR #898, opened 2020-04-15
revert locale name for science DB

aBlueShadow:
so the DB button will work correctly. The declaration of getLocaleName is left untouched, and will be complemented by #896


aBlueShadow:
Like the other PRs I just closed, this won't fix the DB button issue anymore

Issue #398, opened 2016-09-07
Dynamic Science Database


elptito:
Hi,

I want to do a dynamic database without a whole change of the code. I think about adding a isVisible() function for FactionInfo() and for ShipTemplate() and use it in the file "databaseView.cpp", just to check if this function return true of false before show it in the database. But I am not sure where to place it.
Do someones have some advices to do that ?

daid:
> I want to do a dynamic database without a whole change of the code.

Then you are out of luck. Problem is that this data is not send trough clients and thus changing it dynamically on the server does not change it on the clients.

elptito:
Ok, I will make some categories "do not touch !". For now I will be ok :)

PR #901, opened 2020-04-17
Remote Science Database

czenker:
You might know that I have a big thing for dynamcial generated pseudo-random missions. So a Science Database that can be manipulated within mission scripts is very high on my wishlist. Few use cases, where it can enhance a scenario:

* Check the registrar of a ship. Is Baddy McEvil *really* the owner of that gold transport as he claims?
* What is known about the nebula we are about to fly into?
* What is Station DS1 selling? And in which sector is it?
* Who is Jane Doe? How could we convince her to help us?
* Is the atmosphere on Primus toxic?
* or this...

EmptyEpsilon_022
(no kitten today, but a science joke)

I created a draft of this with my highly trained copy-and-paste C++ coding skills, so I am asking for someone - not saying @daid :smile: - to review the code and give a second opinion. Basically I changed the `ScienceDatabase` into a `MultiplayerObject`, let it replicate and recreate the tree structure via ids. But I have too little insight into the net code and possible performance bottle necks. And as always... I probably used strange types. :wink: 

## Things I'm planning before it gets merged

- [x] Sorting of the `Science Database` entries on the clients is random. ~~Should be trivial to sort them by their ascending id~~ Sorting is case-insensitive lexigraphic
- [x] Making sure there is no way where the navigation breaks and you can't navigate away from a page (for instance when a page is removed)
- [x] Extend the LUA Api to find/edit/delete database entries. tackles #398 at least partially and could allow to hide factions or ships in the database that the players should not know about yet.
- [x] Changelog entry

### DIFF
--- a/src/scienceDatabase.cpp
+++ b/src/scienceDatabase.cpp
@@ -14,20 +14,25 @@ REGISTER_SCRIPT_CLASS(ScienceDatabase)
     REGISTER_SCRIPT_CLASS_FUNCTION(ScienceDatabase, setImage);
 }
 
-
+unsigned int ScienceDatabase::next_id = 1;
 PVector<ScienceDatabase> ScienceDatabase::science_databases;
 
+REGISTER_MULTIPLAYER_CLASS(ScienceDatabase, "ScienceDatabase");
 ScienceDatabase::ScienceDatabase()
+: MultiplayerObject("ScienceDatabase")
 {
-    if (game_server) { LOG(ERROR) << "ScienceDatabase objects can not be created during a scenario right now."; destroy(); return; }
+    id = ScienceDatabase::next_id++;
+
+    registerMemberReplication(&id);
+    registerMemberReplication(&parent_id);
+    registerMemberReplication(&name);
+    registerMemberReplication(&model_data_name);
+    registerMemberReplication(&longDescription);
+    registerMemberReplication(&image);
+    registerMemberReplication(&keyValuePairs);
+
     science_databases.push_back(this);
     name = "???";
-}
-
-ScienceDatabase::ScienceDatabase(P<ScienceDatabase> parent, string name)
-{
-    this->parent = parent;
-    this->name = name;
 }
 
 ScienceDatabase::~ScienceDatabase()
@@ -36,8 +41,9 @@ ScienceDatabase::~ScienceDatabase()
 
 P<ScienceDatabase> ScienceDatabase::addEntry(string name)
 {
-    P<ScienceDatabase> e = new ScienceDatabase(this, name);
-    items.push_back(e);
+    P<ScienceDatabase> e = new ScienceDatabase();
+    e->parent_id = this->id;
+    e->setName(name);
     return e;
 }
 
@@ -49,6 +55,34 @@ void ScienceDatabase::addKeyValue(string key, string value)
 void ScienceDatabase::setLongDescription(string text)
 {
     longDescription = text;
+}
+
+
+void ScienceDatabase::setModelData(P<ModelData> model_data)
+{
+    this->model_data_name = model_data->getName();
+}
+
+void ScienceDatabase::setModelDataName(string model_data_name)
+{
+    this->model_data_name = model_data_name;
+}
+
+bool ScienceDatabase::hasModelData()
+{
+    return model_data_name != "";
+}
+
+P<ModelData> ScienceDatabase::getModelData()
+{
+    if (hasModelData())
+    {
+        return ModelData::getModel(model_data_name);
+    }
+    else
+    {
+        return nullptr;
+    }
 }
 
 static string directionLabel(float direction)
@@ -63,6 +97,14 @@ static string directionLabel(float direction)
     if (std::abs(sf::angleDifference(180.0f, direction)) <= 45)
         name = tr("database direction", "Rear");
     return name;
+}
+
+void flushDatabaseData() {
+    for (unsigned i=0; i<ScienceDatabase::science_databases.size(); i++)
+    {
+        ScienceDatabase::science_databases[i]->destroy();
+    }
+    ScienceDatabase::science_databases.resize(0);
 }
 
 void fillDefaultDatabaseData()
@@ -121,7 +163,7 @@ void fillDefaultDatabaseData()
         P<ShipTemplate> ship_template = ShipTemplate::getTemplate(template_name);
         P<ScienceDatabase> entry = class_database_entries[ship_template->getClass()]->addEntry(ship_template->getLocaleName());
         
-        entry->model_data = ship_template->model_data;
+        entry->setModelData(ship_template->model_data);
         entry->setImage(ship_template->radar_trace);
 
         entry->addKeyValue(tr("database", "Class"), ship_template->getClass());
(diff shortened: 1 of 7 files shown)


daid:
Good start. I have a bunch of comments once I get to a PC (on mobile atm). I tried this, but my version crashed the game.

src/scienceDatabase.cpp
```diff
@@ -14,30 +14,36 @@ REGISTER_SCRIPT_CLASS(ScienceDatabase)
     REGISTER_SCRIPT_CLASS_FUNCTION(ScienceDatabase, setImage);
 }
 
-
+unsigned int ScienceDatabase::next_id = 1;
```
daid, line 17: No need to keep an ID count in the science database, each multiplayer object already gets an unique ID:
https://github.com/daid/SeriousProton/blob/master/src/multiplayer.h#L231

src/scienceDatabase.cpp
```diff
@@ -65,6 +99,14 @@ static string directionLabel(float direction)
     return name;
 }
 
+void flushDatabaseData() {
```
daid, line 102: code style, opening brackets on next line (I'm going to be a tiny bit more strict on this then before)

src/scienceDatabase.cpp
```diff
@@ -14,30 +14,36 @@ REGISTER_SCRIPT_CLASS(ScienceDatabase)
...
+    registerMemberReplication(&model_data_name);
+    registerMemberReplication(&longDescription);
+    registerMemberReplication(&image);
+    registerMemberReplication(&keyValuePairs);
+
     science_databases.push_back(this);
     name = "???";
 }
 
-ScienceDatabase::ScienceDatabase(P<ScienceDatabase> parent, string name)
```
daid, line 27: Also remove it from the header file.

src/screenComponents/databaseView.cpp
```diff
@@ -14,53 +14,45 @@ DatabaseViewComponent::DatabaseViewComponent(GuiContainer* owner)
...
     }
-    return false;
+    return nullptr;
 }
 
-bool DatabaseViewComponent::findAndDisplayEntry(string name, P<ScienceDatabase> parent)
+bool DatabaseViewComponent::findAndDisplayEntry(string name)
 {
-    foreach(ScienceDatabase, sd, parent->items)
+    foreach(ScienceDatabase, sd, ScienceDatabase::science_databases)
```
daid, line 48: I know the foreach macro is used a lot in the code. But it can be replaced with range based for loops:
`for(auto sd : ScienceDatabase::science_databases)`
(Not critical, but if we are touching this line anyhow, we might as well fix that)

src/screenComponents/databaseView.cpp
```diff
@@ -69,35 +61,90 @@ void DatabaseViewComponent::fillListBox()
...
-    }else{
-        item_list->addEntry(tr("Back"), "");
-        foreach(ScienceDatabase, sd, selected_entry->items)
+        else if(parent_entry)
         {
-            item_list->addEntry(sd->getName(), sd->getName());
+            item_list->addEntry(tr("Back"), std::to_string(parent_entry->getParentId()));
+        }
+    }
+    if(children_idx.size() > 0)
```
daid, line 110: I think this code becomes a bit smaller and simpler if we assign `children_idx` to `siblings_idx` if `children_idx` is not empty, and then always use the 2nd path.

daid:
Overall, looks good and simple. Knowing the the crazy scenario builders, I would assume their next question will be a way to access and modify the "preloaded" database. And their first bug report will be that the displayed values do not update if the entry is updated from a script while you are viewing it.

src/screenComponents/databaseView.cpp
czenker: Wow. This could become my new favorite syntax. :D

daid: "auto" and "range based for loops" are really powerful features of C++11. When I initially wrote EE, I didn't know about range based for loops, and that you could write your own custom iterators.

src/scienceDatabase.h
```diff
@@ -10,39 +10,61 @@ class ScienceDatabaseKeyValue
...
 
     void addKeyValue(string key, string value);
     void setLongDescription(string text);
 
     void setName(string name) { this->name = name; }
+    string getNormalizedName() {
+        if(normalized_name == "")
+        {
+            normalized_name = name;
+            transform(normalized_name.begin(), normalized_name.end(), normalized_name.begin(), ::tolower);
```
czenker, line 49: Don't solve the same problem twice:

    string("foo").lower()

czenker:
I took care of the Lua API today. Here is the example i tested it with.

`````lua
require("utils.lua")

function assert(condition, msg)
	if not condition then error(msg, 1) end
end

function init()
	local ship_db = queryScienceDatabase("Ships")
	assert(ship_db ~= nil and ship_db.typeName == "ScienceDatabase", "queryScienceDatabase() finds Ships database")

	local ship_db = queryScienceDatabase("shIPS")
	assert(ship_db ~= nil and ship_db.typeName == "ScienceDatabase", "queryScienceDatabase() is case insensitive")

	local foo = queryScienceDatabase("Ships", "Broken", "too")
	assert(foo == nil, "queryScienceDatabase() returns nil when no entry exists")

	local tug_db = ship_db:getEntryByName("No Class"):getEntryByName("Tug")
	assert(tug_db ~= nil and tug_db.typeName == "ScienceDatabase", "getEntryByName() traverses the db")

	-- Example: remove a db entry
	tug_db:destroy()
	local tug_db = ship_db:getEntryByName("No Class"):getEntryByName("Tug")
	assert(tug_db == nil, "getEntryByName() no longer finds destroyed records")
	
	-- Example: traverse the database
	local function printDb(entry, indent)
		indent = indent or 0
		local pad = ""
		for _=1,indent do pad = pad .. " " end

		print(pad .. "* " .. entry:getName())
		if (entry:hasEntries()) then
			for _, child in pairs(entry:getEntries()) do
				printDb(child, indent + 2)
			end
		end
	end
	for _, entry in pairs(getScienceDatabases()) do
		printDb(entry)
	end

	-- Example: full dump a science entry
	local emp_db = queryScienceDatabase("Weapons", "EMP")
	print(emp_db:getName() .. "\n\n" .. emp_db:getLongDescription())
	for key, value in pairs(emp_db:getKeyValues()) do
		print(string.format("%20s: %5s", key, value))
	end

	-- Example: juggling key values

	local test_db = ScienceDatabase():setName("Test")
	test_db:setKeyValue("foo", "bar")
	test_db:setKeyValue("baz", 42)
	test_db:setKeyValue("baz", "1337")
	test_db:setKeyValue("bar", "secret info")
	test_db:removeKey("bar")
	assert(test_db:getKeyValue("foo") == "bar")
	assert(test_db:getKeyValue("bar") == "")

	-- Test replication
	local rev = 1
	addGMFunction("Randomize DB", function()
		local _ = print -- stupid upvalue error :D
		rev = rev + 1
		test_db:setName("Test rev." .. rev)
		test_db:setLongDescription("This is the long description of revision " .. rev)
		test_db:setKeyValue("baz", math.random(1, 999))
	end)

	-- Spawn a player Atlantis.
	player = PlayerSpaceship():setFaction("Human Navy"):setTemplate("Atlantis")
end
`````

From my point of view this is fit to be merged. I played around with it and could not make the game crash - so I think this is a good sign. :smirk:
I guess there are some rough edges, but its an improvement from before already.

Requires https://github.com/daid/SeriousProton/pull/47
Closes https://github.com/daid/EmptyEpsilon/issues/398

Kilted-Klingon:
@czenker Thanks very much for adding this functionality!  No doubt, us "crazy scenario builders" will find ways to be creative with it!  ;]

PR #905, opened 2020-04-22
Refactor Waves Scenario: Unglue Enemies

Xansta:
I noticed that when playing Waves, the enemy ships often seem to get stuck out far away from the action and just sit there. I've added code to handle some of these situations:

- Attack and Defend Orders
  - If the target of the attack or defend order is invalid, change the order to roaming or to attack the current player
- Idle Orders
  - If the enemy ship is idle, change the order to roaming or to attack the current player
- Missile Ships Out of Missiles
  - If the enemy ship has no beam weapons and is out of missiles, wait a minute, then replenish the missile stocks

### DIFF
--- a/scripts/scenario_01_waves.lua
+++ b/scripts/scenario_01_waves.lua
@@ -279,6 +279,20 @@ function update(delta)
 		globalMessage("Wave cleared!");
 	end
 	if friendly_count == 0 then
+		globalMessage(string.format("Stopped on Wave %d", waveNumber));
 		victory("Ghosts");	--Victory for the Ghosts (== defeat for the players)
+	end
+	for pidx=1,32 do
+		local p = getPlayerShip(pidx)
+		if p ~= nil and p:isValid() then
+			if p:hasPlayerAtPosition("Relay") then
+				p.wave_info = "wave_info"
+				p:addCustomInfo("Relay",p.wave_info,string.format(_("Wave %d"), waveNumber))
+			end
+			if p:hasPlayerAtPosition("Operations") then
+				p.wave_info_ops = "wave_info_ops"
+				p:addCustomInfo("Operations",p.wave_info_ops,string.format(_("Wave %d"), waveNumber))
+			end
+		end
 	end
 end
--- a/scripts/scenario_01_waves.lua
+++ b/scripts/scenario_01_waves.lua
@@ -199,34 +199,7 @@ function update(delta)
 		if enemy:isValid() then
 			enemy_count = enemy_count + 1
 			order = enemy:getOrder()
-			if order == "Defend Target" or order == "Attack" then
-				order_target = enemy:getOrderTarget()
-				if order_target ~= nil then
-					if not order_target:isValid() then
-						if random(1,100) < 50 then
-							enemy:orderRoaming()
-						else
-							victim_station = nil
-							repeat
-								victim_station = friendlyList[math.random(1,#friendlyList)]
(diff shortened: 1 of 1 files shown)


scripts/scenario_01_waves.lua
```diff
@@ -191,9 +191,66 @@ function update(delta)
...
+	local order = nil
+	local order_target = nil
+	local ship_type = nil
 	for _, enemy in ipairs(enemyList) do
 		if enemy:isValid() then
 			enemy_count = enemy_count + 1
+			order = enemy:getOrder()
+			if order == "Defend Target" or order == "Attack" then
+				order_target = enemy:getOrderTarget()
+				if order_target ~= nil then
```
daid, line 203: I'm pretty sure the default AI behaviour drops from defend target or attack to something else if the target is gone. So this shouldn't be needed, or that code should be fixed instead.

scripts/scenario_01_waves.lua
```diff
@@ -191,9 +191,66 @@ function update(delta)
...
+							enemy:orderRoaming()
+						else
+							enemy:orderAttack(getPlayerShip(-1))
+						end
+					end
+				else
+					if random(1,100) < 50 then
+						enemy:orderRoaming()
+					else
+						endmy:orderAttack(getPlayerShip(-1))
```
daid, line 215: I'm not a fan of a direct attack command, on multi ship games, this makes it ignore the 2nd ship. I noticed this when playing borderline fever online, some enemy ships would just blindly fly by players to attack a far away station. Making for easy kills.

daid: Instead, you could order a move towards the players current position, or even a "orderRoamingAt" order to roam towards the players current position.

czenker: `orderDefendLocation` works very well for me when enemies should attack something without ignoring their sourrounding. `orderDefendTarget(player)` should work the same when the target is an enemy of the `CpuShip` and move.

As daid wrote, `orderAttack` is just a bit fanatic at times and makes it easy to kite enemies in a multiplayer scenario.

Xansta:
I switched out the attack player order with an order to fly towards a friendly station (friendly to the player, not the enemy). It looks like GitHub picked up my commit from my fork and added it here, but I'm not entirely sure.

As for the verification of the validity of the enemy target, I put that in on the assumption that the way the AI corrects the behavior when the target is invalid is to make the enemy go Idle (or something else). I did not see anywhere in the code that an enemy is ordered idle, but when I tested, there were enemies that were idle. So I added the target validation as well as the code to switch idle enemies to something more active. Let me know if I need to remove the target validation logic. I don't think it hurts, but it's not elegant.

Xansta:
I also added wave number information to final screen and to Relay

scripts/scenario_01_waves.lua
Xansta: Removed target validation code. Testing seems to show it changes to "Idle" if the target is bad

scripts/scenario_01_waves.lua
Xansta: Removed `orderAttack` in favor of `orderFlyTowards` one of the human navy friendly stations as an option when the enemy gets stuck in `orderIdle` apparently when the target of `orderDefend` goes bad

Xansta:
Switched out attack for fly towards human navy station

Xansta:
Looks like this pull request is going nowhere

PR #909, opened 2020-04-25
Fix two UBs

dodomorandi:
Honestly, I don't get why SeriousProton `Engine` destructor is sometimes called before `ImpulseSound` destructor, in any case this avoids a potential issue when the game shuts down.

I encountered the issue in `playerSpaceship.cpp` playing the tutorial.


daid:
Yeah, lifetime of objects are a bit strange in SeriousProton. Bad design decision.

PR #911, opened 2020-04-27
Docstrings

capt-t:
Replace several `//` by `///` to let the "docstrings" of Lua functions be included in `script_reference.html`.

Add a few strings.


Kilted-Klingon:
@capt-t - thanks very much for helping out and adding these changes!

PR #912, opened 2020-04-27
Compile script docs

capt-t:
Improve (docstring and) output of `compile_script_docs.py`.

In particular, list possible string values of several parameter types of the Lua functions.

### DIFF
--- a/compile_script_docs.py
+++ b/compile_script_docs.py
@@ -229,6 +229,16 @@ class DocumentationGenerator(object):
         stream.write('As well as check the already existing scenario and ship data files on how to get started.</p>')
         stream.write('</div>\n')
 
+        # TODO modify the script and read the constants from the cpp files
+        stream.write('<div class="ui-widget ui-widget-content ui-corner-all">')
+        stream.write('<p>Some of the types in the parameters:</p>')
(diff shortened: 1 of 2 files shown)


Kilted-Klingon:
I've not looked at it, but this certainly does seem like it would be useful.  Thanks @capt-t !

daid:
I think it would be better to document the types at the conversion function, for example at this function:
https://github.com/daid/EmptyEpsilon/blob/master/src/playerInfo.cpp#L263
And then extract that data as well. This will also allow the document generator to link to these types instead of them being hardcoded.

capt-t:
I agree that this would be the best solution. But so far this was not possible: To which object or function should the comment lines be added to? The C types are not included in the reference.

At first I had considered to mention the values for `EAlertLevel` at `commandSetAlertLevel`, but I wanted to explain each type only once. The current place is not the perfect solution, but better than the non-documentation before. It was annoying to find the correct capitalization of "RED ALERT" vs. "Normal".

After `compile_script_docs.py` has been extended, the comments are easy to move.

capt-t:
Mentioned the possible improvement in the python file.

Fixed the examples of `ScriptStorage`.

Issue #910, opened 2020-04-26
fail to build with mingw and LTO


StarryWisdom:
bcbf8fbb23709a774bd3e3f66861a51b2f507590 seems to revert the fix in fc4cff5330011a43ced1b71dc22793aed46321f6

The bug there was a failure to build via mingw in WSL, with a large amount of errors (100s/1000s) along the lines of /usr/bin/i686-w64-mingw32-ld: beamEffect.cpp.obj (symbol from plugin):(.gnu.linkonce.t._ZN10BeamEffectD0Ev[__ZTv0_n12_N10BeamEffectD0Ev]+0x0): multiple definition of `non-virtual thunk to BeamEffect::~BeamEffect()'; threatLevelEstimate.cpp.obj (symbol from plugin):(.gnu.linkonce.t._ZN10BeamEffectD0Ev[__ZThn16_N10BeamEffectD0Ev]+0x0): first defined here

I still dont really understand what the original bug was/is, to summarise what I found when fc4c was created.
Its related to LTO being enabled (without LTO it compiles correctly).
The current debian install created by WSL exhibits this bug.
The version daid was using at that time was working.
Daid was able to reproduce the fail to build, but I cant recall if that was via WSL or a linux install.
I dont understand why if it is invalid C++ it is (but would love it explained)
I also do not have a minimal reproduction of the error, it seems to be related to diagonal inheritance though (adding/removing vtable entries masked or showed it), if it is a compiler bug it would be nice to know its in an upstream bug tracker somewhere

oznogon:
Are you using the compiler flag `-flto-partition=1to1`? (If not, can you try with that?)

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto-873

> Specify the partitioning algorithm used by the link-time optimizer. The value is either ‘1to1’ to specify a partitioning mirroring the original source files or ‘balanced’ to specify partitioning into equally sized chunks (whenever possible) or ‘max’ to create new partition for every symbol where possible.

JonathanILevi:
#907 is that pull which appears to have reverted the fix.

I am not able to test; but it looks like WSL is needing the deconstructors to be explicitly defined.  #907 changed the virtual deconstructors to use `= default` (which may or may not be a problem for WSL).  #907 also removed the empty deconstructors, because, you know, they are not necessary... but by the looks of it, when building on WSL, they are.

@StarryWisdom Can you test this?

oznogon:
[...]

PR #931, opened 2020-05-08
Fix for failure to build with newer mingw

StarryWisdom:
see issue #910


daid:
Going to be slightly annoying, sorry :)

Can you add a comment in these destructors why they exist? Else someone (including myself) might decide to clean them up again breaking the build.

PR #937, opened 2020-05-13
Gm create menu

StarryWisdom:
I intended to create 2 topic branches but messed up on my end - I suspect this isnt a issue though

when spawning >32 player ships they stop spawning with a P object assertion in debug mode, in release mode it runs fine

create menu defualts to CPU ships, as such if the new selector isnt changed or new buttons arent pressed nothing has changed, screenshot below

extra tweak menu items below as well
image
image

### DIFF
--- a/src/missileWeaponData.cpp
+++ b/src/missileWeaponData.cpp
@@ -22,6 +22,21 @@ const MissileWeaponData& MissileWeaponData::getDataFor(EMissileWeapons type)
     return missile_data[type];
 }
 
+string getMissileSizeString(EMissileSizes size)
+{
+    switch (size)
+    {
+        case MS_Small:
+            return "small";
+        case MS_Medium:
+            return "medium";
+        case MS_Large:
+            return "large";
+        default:
+            return "unknown size:" + size;
+    }
+}
+
 #ifndef _MSC_VER
 // MFC: GCC does proper external template instantiation, VC++ doesn't.
 #include "missileWeaponData.hpp"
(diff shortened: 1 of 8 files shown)


amir-arad:
hi @StarryWisdom,
notice [max_player_ships](https://github.com/daid/EmptyEpsilon/blob/18f0d758d4a71c2e431e1aa01a85b0d850c62e63/src/gameGlobalInfo.h#L51) constant - the 33rd ship you will create might not fail-fast in production for performance reasons, but it will not serve as a player ship for any practical use-case. for starters you can't connect to it from a remote client.

When dealing with game objects it's important to QA your work with a client connecting to a server. forgetting about network state/command is a very common pitfall.

Xansta:
I like the option of being able to create a player ship from the GM screen. I notice in your example that the player ship names are not ordered alphabetically by ship type name like they are when using the spawn ship. Not a big deal, but with a list that long, it might be easier to find the ship type you want if you can scan alphabetically

daid:
I just added the ability to disable player ship creation from the scenario:
https://github.com/daid/EmptyEpsilon/commit/18f0d758d4a71c2e431e1aa01a85b0d850c62e63
Maybe when that flag is set we should disallow the GM ship creation as well?

StarryWisdom:
I think I've tested everything remotely, though if I have missed anything let me know for future.

The intention was to have them alpabetically sorted, that is fixed above.

allowNewPlayerShips works as expected now, it also will work if its changed mid script (I cant figure out if it being possible to change is a bug or a feature, but it is supported now regardless)

the on_new_player_ship is not triggering when ships are created, but I think that could be argued to be a bug in the callback not triggering for lua created playerobjects, which will implicity solve that

Xansta:
@daid it makes sense to honor the scenario author's wish to prevent player ship creation in the GM screen. Someone determined to break the scenario author's rules can always write their own script

PR #939, opened 2020-05-14
Health max

tdelc:
It allows the GM to limit the maximal health of each system to simulate major failure. Automatic or damcom repair cannot repair it, only GM can fix it with tweak. Engineer can see the maximal health with a black progress bar for the broken system. 

Useful for scenario with GM and LARP.

Change in the actual game : There is no change until GM want to use this feature.

Capture d'écran de 2020-05-14 02-50-59
Capture d'écran de 2020-05-14 02-51-08

### DIFF
--- a/src/repairCrew.cpp
+++ b/src/repairCrew.cpp
@@ -179,13 +179,11 @@ void RepairCrew::update(float delta)
             position = sf::Vector2f(pos);
 
             ESystem system = ship->ship_template->getSystemAtRoom(pos);
-            if (system != SYS_None && ship->systems[system].health < ship->systems[system].health_max)
+            if (system != SYS_None)
             {
                 ship->systems[system].health += repair_per_second * delta;
                 if (ship->systems[system].health > 1.0)
                     ship->systems[system].health = 1.0;
-                if (ship->systems[system].health > ship->systems[system].health_max)
-                    ship->systems[system].health = ship->systems[system].health_max;
                 ship->systems[system].hacked_level -= repair_per_second * delta;
                 if (ship->systems[system].hacked_level < 0.0)
                     ship->systems[system].hacked_level = 0.0;
--- a/src/repairCrew.cpp
+++ b/src/repairCrew.cpp
@@ -179,11 +179,13 @@ void RepairCrew::update(float delta)
             position = sf::Vector2f(pos);
 
             ESystem system = ship->ship_template->getSystemAtRoom(pos);
-            if (system != SYS_None)
+            if (system != SYS_None && ship->systems[system].health < ship->systems[system].health_max)
             {
                 ship->systems[system].health += repair_per_second * delta;
                 if (ship->systems[system].health > 1.0)
                     ship->systems[system].health = 1.0;
+                if (ship->systems[system].health > ship->systems[system].health_max)
+                    ship->systems[system].health = ship->systems[system].health_max;
                 ship->systems[system].hacked_level -= repair_per_second * delta;
                 if (ship->systems[system].hacked_level < 0.0)
                     ship->systems[system].hacked_level = 0.0;
--- a/src/screens/crew6/engineeringScreen.cpp
+++ b/src/screens/crew6/engineeringScreen.cpp
@@ -59,6 +59,8 @@ EngineeringScreen::EngineeringScreen(GuiContainer* owner, ECrewPosition crew_pos
         info.button->setSize(300, GuiElement::GuiSizeMax);
         info.damage_bar = new GuiProgressbar(info.layout, id + "_DAMAGE", 0.0, 1.0, 0.0);
         info.damage_bar->setSize(150, GuiElement::GuiSizeMax);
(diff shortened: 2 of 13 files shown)


src/repairCrew.cpp
```diff
@@ -179,11 +179,13 @@ void RepairCrew::update(float delta)
             position = sf::Vector2f(pos);
 
             ESystem system = ship->ship_template->getSystemAtRoom(pos);
-            if (system != SYS_None)
+            if (system != SYS_None && ship->systems[system].health < ship->systems[system].health_max)
```
daid, line 182: This check is flawed, as it won't unhack systems that are above max health now.

src/screens/crew6/engineeringScreen.cpp
```diff
@@ -59,6 +59,8 @@ EngineeringScreen::EngineeringScreen(GuiContainer* owner, ECrewPosition crew_pos
         info.button->setSize(300, GuiElement::GuiSizeMax);
         info.damage_bar = new GuiProgressbar(info.layout, id + "_DAMAGE", 0.0, 1.0, 0.0);
         info.damage_bar->setSize(150, GuiElement::GuiSizeMax);
+        info.health_max_bar = new GuiProgressbar(info.damage_bar, id + "_MAX", 1.0, 0.0, 0.0);
+        info.health_max_bar->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
```
daid, line 63: Not sure if I like this to show up for everyone all the time for a specific LARP/Scenario feature.

src/spaceObjects/spaceship.h
```diff
@@ -312,6 +313,8 @@ class SpaceShip : public ShipTemplateBasedObject
     void setEnergy(float amount) { if ( (amount > 0.0) && (amount <= max_energy_level)) { energy_level = amount; } }
     float getSystemHealth(ESystem system) { if (system >= SYS_COUNT) return 0.0; if (system <= SYS_None) return 0.0; return systems[system].health; }
     void setSystemHealth(ESystem system, float health) { if (system >= SYS_COUNT) return; if (system <= SYS_None) return; systems[system].health = std::min(1.0f, std::max(-1.0f, health)); }
+    float getSystemHealthMax(ESystem system) { if (system >= SYS_COUNT) return 0.0; if (system <= SYS_None) return 0.0; return systems[system].health_max; }
+    void setSystemHealthMax(ESystem system, float health_max) { if (system >= SYS_COUNT) return; if (system <= SYS_None) return; systems[system].health_max = std::min(1.0f, std::max(0.0f, health_max)); }
```
daid, line 317: As health is from -1.0 to 1.0, shouldn't max health be allowed negative as well?

src/screens/crew6/engineeringScreen.cpp
tdelc: I misspoke, I didn't think about only one LARP or a specific scenario, but rather the lot of scenarios when GM is active (in opposition with arcade scenarios without GM) or if the script takes into account this feature.

For example, in the scenario 56_carrierTurret, the health of the system is 'hacked' to simulate a failure, but the GM cannot tweak it and players only see the problem with the log :  
`tractorShip:setSystemHealth("warp",-.5):setSystemHealth("jumpdrive",-.5):orderStandGround():setCallSign(strikeShipNames[math.random(1,#strikeShipNames)])
			for pidx=1,8 do
				p = getPlayerShip(pidx)
				if p ~= nil and p:isValid() then
					p:addToShipLog(string.format("[%s] Help requested: Our engines are damaged beyond our ability to repair. We are located in %s",tractorShip:getCallSign(),tractorShip:getSectorName()),"#556b2f")
				end
			end`

Some examples of uses in script : 
- Max health can decrease once health is below 0 and playerShip has to be repair in a specific station
- Max health can be limited into some zones or near specific objects
- Max health can be broken during a WARP or JUMP to simulate specific failure 

Note that if you don't use it, players will not see anything.

src/repairCrew.cpp
tdelc: Do you think I need to put a specific part of code like this ?
`ship->systems[system].health = std::min(ship->systems[system].health_max , ship->systems[system].health)`

src/spaceObjects/spaceship.h
tdelc: I thought about 0.0 when the system is completely broken, but yes, a consistency with health is better. So I need to understand how to accept a review into a commit ^^ (and that change the tweak too)

tdelc: After some tests, it is hard to visualize negative health max with the black bar. Maybe I can replace black bar with an icon when the maximal health is below 1.0. What do you think ?

image

tdelc:
I made some corrections and changes to the PR. Do you prefer this version ? And more generally, are you interested by this feature ?

image

daid:
If I disagree with a feature, I try to be very upfront about it instead of reviewing the code :)

tdelc:
Haha, yes of course ! I will remember that :)

PR #944, opened 2020-05-16
Player create callback

StarryWisdom:
techincally I guess this could break scripts, however onNewPlayerShip hasnt been in any released version yet so I think the risk is minimal

there is an argument that the script already knows about ships its created, however I think consitancy between gm created, main menu created and script created is useful

### DIFF
--- a/scripts/scenario_10_empty.lua
+++ b/scripts/scenario_10_empty.lua
@@ -68,8 +68,8 @@ function update(delta)
 	--No victory condition
 end
 
-onNewPlayerShip(function(ship, index)
-    print(ship, index)
+onNewPlayerShip(function(ship)
+    print(ship)
     ship:destroy()
 end)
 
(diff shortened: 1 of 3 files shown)


daid:
The reason I did not do this is that it would also call this callback for ships created from the script.

StarryWisdom:
I think always calling onNewPlayerShip may be a feature rather than a bug - player ships lifecycle would be
created -> onNewPlayerShip gets called (gameplay) -> onDestroyed called -> destroyed

which I think is easier to understand than
created -> onNewPlayerShip gets called (unless created via scripiting) -> (gameplay) -> onDestroyed called -> destroyed

However if you dont feel that way I will make a pull request for fixing the (new) gm create menu creating players not calling onNewPlayerShip instead.

I also cant help shaking the feeling I'm missing a obvious example as to why one of these two methods is obviously better than the other. The difference between them seems small. However I personally lean slightly towards feeling this pull request is a cleaner soultion for lua code as it allows one to rely on onNewPlayerShip always being called.

daid:
And maybe more important even. Now the callback is called after the template is set. With this patch it is called too early

StarryWisdom:
Very good point and one that I had not seen and wouldnt see for quite a few months

PR #945, opened 2020-05-20
Add onGMClick

StarryWisdom:


src/GMScriptCallback.cpp
```diff
@@ -64,3 +64,36 @@ static int getGMSelection(lua_State* L)
...
+    ScriptSimpleCallback callback;
+
+    int idx = 1;
+    convert<ScriptSimpleCallback>::param(L,idx,callback);
+
+    gameGlobalInfo->on_gm_click=callback;
+
+    foreach (Updatable, u, updatableList)
+    {
+        P<GameMasterScreen> game_master_screen = u;
```
daid, line 79: Instead of this piece of slightly complex and convoluted code, it would be much clearer to hide/show the buttons from the gmscreen update function. That is the way most things work in EE (state based instead of event based)

Issue #943, opened 2020-05-16
Weapons console: mark as blocked if in warp


capt-t:
In the weapons etc. console, mark the beams and missiles as blocked if in warp.

PR #947, opened 2020-05-20
Introduce indicator for weapons blocked by warp

capt-t:
Add indicator for weapons (beams/missiles) blocked by warp to the weapons consoles.

This is a small and useful improvement for the weapons officer. So far it can be strange (in particluar in slow warp) why clicking on missiles etc. does have no effect. Additionally this is consistent with other indicators for blocked systems.

This will solve issue #943.

There are some open questions.

1. Is the code correct, in particular the comparison with current_warp?
2. Does it work for all three weapons consoles?
3. (Are weapons blocked during jumps? Do we need a similar indicator?)
4. Is the order (health, BLOCKED, energy, ...) the best?
5. Better or shorter text instead of "BLOCKED BY WARP"?
6. Which color? (It could be a more neutral yellow or blue, not orange or red.)
7. How do we get an icon?

Please test or comment.

Feel free to add modifications.

### DIFF
--- a/resources/gui/colors.ini
+++ b/resources/gui/colors.ini
@@ -43,6 +43,7 @@ label = #FFFFFF
 
 overlay_damaged = #FF0000
 overlay_jammed = #FF0000
+overlay_blocked_by_warp = #0000FF
 overlay_no_power = #FF0000
 overlay_low_energy = #FF8000
 overlay_low_power = #FF8000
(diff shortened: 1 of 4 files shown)


Xansta:
Instead of "Blocked by warp" can the missile fire buttons simply be greyed out (indicating they are disabled) while the ship is at warp? I'm not sure what a similar graphical representation of beams not firing might be. Perhaps a dimmed locked reticle/cursor on the locked target?

Just thinking out loud. I haven't run any tests.

daid:
Problem I have with this is that all other indicators are "problems", while blocked by warp drive is just normal operation.

So disabling the buttons would be a nicer solution, as it produces less visual distraction.

capt-t:
I see the point, but if the power is low, Engineering can usually fix this, if in warp, Helms can.

So a non-alert color like blue or gray will give a visual hint that (and why) weapons are currently not operational. This uses the existing concept and code of the indicators, which is good.

PR #948, opened 2020-05-20
Add SpaceShip:getDockingTarget()

capt-t:
The new Lua function `SpaceShip:getDockedWith` is useful and much more efficient than comparing every target via `isDocked`.

Please test the code.


capt-t:
Apart from the branch name, it is now called `getDockedWith` (instead of `getDockingTarget`).

Does it return `nil` if not docked?

daid:
`getDockedWith` returns something even when not docked, that doesn't sound proper...

capt-t:
If NULL returns `nil` in Lua, this should fix it.

PR #954, opened 2020-05-22
Scenario Unwanted Visitors

Xansta:
A number of objects are in orbit in this scenario. It can be a challenge for Helm to dock with stations in motion


daid:
One small request, can you place the ogg files in a subdirectory? The resources directory is getting quite cluttered.

Kilted-Klingon:
I like having stations in orbit and giving the additional challenge to the crew.  8-)

Xansta:
I assume that the `playSoundFile` function can find the file without a subdirectory path

daid:
You need the path relative to the "resources" directory, so your assumption is incorrect I think.

Xansta:
Adjusted the scenario script to look in the subdirectory

PR #956, opened 2020-05-23
Add more optional hotkeys for system power

capt-t:
Continuation of #951 and #955.

Engineers _can_ now have hotkeys to set the power of the currently selected system to 0, 30, 100, 150, 200 or 300 percent. (No default keys apart from Space for 100.)


daid:
You might want a level 250 as well, the heat generation grows exponential with power.

capt-t:
Added.

capt-t:
Added 50% as well, otherwise this would be the only missing multiple of 50.

PR #963, opened 2020-05-28
Proposal : advanced sectors system

tdelc:
Based on the work of Amir-Arad, I suggest a advanced sectors system to allow large dezoom into the relay and GM station.
All the game remains similar.
It is activated by options.ini and a new playerspaceship function : setFarRangeRadarRange. Once activated, it create two kid of sector : 
- actual sector (20 000 U) in blue
- area (64 sectors) in green
- region (64 areas) in blank
Between each level, an intermediate level is preset to help lisibility of range.

And all sector, area and region can be renamed by scripts.

Here is a gif of this system with a example of renamed area and region.

sector_system


Kilted-Klingon:
I am really liking this idea.  I do have a couple of maps that as GM, it would be very useful to have this extensive zoom out feature.  Thank you very much @tdelc for suggesting!!

PR #964, opened 2020-05-28
Rename mission to scenario clock (on screens)

capt-t:
Also rename "mission" to "scenario" clock on Relay and GM screen.

### DIFF
--- a/src/screens/crew6/relayScreen.cpp
+++ b/src/screens/crew6/relayScreen.cpp
@@ -147,8 +147,8 @@ RelayScreen::RelayScreen(GuiContainer* owner, bool allow_comms)
     info_reputation = new GuiKeyValueDisplay(option_buttons, "INFO_REPUTATION", 0.7, tr("Reputation:"), "");
     info_reputation->setSize(GuiElement::GuiSizeMax, 40);
 
-    // Mission clock display.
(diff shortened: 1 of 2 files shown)


StarryWisdom:
scenario clock as a name feels much more gamey in a way I personally dont like, I also kind of like mission clock due to the precidence with mission time from space missions (though admit there is a clash between "missions" and "the mission")

I want to suggest the left hand text on relay should be more script based, but I lack enough time and knowledge of the current code to be able to suggest an exact change

capt-t:
If someone has a better suggestion for the _overall_ scenario/mission?/voyage/quest/..., I am happy to change it.

I like the total time of the scenario there.
In addition to this, there can be scenario-specific count-downs or count-ups for _individual missions_ (at the same time or after each other) consisting of several steps or tasks or ...

daid:
Just name it "clock"? Or "localtime"?

StarryWisdom:
Personally its really the "scenario" part that I dont like.

"clock" and allow the script writer(s) to imply the clock is whatever feels approprate feels good
"ship time"/"ship clock" sounds good to me (sounds like station time used in some sci fi to me)
"localtime" feels slightly weird to me as its not a time (as it counts up from 0) but IMO sounds significantly better than "scenario time"

Sorry if this whole thing is just me being awkward, but Im probably one of the people least bothered by things staying in charater and if its somewhat jarring for me I suspect it will annoy others

PR #967, opened 2020-05-29
On gm click v2

StarryWisdom:
this is pull request #945 with the commit history cleaned up in the hopes that helps

### DIFF
--- a/src/screens/gm/objectCreationView.cpp
+++ b/src/screens/gm/objectCreationView.cpp
@@ -7,8 +7,8 @@
 #include "gameGlobalInfo.h"
 
 
-GuiObjectCreationView::GuiObjectCreationView(GuiContainer* owner, func_t enterCreateMode)
-: GuiOverlay(owner, "OBJECT_CREATE_SCREEN", sf::Color(0, 0, 0, 128)), enterCreateMode(enterCreateMode)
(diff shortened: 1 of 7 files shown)


src/screens/gm/objectCreationView.cpp
```diff
@@ -142,7 +142,10 @@ bool GuiObjectCreationView::onMouseDown(sf::Vector2f position)
 void GuiObjectCreationView::setCreateScript(string script)
 {
     create_script = script;
-    enterCreateMode();
+    gameGlobalInfo->on_gm_click = [this] (sf::Vector2f position)
```
daid, line 145: I see a potential problem here:
1) open gm screen
2) click create button
3) close gm screen
4) open new gm screen
5) click on screen, the on_gm_click references the old deleted gm screen in the lambda

StarryWisdom:
I think the above commit solves the issues

the above commit fixes the one for the objectCreationView

on_gm_click is reset at gameGlobalInfo.cpp:159 (added in the 2nd commit), that is where the gm buttons are cleared, I believe they share the same lifetime (and thus if gm button reset are in the correct location they both are). That being said I am currently have not looked directly at where/when the lua is loaded or unloaded so if that is wrong in someway I can go look in more depth.

PR #970, opened 2020-06-04
Scenarios: Trim trailing whitespace

capt-t:
Scenarios: Trim trailing whitespace in scenarios 00 to 10.

And fix EOL in 03.


capt-t:
And replace mixture of tabs and spaces by spaces.

These are whitespace-only improvement, which can easily be merged.

PR #974, opened 2020-06-10
onNewPlayership is called everytime one is created

### LINKED ISSUE #973: onNewPlayerShip not always called
currently there are 3 ways to spawn player ships
spawn player ship on the server menu
the create mode for player ships (added in about a fortnight back)
PlayerSpaceship() inside of a script

only the first calls onNewPlayerShip
I can see 2 ways of fixing it.

The first way is via editing setTemplate to call it
The upside of that is it is the only way I can think of that will catch PlayerSpaceship(). I think this is a plus as it allows script writers to use onNewPlayerShip to attach lua variables to every space ship and not be concerned they forgot any call site
The downside is it feels the wrong time to call onNewPlayerShip and as such I am concerned I may be missing a problem.
Also a playerSpaceship can have a template set 0 or 2+ times.
0 times currently causes a crash, but that is a bug that probably should be fixed
2+ times could be guarded against, though causes the soultion to be more ugly

The second way would be saying that onNewPlayerShip isnt called for lua code and fixing the object creation screen.

I like the first soultion more as a script writer, but fear there is going to be some unintended side effect.

I am happy to write either, but wanted to check which soultion was wanted and that there isnt an obvious third way I am missing

StarryWisdom:
Fixed #973

Either this or #975 should be applied - not both, see conversation in issue # 973

### DIFF
--- a/src/spaceObjects/playerSpaceship.cpp
+++ b/src/spaceObjects/playerSpaceship.cpp
@@ -731,6 +731,11 @@ void PlayerSpaceship::applyTemplateValues()
     can_combat_maneuver = ship_template->can_combat_maneuver;
     can_self_destruct = ship_template->can_self_destruct;
     can_launch_probe = ship_template->can_launch_probe;
+    if (!on_new_player_ship_called)
+    {
+        on_new_player_ship_called=true;
+	gameGlobalInfo->on_new_player_ship.call(P<PlayerSpaceship>(this));
+    }
 }
 
 void PlayerSpaceship::executeJump(float distance)
--- a/src/menus/shipSelectionScreen.cpp
+++ b/src/menus/shipSelectionScreen.cpp
@@ -200,10 +200,11 @@ ShipSelectionScreen::ShipSelectionScreen()
 
                 if (ship)
                 {
-                    ship->setTemplate(ship_template_selector->getSelectionValue());
+		    // set the position before the template so that onNewPlayerShip has as much data as possible
                     ship->setRotation(random(0, 360));
                     ship->target_rotation = ship->getRotation();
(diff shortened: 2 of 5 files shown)


src/spaceObjects/playerSpaceship.cpp
```diff
@@ -731,6 +731,11 @@ void PlayerSpaceship::applyTemplateValues()
     can_combat_maneuver = ship_template->can_combat_maneuver;
     can_self_destruct = ship_template->can_self_destruct;
     can_launch_probe = ship_template->can_launch_probe;
+    if (!on_new_player_ship_called)
```
daid, line 734: I think I likethis implementation the best. However, I would move this to the `onUpdate()` as that would ensure not only the template, but also other things like positions are applied.

StarryWisdom: With the current location of on_new_player_ship it means the lua script is called before the playerShip can be seen by anywhere else in the lua code.

This in turn allows one to attach lua variables in the onNewPlayershipShip and assume that every playerShip seen inside of the update loop has got them attached. This in turn allows one to remove a pattern of checking those varibles for nil before use in the lua scripts.

If instead the on_new_player_ship is moved to the update loop then the lua code will have seen the playerShip once before it is updated.

on_new_player_ship could be called in both the shipSelectionScreen and objectCreationView but scripted playerSpaceships would still suffer from that (and it would be somewhat inconsistant)

with this in mind do you want to
1) keep as is
2) move on_new_player_ship into update
3) move on_new_player_ship into update, manually call it in objectCreationView and shipSelectionScreen
4) other that I havent thought of

Personally I think 1 or 3 is best

PR #976, opened 2020-06-13
Scenario descriptions with provided ships

capt-t:
Mention the provided player ships in the scenario descriptions (0 to 8) or the effect of spawning them.

For _Quick Basic_, recommend the default, but allow more ships. Why prevent playing it with a support ship or 4 starfighters instead of a phobos?

If you have suggestions to improve the text, leave a comment.

Apart from the line in Quick Basic, this is a text-only commit and easy to merge.

### DIFF
--- a/scripts/scenario_07_quick_basic.lua
+++ b/scripts/scenario_07_quick_basic.lua
@@ -1,6 +1,8 @@
 -- Name: Quick Basic
 -- Description: Different version of the basic scenario. Which intended to play out quicker. There is only a single small station to defend.
 --- This scenario is designed to be ran on conventions. As you can run a 4 player crew in 20 minutes trough a game with minimal experience.
+---
+--- This scenario is designed for the provided player ship of type Phobos (or Atlantis on Advanced).
 -- Type: Convention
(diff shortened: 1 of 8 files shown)


scripts/scenario_07_quick_basic.lua
```diff
@@ -155,7 +158,8 @@ function init()
     player:setPosition(random(-2000, 2000), random(-2000, 2000)):setCallSign(ship_names[math.random(1,#ship_names)])
     player:setJumpDrive(true)
     player:setWarpDrive(false)
-    allowNewPlayerShips(false)
```
daid, line 158: Have multiple player ships breaks various small things in the scenario. As the primary ship is used for messages and a few other things.

capt-t:
You are right, additional ships in Quick Basic work only as support ships. (I might adapt the scenario at another time.) For now, I removed that change to finish this pull request quickly.

scripts/scenario_07_quick_basic.lua
capt-t: You are right, additional ships in Quick Basic work only as support ships. (I might adapt the scenario at another time.) For now, I removed that change to finish this pull request quickly.

PR #977, opened 2020-06-29
Feature: CpuShips restock missiles

Piglit:
* CpuShips can also restock their missiles docking on ships with the restockMissilesDocked flag. (Not only on stations)
* Missile Cruiser AI searches for missile resupply when out of ammo.


daid:
Overall, looks good. Only thing that I can think off that is missing is a way for the GM screen to order a retreat. But that's not essential.

PR #981, opened 2020-07-13
missileTubeControles: deactivate fire_button in warp

capt-t:
This replaces pull request #947 and solves issue #943.

Please test.

### DIFF
--- a/src/screenComponents/missileTubeControls.cpp
+++ b/src/screenComponents/missileTubeControls.cpp
@@ -141,6 +141,11 @@ void GuiMissileTubeControls::onDraw(sf::RenderTarget& window){
             rows[n].fire_button->setText(tr("missile","Firing"));
             rows[n].loading_bar->hide();
         }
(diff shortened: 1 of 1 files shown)


src/screenComponents/missileTubeControls.cpp
```diff
@@ -141,6 +141,11 @@ void GuiMissileTubeControls::onDraw(sf::RenderTarget& window){
             rows[n].fire_button->setText(tr("missile","Firing"));
             rows[n].loading_bar->hide();
         }
+
+        if (my_spaceship->current_warp > 0.0)
+        {
+            rows[n].fire_button->disable()->show();
```
daid, line 147: Shouldn't be calling "show()" here, that will make the button show up even if the tube is loading a missile during warp

capt-t:
Like that?

It might also be possible to move the code to `if (isLoaded)`.

PR #990, opened 2020-07-19
Scenario 07: refactoring

capt-t:
Mainly refactoring of `scenario_07`.

The scenario with improved code is the the same with a few hardly noticable changes:

Enhancements:

- Fix spelling.
- Give reasons for victory/defeat.

Modifications:

- Reorder sentences in objective.

Following Pull Requests:

- The central nebula is now centered around the center, not the ship (which is close to the center anyway).

Now it is easy to move the entire area to a different center or even create several of those. Moreover, is is also easier to reuse parts of the code.

### DIFF
--- a/scripts/scenario_07_quick_basic.lua
+++ b/scripts/scenario_07_quick_basic.lua
@@ -296,11 +296,12 @@ function startScenario()
         end
     end
 
-    friendlyList[1]:sendCommsMessage(
+    local station = friendlyList[1]
+    station:sendCommsMessage(
         player,
-        string.format([[%s, please inform your Captain and crew that you have a total of %d minutes for this mission.
+        string.format([[%s, your objective is to fend off the incomming Kraylor attack.
+Please inform your Captain and crew that you have a total of %d minutes for this mission.
 The mission started at the arrival of this message.
-Your objective is to fend off the incomming Kraylor attack.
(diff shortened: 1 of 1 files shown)


capt-t:
Wait with merging, there will be an updated version of these commits.

capt-t:
Should be ready for merge.

capt-t:
Should be ready for merge.

capt-t:
Should be ready for merge.

scripts/scenario_07_quick_basic.lua
```diff
@@ -347,14 +352,18 @@ function update(delta)
         -- Note that players can win even if they destroy the enemies by blowing themselves up.
         if enemy_count == 0 then
             victory("Human Navy")
-            setBanner("Mission: SUCCESS")
+            local text = string.format("Mission: SUCCESS (%d seconds left)", math.floor(gametimeleft))
```
daid, line 355: Does this text fit in the global message box? I remember the font being quite large and the box not being super large.

capt-t: Probably not, but this should not be a problem. Compare the detailed mission result in Borderline Fever.

Is there another possibility for a longer interesting message?

If it is a problem, we can remove this detail. The other changes (and the upcoming) are more relevant.

PR #992, opened 2020-07-19
comms: standardize formatting, improve comments

capt-t:
- remove ; at EOL
- indent anonymous functions
- improve a few comments
- rename two loop variables

### DIFF
--- a/scripts/comms_ship.lua
+++ b/scripts/comms_ship.lua
@@ -1,6 +1,14 @@
--- Name: Basic ship comms
--- Description: Simple ship comms that allows setting orders if friendly. Default script for any cpuShip.
+--- Basic ship comms.
+--
+-- Simple ship comms that allows setting orders if friendly.
+-- Default script for any `CpuShip`.
 
+-- NOTE this could be imported
+local MISSILE_TYPES = {"Homing", "Nuke", "Mine", "EMP", "HVLI"}
+
+--- Main menu of communcition.
+--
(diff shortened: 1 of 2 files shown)


scripts/comms_ship.lua
```diff
@@ -17,115 +25,134 @@ function mainMenu()
...
-        if shields == 1 then
-            msg = msg .. "Shield: " .. math.floor(comms_target:getShieldLevel(0) / comms_target:getShieldMax(0) * 100) .. "%\n"
-        elseif shields == 2 then
-            msg = msg .. "Front Shield: " .. math.floor(comms_target:getShieldLevel(0) / comms_target:getShieldMax(0) * 100) .. "%\n"
-            msg = msg .. "Rear Shield: " .. math.floor(comms_target:getShieldLevel(1) / comms_target:getShieldMax(1) * 100) .. "%\n"
-        else
-            for n=0,shields-1 do
-                msg = msg .. "Shield " .. n .. ": " .. math.floor(comms_target:getShieldLevel(n) / comms_target:getShieldMax(n) * 100) .. "%\n"
+    addCommsReply(
+        "Report status",
```
daid, line 67: I don't like this style, I find it much less readable.

capt-t:
(The style settings were consistent with all the other early scenarios. I do not call the style with `function` and `end` of anonymous functions in their own lines and then putting all the other arguments in their own lines as well not as unreadable.)

Nevertheless, this is a reduced version of the original commit.

PR #994, opened 2020-07-21
AI: When target of AI_Attack is destroyed, continue roaming.

Piglit:
prevously ships were idle when the attack target was no longer valid

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -442,7 +442,12 @@ void ShipAI::runOrders()
         }
         break;
     case AI_Attack:          //Attack [order_target] very specificly.
-        pathPlanner.clear();
+        if (owner->getOrderTarget())
+        {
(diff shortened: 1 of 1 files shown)


src/ai/ai.cpp
```diff
@@ -399,7 +399,12 @@ void ShipAI::runOrders()
         }
         break;
     case AI_Attack:          //Attack [order_target] very specificly.
-        pathPlanner.clear();
+        if (owner->getOrderTarget())
+		{
```
nallath, line 403:
```suggestion
	{
```

nallath (changes requested):
It's a bit nitpicky but other than that this seems good.

src/ai/ai.cpp
Piglit: fixed

PR #1013, opened 2020-08-08
Fix remaining whitespace in src/

capt-t:


src/gui/hotkeyConfig.cpp
```diff
@@ -288,17 +288,17 @@ std::vector<std::pair<string, string>> HotkeyConfig::listHotkeysByCategory(strin
...
                     if (key_name.second == item.hotkey.code){
-			string keyModifier = "";
-			if (item.hotkey.shift) {
-				keyModifier = "Shift+";
-			} else if (item.hotkey.control) {
-				keyModifier = "Ctrl+";
-			} else if (item.hotkey.alt){
-				keyModifier = "Alt+";
-			}
+            string keyModifier = "";
```
daid, line 291: This looks wrongly indented now

capt-t: Actually, it was wrongly indented before, I only converted the tabs to spaces.

But the additional commit now also fixes that.

PR #1019, opened 2020-08-13
Scenario - Construction Heist

Excelsior14:
This is my first published scenario. It is a single - ship mission. It should take an hour. There is randomized terrain generation algorithm so that it is replayable. The main mission goals remain constant. There is some difficulty variation.


nallath:
The scenario needs to be put in the scripts folder, otherwise it won't be detected.

I tried the scenario a bit, but I was a bit at a loss what to do. I think it might be a good idea to add a few messages explaining what the goal is.

PR #1049, opened 2020-09-02
Fix typo in the cmake

nallath:
Found a tiny typo


daid:
Best feature ever ;-)

Issue #1034, opened 2020-08-25
feature request: Options submenu to change keybindings


Piglit:
A in-game menu to change keybindings would be nice.
Not every player is able to edit options.ini.

Joystick buttons and axis should be considered, too.

oznogon:
#1050 is merged, adding a keybindings menu to the Options menu. It isn't designed to remap joystick controls at present, and I don't have a joystick to test against.

muerteFR:
@gcask  Maybe close it?

gcask:
Fair point, but I do not yield this much power :)

Issue #1047, opened 2020-09-02
gui_autolayout and destroy


StarryWisdom:
if an GuiElement is destroyed from within a GuiAutoLayout the next frame it is spaced as if it is present (but correctly is not rendered), the frame after that it is all rendered correctly.

A quick grep seems to imply this is not done in the current code base, but its possible I missed it.

gui2_autolayout.cpp:80 is responsible for advancing the location, it doesn't take destroyed status into account, gui2_container.cpp:21 is renders/manages the destroyed status, currently they are run in the reverse order.

gui2_autolayout.cpp:80 could be fixed via wrapping it in a check of the destroyed flag (after it being exposed or friended), this was tested locally and works.

the order of the calls could be changed probably, I have not looked at that, I have also not looked to see if other gui elements are effected.

I do not have a minimal test case to ensure fixes are correct, there is a possibility that this only happens when called from within onDraw.

czenker:
I think I have seen the behavior you describe with custom buttons. Maybe this is the test case you have been looking for:

````lua
-- Name: Test
-- Description: For testing

function init()
    local player = PlayerSpaceship():setFaction("Human Navy"):setTemplate("Atlantis")
    local stateOn, stateOff

    stateOn = function()
        player:removeCustom("state_off")
        player:addCustomButton("helms", "state_on", "On", stateOff)
    end
    stateOff = function()
        player:removeCustom("state_on")
        player:addCustomButton("helms", "state_off", "Off", stateOn)
    end

    stateOn()
end

function update()
end
````
Thanks for digging in so deep to find the root cause of the issue. Did you fix it in the mean time? At least on the master branch the issues with the flickering buttons still exists.

PR #1050, opened 2020-09-03
Add hotkey remapping options

oznogon:
Revisiting PR #437 to make progress toward solving #1034

- Adds a "Basic" hotkey category for unredefinable common keys, like F1 for help (TODO: Either enable rebinding or disable text entry fields in the rebinding screen)
- Adds functions to list hotkeys by category, set hotkeys
- Adds a hotkey configuration button to the interface options page, and hotkey rebinding interface that pulls from existing hotkey lists ~(TODO: Select keys using via key press, instead of typing the SFML keycode name into a text entry field)~ _done_
- Saves keybindings to the options file

Differences from #437:

- Updates layout to use more of the space (~TODO: Use autolayouts and margins, instead of `top + 50` patterns~)
- Adds pagination for long lists of keybinds that overflow the screen (~TODO: Improve paging~)

Still needed:

- Joystick mapping
- Duplicate keybinding detection/validation

EmptyEpsilon 9_6_2020 2_30_46 AM


oznogon:
Things I'm still thinking about, apart from the stuff in the description:

- Not sure about the "Basic" group. Several of those hotkeys aren't rebindable; if they're made to be rebindable, they should probably go in the existing General group, and if they aren't or shouldn't be rebindable, they shouldn't show up in a rebinding UI. (The original PR throws an error popup if you try to rebind anything in the basic group, and I haven't touched that behavior yet other than to rephrase the error message.) Either way, that separate group should probably get cut from this.
- the pagination functionality is hacky, it just moves the oversized panel containing hotkeys left and right. Might be better to have a scrollable view or hide/show pagination like in other parts of EE.
- the listAllHotkeysByCategory function repeats code from listHotkeysByCategory; the difference  provides a way to list hotkeys that don't have a binding already. That should probably be an optional argument on listHotkeysByCategory instead.
- if you put in more than one invalid keycode, then hit save, it flags the first one displayed and removes it, but none of the rest. Neither makes for particularly good UI, since the error doesn't show which function had the bad keybinding, and you get an error again the next time you save it if there's another different binding that's invalid.
- probably need to change the case on HotKey to Hotkey for consistency.
- since this is implemented as a separate menu, this might be invoked outside of the options screen, for instance via a hotkey of its own during gameplay.
- it probably won't take much to add joystick binding, but knowing what to put in the text fields would be even less intuitive. That might need to wait until after rebinding via input is figured out

src/gui/hotkeyConfig.cpp
```diff
@@ -343,3 +368,31 @@ void HotkeyConfigItem::load(string key_config)
         }
     }
 }
+
+bool HotkeyConfig::setHotKey(std::string work_cat, std::pair<string,string> key, string new_value)
+{
+
+    // test if new_value is part of the sfml_list
+    for (std::pair<string, sf::Keyboard::Key> sfml_key : sfml_key_names) {
```
nallath, line 376: Nitpicking: { should be on newline

(yeah yeah, I know, it's not super consistent now, but the more we try and keep it in order, the better it will get :))

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+    {
+        HotKeyMenu::pageHotkeys(-1);
+    });
+    next_page->setPosition(-50, 50, ATopRight)->setSize(50, 50)->disable();
+
+    // frame with keys
+    frame = new GuiPanel(this, "HELP_FRAME");
+    frame->setPosition(50, FRAME_TOP, ATopLeft)->setSize(600, FRAME_HEIGHT);
+
+    // TODO: This label is redundant with the category selector
```
nallath, line 44: Remove todo (it seems to already have been fixed?)

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+    // get chosen category
+    category_index = cat;
+    category = category_list[cat];
+
+    int top = FRAME_PADDING;
+    int left = FRAME_PADDING;
+
+    hotkey_list = hotkeys.listAllHotkeysByCategory(category);
+
+    // TODO: This label is redundant with the category selector
```
nallath, line 108: Fix todo

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+            break;
+        default:
+            break;
+    }
+}
+
+void HotKeyMenu::setCategory(int cat)
+{
+    // remove old entries
+    for (GuiTextEntry* text : text_entries){
```
nallath, line 86:
```suggestion
    for (GuiTextEntry* text : text_entries)
    {
```

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+        text->destroy();
+    }
+    text_entries.clear();
+    for (GuiLabel* label : label_entries){
```
nallath, line 90:
```suggestion
    for (GuiLabel* label : label_entries)
    {
```

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+    int top = FRAME_PADDING;
+    int left = FRAME_PADDING;
+
+    hotkey_list = hotkeys.listAllHotkeysByCategory(category);
+
+    // TODO: This label is redundant with the category selector
+    // category_label->setText(category);
+
+    // begin filling of Hotkey listing for category
+    for (std::pair<string,string> item : hotkey_list) {
```
nallath, line 112:
```suggestion
    for (std::pair<string,string> item : hotkey_list)
    {
```

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+        label_entries.push_back(new GuiLabel(frame, "NAME_LABEL", item.first.append(" = "), 30));
+        label_entries.back()->setAlignment(ACenterRight)->setPosition(left, top, ATopLeft)->setSize(400, 50);
+
+        text_entries.push_back(new GuiTextEntry(frame, "HotKey_Entry", item.second));
+        text_entries.back()->setTextSize(30)->setPosition(left + 400, top, ATopLeft)->setSize(100, 50);
+        top += 50;
+
+        // This keeps adding columns even if they don't fit on the screen
+        if (top > 650) {
```
nallath, line 122:
```suggestion
        if (top > 650) 
        {
```

I'm also more of a fan of calculating things (since now if the layout ever changes, the 650 might not be correct). Although in this case I'm not so sure how easy it is.

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+void HotKeyMenu::updateHotKeys()
+{
+    int i = 0;
+    std::string text = "";
+    bool hotkey_exists = false;
+
+    if (category == "basic") {
```
nallath, line 147:
```suggestion
    if (category == "basic") 
    {
```

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+        (new GuiButton(error_window, "ERROR_OK", "OK", [this]()
+        {
+            // Close this window
+            error_window->destroy();
+        }))->setPosition(0, -10, ABottomCenter)->setSize(200, 50);
+        return;
+    }
+
+    // read in all TextEntry values and update hotkeys
+    for (std::pair<string,string> item : hotkey_list) {
```
nallath, line 161:
```suggestion
    for (std::pair<string,string> item : hotkey_list) 
    {
```

src/menus/hotKeyMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+        text = text_entries[i]->getText();
+        hotkey_exists = hotkeys.setHotKey(category, item, text);
+
+        if (!hotkey_exists) {
```
nallath, line 166:
```suggestion
        if (!hotkey_exists) 
        {
```

src/menus/optionsMenu.cpp
```diff
@@ -14,8 +15,7 @@
 #include "gui/gui2_listbox.h"
 #include "gui/gui2_keyvaluedisplay.h"
 
-OptionsMenu::OptionsMenu()
-{
+OptionsMenu::OptionsMenu() {
```
nallath, line 18:
```suggestion
OptionsMenu::OptionsMenu() 
{
```

nallath:
It _might_ be a good idea to have the hotkeys in their own configuration file. That way it's easier to keep multiple configurations on a single machine. The advantage of that would be is that it's much easier to switch between configurations.

src/menus/hotKeyMenu.cpp
oznogon: Me neither, and I called it out in the PR description — the layout structure was lifted from the 2017 PR and needs updating. It should be easier to use GuiAutoLayouts for organizing the three main "rows" of title/category nav, hotkeys, and back/save nav.

The challenge (for me at least) is implementing an AutoLayout with [`LayoutVerticalColumns`](https://github.com/daid/EmptyEpsilon/blob/master/src/gui/gui2_autolayout.cpp#L55) for the hotkeys, it's not one I remember using before, so I'm not 100% sure if it'll do what I think it will in this case.

oznogon:
Failed CI run is unrelated:

```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libx/libx11/libx11-xcb-dev_1.6.4-3ubuntu0.2_amd64.deb  404  Not Found [IP: 52.177.174.250 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
##[error]Process completed with exit code 100.
```

src/menus/hotKeyMenu.cpp
oznogon: Banged my head on trying to make it an autolayout but haven't figured it out yet, I can't wrap my head around some of the behavior when an element becomes an autolayout and wind up with overlapping items, or can't select text fields or buttons because (I think) some container is on top blocking the click from passing through.

Worked around it for now by at least abstracting out the hardcoded values to reused constants set in the header.

oznogon:
Last commit implements a GuiTextField-derived HotkeyBinder class that converts onKey input when a field has focus to the equivalent hotkey text. ESC/Home/F1 clear the focused text field.

Disclaimer, as usual I don't really know what I'm doing.

oznogon:
List of rebinds is now a nest of autolayouts, and is now affected by #1047 when paging through categories.

daid:
This looks good, but there is now a conflict with the other pullrequest I just merged...

daid:
People are going to love this one I think 👯

Issue #1054, opened 2020-09-07
windows release zip and logs folder


StarryWisdom:
If there is a logs folder then empty epsilon will make log files for each time its started (rather than overriting them all).

Currently the windows zip is lacking such a folder.

If this is not intentional it would be nice if it was added.

oznogon:
Nope, that's for the JSON game-event logs used by the viewer in the repo's `/logs/index.html` (which also isn't bundled in the windows zip).

When that folder's present, you see this line in EmptyEpsilon.log (which stays in the same place and is still overwritten every run):

```
[INFO]  Opened game state log: logs/game_log_07-09-2020_10.41.22.txt
```

When it's not present, you see this line in the log:

```
[WARN]  Failed to open game state log file: logs/game_log_07-09-2020_10.44.36.txt
```

That game logging feature should probably be disabled by default and enabled in options, instead of enabled by default but not generating anything out of the box because the folder doesn't exist.

NinjaSpectre:
These game logs get rather large in size as well since it tries to document the actions of every object.  Wouldn't take overly long before some users became mystified as to where their hard drive space went if this was defaulted on.

daid:
It is intentional that this is not created by default.

PR #1056, opened 2020-09-09
Revise game state logging options and default behavior

oznogon:
- Game state logging is disabled by default. It's previously enabled by default, but doesn't write anything because the relatively-pathed `logs` directory is intentionally not created in the distribution zip (#1054).
- Add game state logging toggle to the interface options. (TODO: Seems like a better option to fit an in-game setting for this in the scenario selection/server settings screen, but that interface is crammed.)
- If a HOME environment variable is set (macOS, Linux, some Windows users), move the game state logging dir to `~/.emptyepsilon/`. This avoids an issue on macOS where the relative `logs` directory must be created inside the app bundle (`EmptyEpsilon.app/Contents/Resources/logs`) to capture game state logs. The behavior on most Windows systems (no HOME envvar) should not change.
- Create `~/.emptyepsilon` closer to game init, instead of waiting for the game to be closed. This allows game state logs to be captured during the first run of EE if logging is enabled, and also allows the Save button on the options screen to actually save `options.ini` on the first run of EE.
- Add a getter to PreferencesManager that doesn't set the preference value to an empty string value by default. This allows for key values to be checked in `options.ini` (`getOnly`) without adding the key to the file.
- Deprecate the `game_logs` preference in favor of `game_state_logs`, to avoid unexpectedly writing game state logs to the new HOME path.

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -204,9 +204,9 @@ void GameGlobalInfo::startScenario(string filename)
     }
 
     // Deprecate the game_logs setting.
-    if (PreferencesManager::get("game_logs", "0").toInt())
+    if (PreferencesManager::getOnly("game_logs") == "1")
     {
-        LOG(WARNING) << "The game_logs setting is deprecated. To record game state logs, set game_state_logs=1 or use the Options menu.";
+        LOG(WARNING) << "The game_logs setting is deprecated. To enable game state logging, set game_state_logs=1 or use the Options menu.";
     }
 }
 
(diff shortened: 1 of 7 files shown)


oznogon:
Draft PR for feedback. Game state logging is difficult to enable on macOS by default because of the use of a relative path for `logs/`, and the setting itself can be unintuitive (enabled but non-functional out of the box, always adding a warning message to the log due to `logs` not existing, and not indicating how to actually enable logging).

Deprecating the setting is a breaking change for people already using game state logging. An alternative might be to make the state logging path configurable.

oznogon:
No feedback; closing.

PR #1057, opened 2020-09-09
Resolve minor compiler warnings

oznogon:
- EMissileSizes can't be appended to strings
- specify onMouseDown override in objectCreationView


oznogon:
Extraneous change left in, will refile

Issue #889, opened 2020-04-13
Shortening long range radar range is not fully functional


Xansta:
Start player ship at 25 units of long range radar range, increase to 35 units of long range radar range, then change it back to 25 units of long range radar range. The longer 35 unit range is still available.

The scroll wheel on the mouse cannot access it, but the zoom widget lets you zoom to 0.7 X and get the larger scan range. Also, if you're already zoomed to maximum when the range shrinks, the long range stays there if you don't fiddle with the zoom.

daid:
Yeah, I knew the ranges could act a bit funky if you changed them while someone is looking at a station. But I think only science is problematic now.

oznogon:
Closed by #1061?

PR #1061, opened 2020-09-10
Recalculate science zoom in onDraw

StarryWisdom:
This prevents science from seeing outside of the maximum range
when it is altered by scripting or tweak menus.
Fixes:#889

### DIFF
--- a/src/screens/crew6/scienceScreen.cpp
+++ b/src/screens/crew6/scienceScreen.cpp
@@ -198,6 +198,8 @@ ScienceScreen::ScienceScreen(GuiContainer* owner, ECrewPosition crew_position)
     // Draw the zoom slider.
     zoom_slider = new GuiSlider(radar_view, "", my_spaceship ? my_spaceship->getLongRangeRadarRange() : 30000.0f, my_spaceship ? my_spaceship->getShortRangeRadarRange() : 5000.0f, my_spaceship ? my_spaceship->getLongRangeRadarRange() : 30000.0f, [this](float value)
     {
+        if (my_spaceship)
+            zoom_label->setText(tr("Zoom: {zoom}x").format({{"zoom", string(my_spaceship->getLongRangeRadarRange() / value, 1)}}));
         science_radar->setDistance(value);
     });
     zoom_slider->setPosition(-20, -20, ABottomRight)->setSize(250, 50);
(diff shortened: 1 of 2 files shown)


src/screens/crew6/scienceScreen.cpp
```diff
@@ -223,24 +221,18 @@ void ScienceScreen::onDraw(sf::RenderTarget& window)
     GuiOverlay::onDraw(window);
     P<ScanProbe> probe;
 
-    // Handle mouse wheel
-    float mouse_wheel_delta = InputHandler::getMouseWheelDelta();
-    if (mouse_wheel_delta != 0.0 && my_spaceship)
```
oznogon, line 228: Removing the conditional means the view distance, zoom slider, and zoom label are recalculated every onDraw tick, even when nothing has changed.

Is there a way for script or tweak menu radar range changes to signal/trigger/callback an update, instead of updating every tick?

StarryWisdom: I think that callbacks arent pratical due to it being on the client and the callbacks being server side, I think it would end up looking like
```
    float view_distance = science_radar->getDistance();
    float mouse_wheel_delta=InputHandler::getMouseWheelDelta();
    if (mouse_wheel_delta!=0)
      view_distance *= (1.0 - (mouse_wheel_delta * 0.1f));
    view_distance = std::min(view_distance,my_spaceship->getLongRangeRadarRange());
    view_distance = std::max(view_distance,my_spaceship->getShortRangeRadarRange());
    if (view_distance!=science_radar->getDistance())
    {
      science_radar->setDistance(view_distance);
      // Keep the zoom slider in sync.
      zoom_slider->setValue(view_distance)->setRange(my_spaceship->getLongRangeRadarRange(),my_spaceship->getShortRangeRadarRange());
      zoom_label->setText(tr("Zoom: {zoom}x").format({{"zoom", string(my_spaceship->getLongRangeRadarRange() / view_distance, 1)}}));
    }
```
I think it is easier to read without the branching logic, and I think the performance impact is likely to be minimal, with the possible exepction of the translation - I dont know how heavy that is.

If people think the above is better I am happy to edit and change and test it though

oznogon: Yeah, I don't see anything egregious here. Of the crew stations that tend to strain lower-end systems, science is the heaviest, so my concern is mostly about adding to that.

How often are people changing radar view distance on ships in the middle of a game?

StarryWisdom: I don't feel strongly on it so I will change it.
In testing it there is also a need to cache the last values for long and short range radar and compare those as well.
I kind of wonder if the logic to check if it needs updating may consume more time than the update itself, it feels branchy but without knowing how expensive the tr call is I have no instinct in the matter (and strings can be expensive).

I have seen small issues with the scanning popup locally - they felt very gui related, I don't know if they are the same ones as low end machines though.

oznogon: (I didn't mean "How often are people changing..." to come off as contrarian, sorry, I genuinely don't know)

oznogon: I don't have any serious concerns with the PR, and if it fixes a problem, it's worth having

daid: Overhead of these checks is really low.

daid:
Good addition, know this could happen but did not see it as a huge issue when I added the ship radar range settings.

Issue #1062, opened 2020-09-11
trigger hardware/DMX event when missile is launched


mhthies:
Is there a way to reliably trigger a hardware/DMX *event*, when any missile is launched? I know, I can use `TubeFiring[1-15]` as a trigger variable, but it only seems to work for HVLI missiles.

I assume that, when launching another missile, the `weapon_tube.isFiring()` state is `true` for a too short period of time to be detected by the hardwareController. Or am I missing something else?

oznogon:
I'm not experienced with hardware/DMX events, but as a starting point, HVLI is the only weapon type that puts the tube into the firing state when fired. Other weapons don't hit that state at all: https://github.com/daid/EmptyEpsilon/blob/5869678b0c9f117f379aa1586a61286c934eb1a3/src/spaceObjects/spaceshipParts/weaponTube.cpp#L87-L105

NinjaSpectre:
Most likely another boolean, `hasFiredRecently` or similar, would need to be set and then reset after "x" update cycles.  I don't have enough knowledge as to what an acceptable value for "x" would be though.

J-Pitherin:
Has there been any movement on this issue lately? It would be nice to be able to trigger different sound effects through DMX for when a tube is unloaded vs when a tube is being fired.

PR #1063, opened 2020-09-12
Weapon tube firing state

NinjaSpectre:
#1062

`WTS_Firing` state was only being set for loaded missile type `MW_HVLI`


daid:
This is intentional, as it is the only missile type that takes a non-zero time to fire.

PR #1066, opened 2020-09-16
Add weapon tube fired event

NinjaSpectre:
#1062

Add `TubeFired` to hardware events


src/spaceObjects/spaceshipParts/weaponTube.cpp
```diff
@@ -91,6 +92,7 @@ void WeaponTube::fire(float target_angle)
     if (parent->docking_state != DS_NotDocking) return;
     if (parent->current_warp > 0.0) return;
     if (state != WTS_Loaded) return;
+    has_fired = true;
```
daid, line 95: This won't work properly on clients (as the fire function is only called on the server)

src/spaceObjects/spaceshipParts/weaponTube.cpp
```diff
@@ -284,6 +285,13 @@ bool WeaponTube::isUnloading()
     return state == WTS_Unloading;
 }
 
+bool WeaponTube::hasFired()
+{
+    bool flag = has_fired;
+    if (has_fired) has_fired = false;
```
daid, line 291: It's very implicit behavour that a "hasFired" function resets the flag as well. Not saying I have a better solution, but it's not clear from the function name that calling it will change internal state of the weapon tube.

daid (changes requested):

src/spaceObjects/spaceshipParts/weaponTube.cpp
NinjaSpectre: I guess I'm missing something here as I don't see why the client should be allowed to set this value at all.  Perhaps the entire function should be guarded inside a `if (game_server)` check?

NinjaSpectre: Obviously I didn't initially comprehend what you meant.... I can be a bit daft sometimes...

hemmond:
I'm wondering why do we need info if a tube has fired. It it is for mission scripting, I would think it would be better to have a callback function to be called when tube (or any tube on ship) fired. So it will simply run a callback (or nil) on every missile launch (or until callback has been set back to nil).

NinjaSpectre:
This is to have a software trigger so that a [Digital Multiplex (DMX) controller](https://github.com/daid/EmptyEpsilon/wiki/DMX-Configuration) can be used to display when a weapon tube has been fired.  Currently there is only a trigger for when a tube is constantly firing, a condition that only happens when HVLI missiles are fired.

oznogon:
Filed #2684 as a followup.

PR #1074, opened 2020-10-02
Ask for supply ship with jump drive

Xansta:
Some players said that they rarely request a supply drop from a station because the ship takes so long to get to where they need it. So, I added an option to choose a ship with a jump drive to provide the supply drop. It requires more reputation.


Xansta:
Another time perhaps

PR #1076, opened 2020-10-06
fix: black holes and worm holes do not affect zones, beams and electric explosions #961

### LINKED ISSUE #961: Feature request: Be able to configure zones to ignore black holes
I would like for some zones to just be markers in space for regions. If the zone gets sucked into a black hole, it makes it difficult. Can there be an option to make zones immune to black holes? I'd prefer for it to be a script exposed option rather than a global "all zones must ignore black holes." I can see where zones being affected by black holes could be useful.

czenker:
closes #961

### DIFF
--- a/src/spaceObjects/blackHole.cpp
+++ b/src/spaceObjects/blackHole.cpp
@@ -65,6 +65,10 @@ void BlackHole::collide(Collisionable* target, float collision_force)
     if (update_delta == 0.0)
         return;
 
+    P<SpaceObject> obj = P<Collisionable>(target);
+    if (!obj) return;
+    if (!obj->hasWeight()) { return; } // the object is not affected by gravitation
+
     sf::Vector2f diff = getPosition() - target->getPosition();
     float distance = sf::length(diff);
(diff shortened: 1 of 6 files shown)


src/spaceObjects/blackHole.cpp
```diff
@@ -65,6 +65,10 @@ void BlackHole::collide(Collisionable* target, float collision_force)
     if (update_delta == 0.0)
         return;
 
+    P<SpaceObject> obj = P<Collisionable>(target);
```
czenker, line 68: @daid As far as I know all Collisionables inside EE are also SpaceObjects, right?

daid: Yes, but still, checking is good practice and very cheap.

czenker: I'm always checking. :laughing: 

I was asking, because with returning early any non-spaceObject Collisionable would not be moved as before. I could have made this clearer in the question.

daid: If it is a spaceobject, we do not really care about it now, as no wait to check if it has weight :)

Issue #1038, opened 2020-08-25
server settings and tweak or scripting


StarryWisdom:
It would be nice if most or all of the server settings where either available in the tweak menu or via scripting.

In some cases it may be possible to remove them from the main server setting window after doing so.

Brought to you by a GM who selected the wrong option for scan complexity before spending about 30 minutes setting up.

czenker:
I created #1077 to add a possibility to read the settings via a script.

I was also having the issue, that some missions, I came up with, do not make sense if a certain setting is not set. For instance I have one where the players are asked to stop a ship from escaping without destroying it. Targeting the impulse drive is a good way to achieve that. But this kind of mission does not make sense if Per-System Damage is turned off. Now it is possible to disable a mission like that or show an error when starting the scenario. This would probably also help you, @StarryWisdom.

About **writing the settings**: I browsed the code a little and I don't see any major problem with changing the settings during a running game. But still I don't have a good feeling about it yet.

PR #1077, opened 2020-10-08
expose scenario server settings for scripts #1038

czenker:
As considered in #1038 

I only added functions to *read* the settings for now. I don't think writing them during a running game is a bad idea per se, but I am a bit relcutant.


daid:
Yeah, setting them at runtime could cause problems. Also confusion for the person who started the server, like "I set this at X, why is it at Y now?"

PR #1082, opened 2020-10-10
fix parsing of the author of a scenario

czenker:
The author is not used anywhere, but still it was obviously wrong.


Xansta:
I'm curious as to how the "Author" feature is supposed to be used. Can someone explain or provide an example?

czenker:
The author information is not used at all in EE. But you can put it in your scenario files as human-readable meta information and the parser will *not* issue a warning.

```lua
-- author: John Doe
```

PR #1087, opened 2020-10-11
hide call signs of unscanned ships

czenker:
This pull request is up for discussion. 

I always thought that it was better to not show the call signs of ships that have not been scanned, because the name can give a lot away: The "USF Michigan" is probably not just a small fighter and "Crimson Raider" and "Skullcrusher" flying at you are probably not up to any good.

It also gives one more incentive for Science to scan a ship, because the other stations can not communicate efficiently about it if they do not know a ships name.

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -644,8 +644,16 @@ void GuiRadarView::drawObjects(sf::RenderTarget& window_normal, sf::RenderTarget
             if (!obj->canHideInNebula())
                 window = &window_alpha;
             obj->drawOnRadar(*window, object_position_on_screen, scale, view_rotation, long_range);
+
             if (show_callsigns && obj->getCallSign() != "")
-                drawText(*window, sf::FloatRect(object_position_on_screen.x, object_position_on_screen.y - 15, 0, 0), obj->getCallSign(), ACenter, 15, bold_font);
+            {
+                P<SpaceShip> ship = P<SpaceObject>(obj);
+                if (!my_spaceship || !ship || ship->getScannedStateFor(my_spaceship) >= SS_SimpleScan)
+                {
+                    // only show callsign of scanned space ships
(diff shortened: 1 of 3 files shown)


Xansta:
Sounds reasonable to me. Will the call sign appear once the ship has been identified as friend or foe? In particular, if the ship fires on the player or a friendly target, it turns red. The question is, is that enough to reveal the call sign?

czenker:
The call sign will show after a simple scan. FoF-identification won't reveal its call sign. This is the same point when the faction of a ship is revealed.

daid:
> because the other stations can not communicate efficiently about it if they do not know a ships name.

Which is a big reason for me not to consider this change. The whole "meta" game is about communication between the crew, so hampering that doesn't sound like a good idea.

If you want to hide the callsign of certain ships until scanned, you can handle this (in a limited way) with scripting.

StarryWisdom:
If it is done I would like a way to enable / disable it via scripting, gm contols or server settings.

syelle:
> > because the other stations can not communicate efficiently about it if they do not know a ships name.
> 
> Which is a big reason for me not to consider this change. The whole "meta" game is about communication between the crew, so hampering that doesn't sound like a good idea.
> 
> If you want to hide the callsign of certain ships until scanned, you can handle this (in a limited way) with scripting.

I do agree that people will need to communicate about an unscanned ship. I also agree that callsigns can leak info that reduces the need for crew to exchange information.

Some possibilities:

1. People can refer to the ship by its heading until it has a name (e.g. "Captain, I've detected a ship at bearing 135." -> "Scan the ship at bearing 135.")
1. Any station that can see the unscanned vessel can assign a name to an unidentified ship via the UI. Other stations can see this temporary name. This name is automatically replaced with the vessel's callsign when it's available. 
1. Each ship gets some auto-generated ID as a placeholder callsign until scanned (e.g. "UV-001"). People can use that to reference the ship until they're able to extract a callsign.

As a player, I think I'd prefer one of the first two options. The first option seems easiest to implement, but would be problematic if you had a large cluster of ships at a certain heading. The second option solves for this, and another provides a low-effort task that stations can perform.

PR #1088, opened 2020-10-11
expose science and relay selection to scripting

czenker:
This allows to write functions that are dependend on the target selection of science or relay.


daid:
This breaks having multiple relay/science stations pretty hard.

PR #1092, opened 2020-10-13
add the AI "evasion" that flees from enemies

czenker:
This AI is supposed to be used with transports and other unarmed ships. It tries to escape from enemies – if it is fast enough it might even fly around them.

It only acts on "Fly Towards" and "Dock" commands, so any attack specific commands and the "Fly Towards (Ignore Enemies)" command are unaffected.

````lua
ship:setAI("evasion")
````


daid:
YAY, finally some more use of the AI system :D

PR #1095, opened 2020-10-14
gameGlobalInfo: getSectorName

capt-t:
Make `getSectorName` for coordinates available in Lua.

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -246,6 +246,17 @@ string getSectorName(sf::Vector2f position)
     return y + x;
 }
 
+int getSectorName(lua_State* L)
+{
+    float x = luaL_checknumber(L, 1);
+    float y = luaL_checknumber(L, 2);
+    lua_pushstring(L, getSectorName(sf::Vector2f(x, y)).c_str());
+    return 1;
+}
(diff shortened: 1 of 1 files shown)


capt-t:
Suggestions to fix this?

capt-t:
Okay.

Could we replace
```cpp
string getSectorName(sf::Vector2f position)
{
    int sector_x = floorf(position.x / sector_size) + 5;
```
with the following?
```cpp
string getSectorName(lua_State* L)
{
    float x = luaL_checknumber(L, 1);
    // ...
    int sector_x = floorf(x / sector_size) + 5;
}
```
But this changes the signature.

Or is it better to introduce this function?
```cpp
string getSectorNameXY(lua_State* L)
{
    return getSectorName(x, y);
}
```

NinjaSpectre:
Close... it can be done by adding the following.  I named the additional function the same.
```c++
int getSectorName(lua_State* L)
{
    float x = luaL_checknumber(L, 1);
    float y = luaL_checknumber(L, 2);
    sf::Vector2f position(x, y);
    lua_pushstring(L, getSectorName(position).c_str());
    return 1;
}
```

capt-t:
I see the interaction between C++ and Lua.

Thank you!

PR #1098, opened 2020-10-15
Add getGMInfo to several spaceObjects

capt-t:
There might be small things that can be improved.

Please review/comment/fix/merge.


capt-t:
This can be merged when the compiled version has been tested.

Further refinemens and additions can come after the next release.

src/spaceObjects/asteroid.cpp
```diff
@@ -65,6 +65,13 @@ void Asteroid::drawOnRadar(sf::RenderTarget& window, sf::Vector2f position, floa
     window.draw(object_sprite);
 }
 
+std::unordered_map<string, string> Asteroid::getGMInfo()
+{
+    std::unordered_map<string, string> ret = SpaceObject::getGMInfo();
+    ret["typeName"] = "Asteroid";
```
daid, line 71: No need for these typeName, as the GM screen could just use the `getMultiplayerClassIdentifier` from the MultiplayerObject base class to get the type of any object.

src/spaceObjects/scanProbe.cpp
```diff
@@ -112,6 +112,13 @@ void ScanProbe::drawOnGMRadar(sf::RenderTarget& window, sf::Vector2f position, f
     }
 }
 
+std::unordered_map<string, string> ScanProbe::getGMInfo()
+{
+    std::unordered_map<string, string> ret = SpaceObject::getGMInfo();
+    ret["Lifetime"] = string(lifetime);
```
daid, line 118: Inconsistent upper/lower case, this goes for more keys of the GMInfo.

src/spaceObjects/spaceObject.cpp
```diff
@@ -304,6 +304,24 @@ void SpaceObject::drawOnGMRadar(sf::RenderTarget& window, sf::Vector2f position,
...
 }
 
+std::unordered_map<string, string> SpaceObject::getGMInfo()
+{
+    std::unordered_map<string, string> ret = std::unordered_map<string, string>();
+
+    ret["Scanning Depth * Compl."] = string(scanning_depth_value) + " * " + string(scanning_complexity_value);
+
+    if (object_description.not_scanned != "")
+       ret["Descr. 0 (NotSc)"] = object_description.not_scanned;
```
daid, line 314: Descriptions are be long and are part of the tweak menu, don't think it is smart to put them here.

daid (changes requested):

StarryWisdom:
it would be nice if Artifacts callsign was exposed to via the gm interface (we often use them as markers for players to go to that location).

Thinking about it - callsigns for everything would be nice. It is something that is very immedate for players and quite slow to find for the GM.

Supply drop contents would be nice to be exposed (missiles and energy, I need to add that to the tweak menu as well at some point).

src/spaceObjects/asteroid.cpp
capt-t: 1. Simply `ret["typeName"] = getMultiplayerClassIdentifier()` for `SpaceObject`? (That would be what I was looking for.)

2. Any suggestion for "typeName"?

PR #1099, opened 2020-10-15
Draft to expose selections to custom functions

czenker:
This is my second take on #1088 to make the radar selection on a players screen available to scripts. The idea is that `commandCustomFunction` can also send a `multiplayer_id` to the server.

So far so good. The code compiles, but it `SEGFAULT`s with the following error as soon as a custom button is clicked:

```
Thread 1 "EmptyEpsilon" received signal SIGSEGV, Segmentation fault.
0x00000000007f7858 in GuiRadarView::getTarget (
    this=0xffffff00ffffff)
    at EmptyEpsilon/src/screenComponents/radarView.cpp:811
811	    if (targets)
```

From what I found out, I am probably experiencing something called a _dangling reference_ inside the lambda. I assume it is caused, because C++ does not keep track of bound variables inside lambdas and could destroy objects. 

I gave it a few nights of sleep and multiple different approaches, but I could not figure out how to prevent the segfault. 

Is there anyone who can help?

### DIFF
--- a/src/screenComponents/customShipFunctions.cpp
+++ b/src/screenComponents/customShipFunctions.cpp
@@ -1,5 +1,6 @@
 #include "customShipFunctions.h"
 #include "playerInfo.h"
+#include "radarView.h"
 #include "spaceObjects/playerSpaceship.h"
 #include "gui/gui2_button.h"
 #include "gui/gui2_label.h"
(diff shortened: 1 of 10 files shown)


czenker:
Nevermind – it was my own blindness and an expectation of non-existant contructor magic. Stay tuned for an update.

PR #1103, opened 2020-10-16
expose target selection to scripts when custom button is clicked

czenker:
This is my second take on #1088 and works with multiple similar stations in the same ship.

It can be used like this in scripts:

´´´lua
local player = PlayerSpaceship():setFaction("Human Navy"):setCallSign("Alpha"):setTemplate("Atlantis"):setPosition(-1000, 0)
    CpuShip():setFaction("Kraylor"):setCallSign("Beta"):setTemplate("Personnel Freighter 2"):setPosition(1000, -1000):orderRoaming()
CpuShip():setFaction("Kraylor"):setCallSign("Gamma"):setTemplate("Personnel Freighter 2"):setPosition(1000, 0):orderRoaming()
CpuShip():setFaction("Kraylor"):setCallSign("Delta"):setTemplate("Personnel Freighter 2"):setPosition(1000, 1000):orderRoaming()

player:addCustomButton("weapons", "weapons", "weapons", function(self, target)
    print("weapons", self:getCallSign(), target and target:getCallSign() or nil)
end)
player:addCustomButton("science", "science", "science", function(self, target)
    print("science", self:getCallSign(), target and target:getCallSign() or nil)
end)
```

### DIFF
--- a/src/screenComponents/customShipFunctions.h
+++ b/src/screenComponents/customShipFunctions.h
@@ -3,11 +3,13 @@
 
 #include "playerInfo.h"
 #include "gui/gui2_autolayout.h"
+#include "screenComponents/radarView.h"
 
 class GuiCustomShipFunctions : public GuiAutoLayout
 {
 public:
     GuiCustomShipFunctions(GuiContainer* owner, ECrewPosition position, string id);
+    GuiCustomShipFunctions(GuiContainer* owner, ECrewPosition position, string id, TargetsContainer* targets);
 
     virtual void onDraw(sf::RenderTarget& window) override;
 
@@ -19,6 +21,8 @@ private:
         string name;
         GuiElement* element;
     };
+
+    TargetsContainer* targets;
 
     ECrewPosition position;
     std::vector<Entry> entries;
--- a/src/spaceObjects/playerSpaceship.cpp
+++ b/src/spaceObjects/playerSpaceship.cpp
@@ -63,6 +63,8 @@ REGISTER_SCRIPT_SUBCLASS(PlayerSpaceship, SpaceShip)
     REGISTER_SCRIPT_CLASS_FUNCTION(PlayerSpaceship, setMaxScanProbeCount);
     REGISTER_SCRIPT_CLASS_FUNCTION(PlayerSpaceship, getMaxScanProbeCount);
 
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: src/spaceObjects/playerSpaceship.cpp [hunk 1]
```diff
@@ -1633,14 +1635,24 @@ void PlayerSpaceship::onReceiveClientCommand(int32_t client_id, sf::Packet& pack
...
+            packet >> target_id;
+
             for(CustomShipFunction& csf : custom_functions)
             {
                 if (csf.name == name)
                 {
                     if (csf.type == CustomShipFunction::Type::Button || csf.type == CustomShipFunction::Type::Message)
                     {
-                        csf.callback.call();
+                        P<SpaceObject> target = nullptr;
```
daid, line 1648: I think the code can be simpler, as `getObjectById` will return nullptr on a number that isn't "valid" (including 0 and -1) and a nullptr will translate into a nil on lua.

So it could be reduced to a single line `csf.callback.call(P<PlayerSpaceship>(this), game_server->getObjectById(target_id));`

czenker: I have already tried this, but `nullptr` in call arguments was not handled yet.

But since you asked... :smile: [Here is the pull request](https://github.com/daid/SeriousProton/pull/55) in Serious Proton.

czenker:
Requires https://github.com/daid/SeriousProton/pull/55 to be merged.

src/screenComponents/customShipFunctions.h
```diff
@@ -3,11 +3,13 @@
 
 #include "playerInfo.h"
 #include "gui/gui2_autolayout.h"
+#include "screenComponents/radarView.h"
 
 class GuiCustomShipFunctions : public GuiAutoLayout
 {
 public:
     GuiCustomShipFunctions(GuiContainer* owner, ECrewPosition position, string id);
+    GuiCustomShipFunctions(GuiContainer* owner, ECrewPosition position, string id, TargetsContainer* targets);
```
daid, line 12: Instead of 2 constructors we can just have 1 with a default:
```GuiCustomShipFunctions(GuiContainer* owner, ECrewPosition position, string id, TargetsContainer* targets=nullptr);```

PR #1125, opened 2020-11-12
Scenario Scurvy Scavenger

Xansta:
New scenario


Xansta:
Bug observed when starting scenario unpaused (ie headless). To fix, place this line at the top of the init() function:
`	accumulated_delta = 0`

This will eventually get fixed, but if you want to move on to play this with a headless server...

PR #1135, opened 2020-11-19
Scenario comments

capt-t:
Improve comments for scenarios 00 to 10 and some script files.

The dummy comments are necessary for tools like ldoc that need a comment starting with `---` to include a function in the documentation.

I combined similar changes in one commit, I do not recommend to squash these commits too much further.


scripts/perlin_noise.lua
```diff
@@ -30,6 +31,7 @@ perlin.gy = {}
 perlin.randMax = 256
 perlin.loaded = false
 
+--- The method load.
```
daid, line 34: meaningless redundant comment is redundantly meaningless.

Documentation should state "why" not "what" the "what" is already in the code.

scripts/scenario_00_basic.lua
```diff
@@ -62,19 +63,21 @@ function addWave(enemyList, type, a, d)
...
 end
 
--- Returns a semi-random heading.
--- cnt: A counter, generally between 1 and the number of enemy groups.
--- enemy_group_count: A number of enemy groups, generally set by the scenario type.
+--- Returns a semi-random heading.
+--
+-- @tparam number cnt A counter, generally between 1 and the number of enemy groups.
+-- @tparam number enemy_group_count A number of enemy groups, generally set by the scenario type.
 function setWaveAngle(cnt, enemy_group_count)
```
daid, line 70: If we are cleaning this up, "setWaveAngle" doesn't set anything. So "calcWaveAngle" would be a better name.

scripts/scenario_00_basic.lua
```diff
@@ -62,19 +63,21 @@ function addWave(enemyList, type, a, d)
...
     return cnt * 360 / enemy_group_count + random(-60, 60)
 end
 
--- Returns a semi-random distance.
--- enemy_group_count: A number of enemy groups, generally set by the scenario type.
+--- Returns a semi-random distance.
+-- @param enemy_group_count A number of enemy groups, generally set by the scenario type.
 function setWaveDistance(enemy_group_count)
```
daid, line 76: See setWaveAngle

daid (commented):
I don't like all the `--- function` comments everywhere, document a function, or don't, but don't pretend they are documented when they are not.

scripts/scenario_00_basic.lua
capt-t: I agree, the code is changed in a different pull request.

capt-t:
I have adjusted the pull request. The functions are renamed in #1141.

PR #1140, opened 2020-11-20
Scenario order

capt-t:
From short to long, easy to complex, fascinating to frustrating.

Special (e.g. PvP) or unplayable (battlefield) scenarios at the end.

Much more friendly for new groups of players.


daid:
You didn't rename the locale files. Idea to reorder them is good however.

FaramosCZ:
I was quite frustrated, when making custom scenarios, that I had to search for them amongst others everytime I wanted to test them, which was quite often when I was developing them.
So I made three scenarios, and all named "scenario_00_*", so they would appear at the beginning. Still they wasn't the first three, since the "b" character in "scenario_00_basic.lua" is at the beginning of the alphabet.

So totally :+1:  from me :)

> From short to long, easy to complex, fascinating to frustrating.
> Special (e.g. PvP) or unplayable (battlefield) scenarios at the end.
> Much more friendly for new groups of players.

From what you wrote, I would understand you would like an option, to sort and filter the scenarios in-game - which would also be interesting - but the commits you made suggests you only meant to update the hard-coded values with prettier hard-coded values.

aBlueShadow:
@FaramosCZ just add one more zero, and they will appear at first (btw this PR won't change anything about that situation).
`Scenario_000* `is how I name my testing scenarios.

capt-t:
If we agree on the order and on this PR, I will add the reordering to the affected language files.
Please tell me today.

capt-t:
Done.

PR #1150, opened 2020-11-25
Reusable comms functions

capt-t:
Require `comms.lua` and reuse the functions.


capt-t:
I will change a little bit more, so this PR can wait until then.

daid:
Note that there are also conflicts now on this PR.

daid:
With the `require()` comms callback bug fixed, I think this can be revisisted?

PR #1153, opened 2020-11-28
Sciencedbmodifs value2811

muerteFR:
1. Translation of values ​​and Mine modification: drop distance and trigger distance
2. I continued the corrections and the translation of the interface: 3/4 players, one player ...
However, there are still a lot of words to translate in interface, 1P, GM, Main screen control...but they concern .cpp files.

Do I have to do fewer Pull requests and come up with more changes all at once or it's ok? I am new to Github and am not yet comfortable. Thanks


daid:
Small pull requests like this is fine. It is much easier to check small changes then big changes

muerteFR:
Understood.
Thanks

PR #1156, opened 2020-11-30
scenario_05: fix typos

capt-t:


daid:
Only 3? Damn, usually I make a lot more typos :D

PR #1158, opened 2020-11-30
Improve GitHub Actions to Build Artifacts

k9lego:
This change uses GitHub-Actions to build "testing" target releases on every commit/with every pull request.
It uses `ubuntu-latest` (18.04, soon to be 20.04 within the next day or two) to build Windows, Android, and Linux targets. It also builds a testing MacOS release, even though it is to be expected that MacOS will not be ever officially supported.

My proposal is this: Include this in your repository to encourage testing, and make it clear that these builds are NOT supported in any fashion unlike the official release builds. The MacOS build is helpful so those who have a Mac can use EmptyEpsilon even if it is not officially supported. You could potentially consider putting it on the main site as [UNTESTED, UNSUPPORTED] like Android is [BETA], just to provide easy access to the Mac builds, though the downside is people might still think it's "officially supported".

You need to be logged into GitHub to download the artifacts produced by these actions, so this will help prevent random people from downloading them thinking they are an official release. I hope that you will consider this Pull Request! Especially since GitHub Actions (including MacOS time) and Artifact Storage both are free to Public Repositories. Thanks!

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -6,15 +6,81 @@ on:
       - master
   pull_request:
 jobs:
-  build:
-    name: Build and test
+  ubuntu:
+    name: Build Linux Deb on Ubuntu
     runs-on: ubuntu-latest
     steps:
-    - name: Update apt index
+    - name: Update Container
       run: sudo apt-get update -qq
     - name: Instal SFML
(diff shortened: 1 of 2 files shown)


k9lego:
Forgot to make a separate branch, and made a commit to this (oops).

PR #1159, opened 2020-12-01
 Improve GitHub Actions to Build Artifacts

k9lego:
Change: This change uses GitHub Actions to build "testing" target releases on every commit/with every pull request.

It uses ubuntu-latest (18.04, soon to be 20.04 within the next day or two) to build Windows, Android, and Linux targets. It also builds a testing MacOS release, even though it is to be expected that MacOS will not be ever officially supported.

My proposal is this: Include this in your repository to encourage testing, and make it clear that these builds are NOT supported in any fashion unlike the official release builds. The MacOS build is helpful so those who have a Mac can use EmptyEpsilon even if it is not officially supported. You could potentially consider putting it on the main site as [UNTESTED, UNSUPPORTED] like Android is [BETA], just to provide easy access to the Mac builds, though the downside is people might still think it's "officially supported".

You need to be logged into GitHub to download the artifacts produced by these actions, so this will help prevent random people from downloading them thinking they are an official release. I hope that you will consider this Pull Request! Especially since GitHub Actions (including MacOS time) and Artifact Storage both are free to Public Repositories. Thanks!

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -6,15 +6,81 @@ on:
       - master
   pull_request:
 jobs:
-  build:
-    name: Build and test
+  ubuntu:
+    name: Build Linux Deb on Ubuntu
     runs-on: ubuntu-latest
     steps:
-    - name: Update apt index
+    - name: Update Container
       run: sudo apt-get update -qq
     - name: Instal SFML
       run: sudo apt-get install libsfml-dev
     - name: Checkout EmptyEpsilon
       uses: actions/checkout@v2
-    - name: Build and test
-      run: docker/build.sh
+    - name: Build
+      run: docker/build.sh linux
+    - name: Upload EmptyEpsilon.deb artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: Linux Debian
+        path: /home/runner/work/EmptyEpsilon/EmptyEpsilon/build/EmptyEpsilon.deb
+        if-no-files-found: error
+    - name: Upload script_reference.html artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: Script Reference
+        path: /home/runner/work/EmptyEpsilon/EmptyEpsilon/build/script_reference.html
+        if-no-files-found: error
+  win32:
+    name: Build Win32 Exe on Ubuntu
+    runs-on: ubuntu-latest
+    steps:
+    - name: Update Container
+      run: sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


daid:
As the android build uses a new key every build, installing the android version from actions will require you to uninstall it before you can install a new version...

I'm also not a huge fan of putting a gigabyte of data on github with every commit. No idea if they purge older artifact data, but feels wasteful.

nallath:
We can do that with a cron action. A quick google led me to this: https://github.com/marketplace/actions/purge-artifacts

k9lego:
Android version - good point, what could be done instead is the key file is stored in the EE repo, though I’m not sure if there’s any security concerns there (though the passwords are public...)

By default artifacts are deleted after 90 days, but I could set them to expire in a few
days or a week to prevent the excessive waste. `retention-days: 7`

k9lego:
Made it reuse a keystore and delete all build files (except Script Reference) after 5 days. Let me know if you want any other changes.
I thought of using one machine for Win32/Linux/Android but this feels cleaner in terms of making it clear which build failed, and it is faster too. However, if you want I will change this to one Ubuntu machine for Win32/Linux/Android.

k9lego:
Oh yeah and GH actions decided to take downtime right around when I committed. It worked on my repo when I tested it

daid:
> I’m not sure if there’s any security concerns there (though the passwords are public...)

There is, pretty sure the repo will get flagged for it. So adding a key to the repo feels worse then the generating a new key every time. There is a way to have a "secret" in which only actions on the daid/EmptyEpsilon can read then, but not sure how to set it up https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets

k9lego:
Usage:
Get the Base64 of a Keystore file, then upload it named as "ANDROID_KEYSTORE_BASE64".
https://github.com/k9lego/EmptyEpsilon/settings/secrets/actions/new
Let me know if you need any help! Personally I'd use a separate one from the official releases! :)

nallath:
It seems that uploading the android artifact did fail on your last commit though.

k9lego:
Yeah I tested it on the branch “check” but it failed as a fork for some reason. My suspicion is it’s creating a blank file, I’ll troubleshoot it when I get a chance

k9lego:
Fixed the bug, sorry for the delay! Let me know if you have any feedback!
One thing that bugs me is the unzipping spam for Android, I might try to fix that soon or in another PR.

PR #1163, opened 2020-12-06
Damage info on callback

Tsht:
Additional parameters for using onTakingDamage in LUA scripts.
(on_taking_damage on shipTemplateBasedObject)

This allows scripters to take actions based on damage type, frequency, sub system and hit shield.
These informations are only available if an instigator exists on the DamageInfo.


Tsht:
Just thinking... wouldn't be better to put the callback before the actual damage applies ?

PR #1165, opened 2020-12-07
luax: add table.equals, table.extend, table.clear

capt-t:
More universal functions for lists.

Don't repeat yourself.

### DIFF
--- a/scripts/luax.lua
+++ b/scripts/luax.lua
@@ -6,11 +6,20 @@
 -- Only general functions that could be in the standard library
 -- are added to these modules.
 --
+-- Naming as in Lua's library: lowercase without underscores.
+-- Several functions are inspired by Python.
+--
 -- Require this module in scenario files or modules:
 --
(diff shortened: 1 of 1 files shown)


scripts/luax.lua
```diff
@@ -110,6 +189,35 @@ function table.filter(list, predicate)
...
+-- print(#list, #ref) -- 0   0
+-- print(list == ref) -- true
+-- -- (usually) do not
+-- local list = {"a", "b"}
+-- local ref = list
+-- print(#list, #ref) -- 2   2
+-- list = {}
+-- print(#list, #ref) -- 0   2
+-- print(list == ref) -- false
+function table.clear(list)
```
daid, line 215: Itn't `list = {}` much more efficient?

capt-t: Of course, but it creates a new table, which might not be what you need. See the example.

PR #1173, opened 2020-12-18
comms: prefer string.format to .., use message

capt-t:


daid:
If you could also wrap the strings in `_("...")` at the same time, it would setup those strings for translation, which the french guys will love ;-)

capt-t:
Will be done in a later commit.

PR #1176, opened 2020-12-28
Allow font configuration, add compatible font with lowercase

oznogon:
The Bebas Neue font is hardcoded in `main.cpp`. Clients can replace the `BebasNeue` font files in `resources/gui/fonts/` with other fonts, but must use the `BebasNeue` file names. This complicates supporting multiple languages, using diacritics or lowercase letters, or adding dyslexia-friendly font options from a single client installation.

Instead of hardcoding `gui/fonts/BebasNeue` fonts, add `font_regular` and `font_bold` options to make them configurable. Paths are relative to the `resources` directory. This PR does not change any default behaviors; the default values remain:

```
font_bold=gui/fonts/BebasNeue Bold.otf
font_regular=gui/fonts/BebasNeue Regular.otf
```

This PR also adds ["Big Shoulders"](https://www.fontsquirrel.com/fonts/big-shoulders), another SIL-licensed font with compatible widths and both uppercase and lowercase characters.

Example:

With these lines in `options.ini`:

```
font_bold=gui/fonts/BigShouldersDisplay-ExtraBold.ttf
font_regular=gui/fonts/BigShouldersDisplay-SemiBold.ttf
```

The interface looks like:

image

image

image

image

Arbitrary fonts can be arbitrarily added to `resources/gui/fonts/` on the client side without replacing existing font files, allowing many fonts to be copied into the directory and changed via `options.ini`. Copying Comic Sans into `resources/gui/fonts/` and adding it to `options.ini`:

```
font_bold=gui/fonts/comic.ttf
font_regular=gui/fonts/comicbd.ttf
```

Results in:

image

Returning to default `BebasNeue` is as easy as deleting the `font_...` preferences from `options.ini`.


daid:
Comic sans EE, like god intended ;)

Xansta:
I started using the BigShoulders font. It looks great, however, on Relay, the last line of the ship log no longer appears along the bottom of the screen. When you click the ship log, the log information shows up just fine including the last line, it's just that the last line is not visible when the ship log is collapsed. 

I think this is related to the height of the font. I noticed that the custom messages have fewer visible lines before the scroll bar appears.

I'm not sure whether to make this an issue or not.

PR #1180, opened 2020-12-29
Fix failed assert in debug build with blackholes destroying spaceObjects

StarryWisdom:


daid:
Valid attempt, but not a full fix, as `takeDamage` might destroy the object as well, so after any call to `takeDamage` the pointer could be pointing to a dead object.

PR #1181, opened 2020-12-29
Make destroyed gui elements invisible before they are destroyed

StarryWisdom:
it feels kind of ugly doing it this way, but I couldn't think of a answer for when do you want a visible destroyed element

Other options for fixes that I can see are either
a) adding an isDestroyed function to elements and altering about half a dozen places in auto layout and maybe elsewhere
b) adding a GuiContainer::removeDestroyed function and calling that before anywhere that it is needed

If either of those would be preferred I can look into them, likewise if I am missing another better option

### DIFF
--- a/src/gui/gui2_autolayout.cpp
+++ b/src/gui/gui2_autolayout.cpp
@@ -37,12 +37,12 @@ void GuiAutoLayout::drawElements(sf::FloatRect parent_rect, sf::RenderTarget& wi
             int count = 0;
             for(GuiElement* element : elements)
             {
-                if (element->isVisible())
+                if (!element->isDestroyed() && element->isVisible())
                     count += 1;
             }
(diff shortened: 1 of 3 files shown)


daid:
If this changes anything then we have a different bug, as as soon as an object is destroyed, nothing should be accessing it anymore....

StarryWisdom:
note - I wasn't able to check it works for any of the cases other than LayoutVerticalColumns (where it does stop the issue)
I am unaware of any broken examples for the other ones in the current code base

My knowledge of the gui code is low enough to say if the other isVisible calls outside of gui_autolayout

aBlueShadow:
The underlying bug might be different, but at least it is a workaround that solves #1047
With this PR in place, custom function don't flicker anymore if some of them get replaced or removed.

PR #1183, opened 2020-12-29
Ship selection screen translate 2912

muerteFR:
Addition of 30 tr() occurrences.


src/menus/shipSelectionScreen.cpp
```diff
@@ -298,7 +298,7 @@ ShipSelectionScreen::ShipSelectionScreen()
                 LOG(INFO) << "Password doesn't match control code. Attempt: " << password;
                 my_player_info->commandSetShipId(-1);
                 // Notify the player.
-                password_label->setText("Incorrect control code. Re-enter code for " + ship->getCallSign() + ":");
+                password_label->setText(tr("Incorrect control code. Re-enter code for ") + ship->getCallSign() + ":");
```
daid, line 301: Just like in the other PR, this needs a `.format` to remove the `+` and make it a single proper sentence to translate.

daid (changes requested):

src/menus/shipSelectionScreen.cpp
```diff
@@ -298,7 +298,7 @@ ShipSelectionScreen::ShipSelectionScreen()
                 LOG(INFO) << "Password doesn't match control code. Attempt: " << password;
                 my_player_info->commandSetShipId(-1);
                 // Notify the player.
-                password_label->setText(tr("Incorrect control code. Re-enter code for ") + ship->getCallSign() + ":");
+                password_label->setText(tr("Incorrect control code. Re-enter code for {callsign}").format({{"callsign", ship->getCallSign()}}) + ":");
```
muerteFR, line 301: 1. Is it like line 301 or like line 310? 310 I think.

2. Is it a variable that is between {}? I put what defines the most what it is?

3. If it's the same variable like {callsign} for example it doesn't matter because they don't have any link between them right?
Thanks

PR #1185, opened 2020-12-30
Allow resupply freighter an optional jump drive

Xansta:
This change allows the scenario to ask for the freighter to have a jump drive

Example usage based on comms_station.lua excerpt:

```lua
local position_x, position_y = comms_target:getPosition()
local target_x, target_y = player:getWaypoint(n)
local script = Script()
script:setVariable("position_x", position_x):setVariable("position_y", position_y)
script:setVariable("target_x", target_x):setVariable("target_y", target_y)
script:setVariable("jump_freighter","Yes")  --this is the line to add
script:setVariable("faction_id", comms_target:getFactionId()):run("supply_drop.lua")
setCommsMessage("We have dispatched a supply ship toward WP" .. n);
```

### DIFF
--- a/scripts/supply_drop.lua
+++ b/scripts/supply_drop.lua
@@ -4,6 +4,9 @@
 
 function init()
 	my_ship = CpuShip():setCommsScript("comms_supply_drop.lua"):setFactionId(faction_id):setPosition(position_x, position_y):setTemplate("Flavia"):setScanned(true):orderFlyTowardsBlind(target_x, target_y)
+	if jump_freighter ~= nil and jump_freighter == "Yes" then
+		my_ship:setJumpDrive(true):setJumpDriveRange(5000,25000):setJumpDriveCharge(20000)
+	end
 	state = 0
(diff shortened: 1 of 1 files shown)


daid (commented):
Instead of a string "Yes", we can just use `true` for a cleaner api

Xansta:
I tried `true` and setVariable() denied the use of a boolean: strings only are allowed. I would much prefer to use a boolean.

Error provided by Lua:
`[ERROR] Callback function error: [string "scenario_68_chaos.lua"]:4660: bad argument #2 to 'setVariable' (string expected, got boolean)`

Related code:
```lua
script:setVariable("jump_freighter",true)
```

daid:
Oh, yes, very strict bindings... bah. We can do this better in SP2...

daid:
upper/lower case is my only remaining comment then I guess... I think all strings as parameters are lowercase everywhere else (but I could be wrong, it's messy)

PR #1186, opened 2020-12-30
MSVC native support

gcask:
Native support for MSVC (2017 toolchain, matching SFML).

Currently focusing on 32b only.

Most of was just conformance (now that MS is doing a much better job at it, it's a lot easier).

CMake install targets would need an update (doesn't need all the DLLs from MinGW, requires the SFML ones to be copied over).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,8 +46,12 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     endif()
 endif()
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror=return-type")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror=return-type")
+if(MSVC)
+	add_definitions(-DNOMINMAX)
+else()
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror=return-type")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror=return-type")
+endif()
 
 # install resources, scripts, and packs to app bundle instead of system dir.
 if(APPLE)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -28,7 +28,7 @@
 #include "tutorialGame.h"
 
 #include "hardware/hardwareController.h"
-#ifdef __WIN32__
+#ifdef _WIN32
 #include "discord.h"
 #endif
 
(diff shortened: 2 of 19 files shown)


CMakeLists.txt
```diff
@@ -46,8 +46,12 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     endif()
 endif()
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror=return-type")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror=return-type")
+if(MSVC)
+	add_definitions(-DNOMINMAX)
```
daid, line 50: maybe MSVC needs some warning flags as well? I'm not familiar enough with MSVC to judge that.

src/main.cpp
```diff
@@ -157,7 +157,7 @@ int main(int argc, char** argv)
     new DirectoryResourceProvider("resources/");
     new DirectoryResourceProvider("scripts/");
     new DirectoryResourceProvider("packs/SolCommand/");
-    PackResourceProvider::addPackResourcesForDirectory("packs");
+    PackResourceProvider::addPackResourcesForDirectory("packs/");
```
daid, line 160: As pointed out in the SeriousProton pull request, I don't like the API change.

daid (changes requested):
Other then the API change, it's good.

CMakeLists.txt
gcask: There are several changes that needs to happen yeah - I intend to do a pass on CMake config for msvc later on.

src/main.cpp
gcask: I relaxed the API change.
However, I would argue this is worth keeping for the sake of consistency?
All other paths (besides mod) are currently forward slash terminated...

I'll revert this if it's really a bother though :)

daid: Oh, don't mind it here, only mind the API change :)

src/packResourceProvider.cpp
```diff
@@ -67,6 +69,25 @@ std::vector<string> PackResourceProvider::findResources(const string searchPatte
 
 void PackResourceProvider::addPackResourcesForDirectory(const string directory)
 {
+#ifdef _WIN32
+    WIN32_FIND_DATAA data;
+    assert(directory.endswith("/"));
```
daid, line 74: Hench, this one...

gcask: 🤦 I thought both were in Serious Proton (and thought I dreamed the second one since I could no longer find it 😂 )

daid: Never said the code is well organized ;-)

gcask: It's not *so* bad actually.

Issue #1071, opened 2020-09-27
FEATURE REQUEST: Be able to set Reactor output via lua


fiskn:
It would be very useful to be able to set the reactor output via lua, so it could be defined in the ship template. It currently looks like reactor output is set at 2 units per second at 100% setting. I think this should be customisable per ship type for a few reasons:

1. Why should a big battle ship have the same reactor output as a tiny fighter?
2. For ships that have multiple shields, it could provide some interesting challenges if additional shield generators were available to match number of shields, this would require a bigger reactor
3. In scenarios like carriers & turrets where the fighters draw power from the carrier when docked, the carrier can quickly run out of power, especially if more fighters are spawned than default.

StarryWisdom:
I would use this, though I would ideally like the whole PlayerSpaceship::system_power_user_factor to be accessable to lua and the tweak menu.

A different use case I have contemplated was a ship where the max reactor power decays based on how long its been out of dock.

oznogon:
PR #1244 closes this.

PR #1190, opened 2020-12-31
#1071 feature request reactor output

gcask:
/!\ Draft because it contains the changes of #1186 (msvc support).
I can rebase based on priority (although annoying)

For #1071 : Exposes the values in the tweak menu (screenshot attached), and enables lua access.

The code uses an adjusted (ie fraction) value, however for both lua and tweak we keep the larger value displayed (usually users aren't too ecstatic dealing in infinitesimal increments).

image


daid:
Quick glance looks good, but let's wrap up #1186 first indeed.

gcask:
Closing this, so I can rebase the branch without the msvc noise.

PR #1193, opened 2020-12-31
Update tr() scienceScreen.cpp

muerteFR:
Addition of 1 tr() occurrence for translate.

And on ligne 206 there is : 
 zoom_label = new GuiLabel(zoom_slider, "", "Zoom: 1.0x", 30);

do I have to change it too? I ask because it does not affect the game interface (I think)


daid:
I think that one is directly overruled, so no need to tr tgat one

PR #1194, opened 2020-12-31
Update tr() serverBrowseMenu.cpp

muerteFR:
Addition of 3 tr() occurrences for translate.


muerteFR:
I can't find the errors.

daid:
> I can't find the errors. Should I had #include <i18n.h> ?

Very likely yes

PR #1201, opened 2021-01-01
Update tr() jumpIndicator.cpp

muerteFR:
add 2 tr()


src/screenComponents/jumpIndicator.cpp
```diff
@@ -12,7 +12,7 @@ GuiJumpIndicator::GuiJumpIndicator(GuiContainer* owner)
 
     box = new GuiPanel(owner, "JUMP_BOX");
     box->setSize(800, 100)->setPosition(0, 200, ATopCenter);
-    label = new GuiLabel(box, "JUMP_LABEL", "Jump in: ", 50);
+    label = new GuiLabel(box, "JUMP_LABEL", tr("Jump in: "), 50);
```
daid, line 15: As this gets overruled, it doesn't really need to be translated.

PR #1212, opened 2021-01-01
Update tr() cpuShip.cpp

muerteFR:
add 11 tr()

### DIFF
--- a/src/spaceObjects/cpuShip.cpp
+++ b/src/spaceObjects/cpuShip.cpp
@@ -217,7 +217,7 @@ void CpuShip::orderFlyTowards(sf::Vector2f target)
     orders = AI_FlyTowards;
     order_target = NULL;
     order_target_location = target;
-    this->addBroadcast(FVF_Friendly, "Moving to " + string(target.x) + "," + string(target.y) + ".");
+    this->addBroadcast(FVF_Friendly, tr("cpulog", "Moving to {target},").format({{"target", string(target.x)}}) + string(target.y) + ".");
 }
 
 void CpuShip::orderFlyTowardsBlind(sf::Vector2f target)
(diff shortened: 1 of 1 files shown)


src/spaceObjects/cpuShip.cpp
```diff
@@ -158,7 +158,7 @@ void CpuShip::orderRoamingAt(sf::Vector2f position)
     orders = AI_Roaming;
     order_target = NULL;
     order_target_location = position;
-    this->addBroadcast(FVF_Friendly, "Searching for hostiles around " + string(position.x) + "," + string(position.y) + ".");
+    this->addBroadcast(FVF_Friendly, tr("cpulog", "Searching for hostiles around {posx},").format({{"posx", string(position.x)}}) + string(position.y) + ".");
```
daid, line 161: This looks off:
`tr("cpulog", "Searching for hostiles around {x}, {y}").format({{"x", string(position.x)}, {"y", string(position.y)}})`

src/spaceObjects/cpuShip.cpp
```diff
@@ -286,17 +286,17 @@ string getAIOrderString(EAIOrder order)
...
-    case AI_Retreat: return "Retreat";
-    case AI_StandGround: return "Stand Ground";
-    case AI_DefendLocation: return "Defend Location";
-    case AI_DefendTarget: return "Defend Target";
-    case AI_FlyFormation: return "Fly in formation";
-    case AI_FlyTowards: return "Fly towards";
-    case AI_FlyTowardsBlind: return "Fly towards (ignore all)";
-    case AI_Attack: return "Attack";
-    case AI_Dock: return "Dock";
+    case AI_Idle: return tr("orderscpu", "Idle");
```
daid, line 289: Sadly, we cannot do this like this. As the `getAIOrderString` function is also used for the lua scripting API, so this will break things if a different language is selected.

SystemName had the same issue, see how we solved it here: https://github.com/daid/EmptyEpsilon/blob/master/src/shipTemplate.cpp#L374

muerteFR (commented):
Like that?

PR #1220, opened 2021-01-02
Add language selector to options menu

oznogon:
List is hardcoded to the languages with `.po` files currently in locale. Could potentially be improved by parsing the directory contents via `dirent` to get the list of languages.

Screen Shot 2021-01-01 at 7 43 00 PM
Screen Shot 2021-01-01 at 7 42 49 PM


src/menus/optionsMenu.cpp
```diff
@@ -158,6 +161,18 @@ OptionsMenu::OptionsMenu()
...
 
+    // Language.
+    (new GuiLabel(interface_page, "LANGUAGE", tr("Language"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
+    language_selection = new GuiSelector(interface_page, "LANGUAGE_SELECTION", [](int index, string value)
+    {
+        PreferencesManager::set("language", string(value));
+        i18n::reset();
+        i18n::load("locale/" + PreferencesManager::get("language", "en") + ".po");
+    });
+    language_selection->setOptions({"cz", "de", "en", "fr", "it"})->setSize(GuiElement::GuiSizeMax, 50);
```
daid, line 172: https://github.com/daid/SeriousProton/blob/master/src/resources.h#L39
The `findResources("locale/*.po")` should be able to list all locales.

oznogon: List is now populated from locale files that match most common two- or three-letter locale shortname patterns with two-letter extensions (en.po, en_uk.po, zzz_AA.po, etc.).

aBlueShadow:
Unfortunately this does not work on android (the android version has problems reading/writing files in general.). So maybe fall back to a hardcoded list in this case?

daid:
> Unfortunately this does not work on android (the android version has problems reading/writing files in general.). So maybe fall back to a hardcoded list in this case? Although I am not sure if the po-files can be used at all-on android

Actually, that is a different bug, that could be fixed by backporting this SP2 code:
https://github.com/daid/SeriousProton2/blob/master/src/io/resourceProvider.cpp#L84
to:
https://github.com/daid/SeriousProton/blob/master/src/resources.cpp#L111

oznogon:
> Unfortunately this does not work on android (the android version has problems reading/writing files in general.). So maybe fall back to a hardcoded list in this case? Although I am not sure if the po-files can be used at all-on android

A hardcoded list doesn't work on Android either until @daid's suggestion is implemented.

Is there already a way to set or change the locale on Android? I was under the impression that none of the locale files worked there anyway, so this PR introduces no net change in behavior.

aBlueShadow:
> Is there already a way to set or change the locale on Android? I was under the impression that none of the locale files worked there anyway, so this PR introduces no net change in behavior.

I did not talk about a change of behaviour, in contrary, I thought this PR might have been able to introduce translation to the android build. I did'nt  really have an overview which files can be read (seems true for ship templates and sciencedb) and which can't (options.ini, scenarios, and obviously po-files)

oznogon:
I only mention behavior changes because I don't want Android support to block this, especially since Android support would rely on a change in SP that hasn't happened yet and might be harder to implement.

NinjaSpectre:
I can't make an Android build and test but it looks like converting daid's code from using SDL to SFML may not be too difficult.

Replace `#include <SDL.h>` with `#include <SFML/System/NativeActivity.hpp>`
Add `ANativeActivity *native_activity = sf::getNativeActivity();`
Then `JNIEnv* env = (JNIEnv*)SDL_AndroidGetJNIEnv();` becomes `JNIEnv* env = native_activity->env;`
And `jobject activity = (jobject)SDL_AndroidGetActivity();` becomes `jobject activity = native_activity->clazz;`

daid:
> I only mention behavior changes because I don't want Android support to block this, especially since Android support would rely on a change in SP that hasn't happened yet and might be harder to implement.

Until we have a fix for scenarios breaking due to language selection, I'm not going to merge in a visible language selection. So there is plenty of time ;-)

oznogon:
Closing to reopen after all translation efforts are complete.

PR #1223, opened 2021-01-02
Round floats to nearest integral value on eng

### LINKED ISSUE #949: Float problems with DECREASE_POWER
The hotkey `DECREASE_POWER` reduces from 100 to 89, but it should reduce to 90.

This should be easy to fix by a little rounding in `engineeringScreen.cpp` (and `powerManagement.cpp`). (Other solutions could be adding an epsilon to the initial value or better comparisons for floats than `< 0.3` and then displaying 29.99999 as 30 in the engineering console.)

This is particularly anoying because 0.2999999 < 0.3, but it should _not_ be <.

oznogon:
Fixes #949.

### DIFF
--- a/src/screens/crew6/engineeringScreen.cpp
+++ b/src/screens/crew6/engineeringScreen.cpp
@@ -176,12 +176,12 @@ void EngineeringScreen::onDraw(sf::RenderTarget& window)
             }
         }
 
-        energy_display->setValue(toNearbyInt(my_spaceship->energy_level) + " (" + tr("{energy}/min").format({{"energy", toNearbyInt(average_energy_delta * 60.0f)}}) + ")");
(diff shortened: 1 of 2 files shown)


src/screens/crew6/engineeringScreen.cpp
```diff
@@ -476,3 +475,8 @@ void EngineeringScreen::addSystemEffect(string key, string value)
     }
     system_effects_index++;
 }
+
+string EngineeringScreen::toNearbyInt(float value)
```
daid, line 479: The solution is fine, but the function name is misleading, as you would assume it returns an int from the name.

oznogon: I felt the same way about `nearbyint`, which returns a float.

oznogon: Clarified the function name.

PR #1227, opened 2021-01-02
Don't check non-existent missile owners

oznogon:
When game state logging is enabled, EE might crash when checking
the multiplayer_id of a missile's owner who longer exists.

Write the owner ID only if it still exists.

### DIFF
--- a/src/gameStateLogger.cpp
+++ b/src/gameStateLogger.cpp
@@ -535,9 +535,7 @@ void GameStateLogger::writeMissileEntry(JSONGenerator& json, P<MissileWeapon> mi
     json.write("category_modifier", missile->category_modifier);
 
     // A missile's owner might not exist when we log data. Skip the owner_id if so.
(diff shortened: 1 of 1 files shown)


src/gameStateLogger.cpp
```diff
@@ -533,7 +533,14 @@ void GameStateLogger::writeStationEntry(JSONGenerator& json, P<SpaceStation> sta
 void GameStateLogger::writeMissileEntry(JSONGenerator& json, P<MissileWeapon> missile)
 {
     json.write("category_modifier", missile->category_modifier);
-    json.write("owner_id", missile->owner->getMultiplayerId());
+
+    // A missile's owner might not exist when we log data. Skip the owner_id if so.
+    P<SpaceObject> owner = missile->owner;
+
+    if (owner)
```
daid, line 540: This can just be `if (missile->owner) {}` no need for the local variable. Or else use the local variable to get the ID from.

PR #1241, opened 2021-01-03
Additional CIs, fixing Cross-compile Windows build

gcask:
Added MacOS and Windows cross-compile to the PR CI (so that kind of snafu doesn't happen again).

Time permitting, I'll land android in this week.

Doing the same fix as in https://github.com/daid/SeriousProton/pull/58 for the pack resources.


gcask:
Note - because I'm submitting the two PRs side by side, the Windows build should fail on the SeriousProton issue

PR #1244, opened 2021-01-03
#1071 feature request reactor output

gcask:
Changes for #1071
* Instead of exposing ONLY the reactor output, everything gets exposed, all the "power factors" are exposed
* They're scaled in code (_magic factor_), but we expose raw, larger values in scripts/GM screen.
* See #1190 for the GM screen widget.


gcask:
So apparently the [macos jobs can derp](https://github.com/actions/virtual-environments/issues/841).

Re running the jobs should "fix it".

PR #1249, opened 2021-01-04
Collapse defines into one location (cmake)

gcask:
During my windows native build CI test I ran into a CMake issue with defines (not to worry, turns out it was just a wrong setup on my end).

I ended collapsing the defines into a `target_compile_definitions()` call, which made it easier for me to track them.

Thought it might be worth keeping.


daid:
I preferrer keeping things together that are related. Instead of keeping things together that are the same thing. (You also don't put all your integer variables in a single file because they are all integers :) )

For example, the crash logger define is added where the crash logger library is also added.

PR #1255, opened 2021-01-05
Disabling mac builds

gcask:
Too unstable atm.


k9lego:
Maybe in the future we could use this to build “untested” release artifacts for MacOS?

gcask:
The instability stems from GitHub infrastructure, not us: https://github.com/actions/virtual-environments/issues/841

We could use a manual flow to build artifacts (I guess), but that's an entirely different topic :)

MindTooth:
This seems to be related to Big Sur (11.0).  Do you accept a patch for using 10.15 instead?

Ref.: https://github.com/actions/virtual-environments/issues/841#issuecomment-767200198

gcask:
fwiw, I've already re-enabled them on my fork: https://github.com/gcask/EmptyEpsilon/blob/master/.github/workflows/cicd.yml

(along with other additions).

Issue #1256, opened 2021-01-05
Various "problems" with the .po translation


muerteFR:
I notice on the last compiled (today) version that there are several texts which are not translated :

**1. RESOLVED**
"**Self Destruct Activated!**" in selfDestructEntry.cpp (Ingenierie and Ingenierie+) 
The fr.po translation does not work, but the it.po does. After watching I really don't understand why, it seems ok in fr.po : 
#: src/screenComponents/selfDestructEntry.cpp:22
msgid "Self destruct activated!"
msgstr "Autodestruction activee !"

**2. RESOLVED**  
 "**ADV. Striker**" in Science Database (classe Starfighter)
The fr.po translation doesn't just work on the ship name and works well on the description.
I just observed in ShipTemplates_Old.lua that Setclass and SetModel were reversed unlike the other lines.

**3. RESOLVED**  
 **In the GM console, the names of the ships** to be created are not translated.

**4. RESOLVED**  
 I can't find anywhere (.cpp and .lua) where translate the stations to be created **in GM: "Huge station", "small station",** ...

**5.  RESOLVED**
 #1257   I don't know where to find where translate **the systems in the hacking console:** In Target system: "Beam weapons," "Maneuvering "etc

**6. RESOLVED** 
#1258   Where to translate "**Visual Asteroide**" from the GM console in Create.

7. Where to translate in the GM console always: the indications at the top right? "**Callsign", "Shield", "Hull", "position", "Orders"** Some are already translated as "Time elapsed"

8. Still in GM, the orders at the bottom right are translated but at the top right they are not : (**"roaming", Idle", "Defend location"** etc.

If it's within my reach I will make the changes myself, but I think for the most part I couldn't help because I'm not a developer.

oznogon:
It would be easier to track and comment on these problems if they were separate issue tickets in GitHub.

> "Self destruct activated!" in selfDestructEntry.cpp (Ingenierie and Ingenierie+)

<img width="902" alt="Screen Shot 2021-01-06 at 12 34 21 AM" src="https://user-images.githubusercontent.com/19192104/103747111-ffd93800-4fb6-11eb-9c6b-b44b24291c1b.png">

Needs separate research from the other problems in this issue. This appears to be a GuiLabel, and other GuiLabels are translated, so the problem is unclear.

> "Adv. Striker" in Science Database (classe Starfighter) The fr.po translation doesn't just work on the ship name

[...]

PR #1258, opened 2021-01-06
Update visual asteroid translations in FR and DE

oznogon:
Address point 6 in #1256


muerteFR:
Sorry, It escaped me. I no longer remembered that I had added the tr () in .cpp file

PR #1259, opened 2021-01-07
Spinbox GUI numeric entry control

oznogon:
Create a gui control that combines arrow buttons (like a selector) with a textentry, for more precise control over a numeric value than a slider, but without requiring a keyboard. Respects min/max values and has a configurable interval.

Example implementation:

```
    (new GuiSpinBox(this, "SPINBOX_TEST", -100.0f, 100.0f, 0.0f, 1, 1.0f, [](string text) {
        LOG(INFO) << "Spinbox test value: " << text;
    }));
```

After the string ID `SPINBOX_TEST`, the remaining parameters are:

- minimum value
- maximum value
- starting value
- digits after the decimal point (can be zero)
- increment/decrement interval
- callback function executed on every text entry or increment/decrement action

There are additional public functions to change these values, and an `enterCallback` just like TextEntry elements that triggers when <kbd>enter</kbd>/<kbd>return</kbd> is pressed.

Some features:

- Clicking/tapping the text is the same as focusing a TextEntry element, but validates input to allow only numbers, decimal points, and negative signs
- Left and right arrow buttons decrement and increment, respectively, by the given interval
- When focused, Up and Left arrow keyboard keys also increment the value, RIght and Down keys also decrement it
- Holding down either <kbd>shift</kbd> key while incrementing or decrementing the value should do so at 10x the interval (not working on all Macs due to SFML limitations)
- Value can be returned as a float via `getValue()`
- <kbd>enter</kbd>/<kbd>return</kbd> key validates and refreshes the value

Help wanted:

- You can read the TextEntry's text directly, but the text can be invalid as a numeric value
- Some validation is present, but it's still possible to enter invalid text or have `getValue()` fail when trying to convert the text to a number
- Holding down the arrow buttons doesn't continue incrementing/decrementing
- The <kbd>-</kbd> key on US/UK Macs relies on daid/SeriousProton#63 because of SFML limitations
- Not tested in Linux or Android

https://user-images.githubusercontent.com/19192104/103879923-4f8a3300-508d-11eb-810a-f5cfb6a59333.mov


oznogon:
SFML/SFML#7 blocks entering a hyphen-minus for negative numbers on US/UK Macs. daid/SeriousProton#63 is a proposed workaround.

daid:
One potential issue I see with this current implementation is that it only updates on enter/focus loss. But, if the gui is destroyed (close is clicked on dialog) both things don't happen. And the textField doesn't do this, as it always updates the value, so that doesn't match in behavour.

oznogon:
It's not meant to match TextField in behavior. Its behavior is different on purpose. The value has to be valid and updating it on every text entry is risky.

I can validate/update on destruction.

oznogon:
Revised and implemented on the GM tweak screen Power tab. CC @daid

PR #1260, opened 2021-01-07
Asteroid fixes

StarryWisdom:
add a gm tab for asteroids, with size exposed
getSize for both visual and non visual asteroids for use in lua
export string works with size of asteroids
scripting doc makes it clear size = radius rather than diameter

### DIFF
--- a/src/screens/gm/gameMasterScreen.cpp
+++ b/src/screens/gm/gameMasterScreen.cpp
@@ -113,6 +113,10 @@ GameMasterScreen::GameMasterScreen()
             {
                 jammer_tweak_dialog->open(obj);
             }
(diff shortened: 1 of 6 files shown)


daid:
I'm kinda scared of planet sized asteroids, as well as 0 sized asteroids...

StarryWisdom:
both of those are easily possible by scripting (though 0 probably should be disabled as its too easy in sliders to select)
After me having miss judged how large people wanted to make jammers erred on the side of allowing large sizes
the CUF sandbox has one that is size 734, and a number at around size 400
on the low end we go down to 14

what range would not make you feel scared? - my next attempt would be 10-500

daid:
10-500 sounds good

PR #1267, opened 2021-01-09
locale name for ship template - db button fix

### LINKED ISSUE #1080: DB Button does not work if a different local shipname is set.
If the locale name of a ship is different from the general name, the DB button on science is not functioning, as no DB entry is found.

#898 has been proposed to fix this (at the price of reverting the old behaviour of showing the general names in the DB), however it conflicts now, and I am not familiar with the new DB structure in order to fix this.

To reproduce this issue without a localized po in place: Spawn a "Cruiser"*, scan it and try to use the DB button.

*Not the player ship, so spawn it in GM screen.

aBlueShadow:
This fixes #1080 ,
implementing @ŧdelc 's patch.


daid:
Wouldn't this break scenarios that check for the typename?

PR #1277, opened 2021-01-16
Allow configurable radar transparency

oznogon:
Add `setBackgroundAlpha()` to set a GuiRadarView background's
transparency.

No currently implemented screens take advantage of this, but this would ease implementation of new screen types, like this cockpit radar overlaid on a Viewport3D with a radar background alpha value of 192:

```
radar->setBackgroundAlpha(192);
```

Screen Shot 2021-01-15 at 10 00 27 PM


daid:
Might also be nice for the corner radar on the main screen, just 90-80% or something.

PR #1278, opened 2021-01-16
Add cockpit view 3D screen

oznogon:
Add a new single-player screen with a 3D viewport background, smaller radar, and controls overlaid on top of the viewport.

Screen Shot 2021-02-04 at 12 31 05 PM

Like Single Player, the view includes controls and information for weapons, navigation, docking, and comms, but not for engineering, scanning, hacking, waypoint placement, shield or beam frequency calibration, beam system targeting, or probe management.

The viewport can be rotated forward, left, rear, or right via keybindings (default <kbd>I</kbd> <kbd>J</kbd> <kbd>K</kbd> <kbd>L</kbd>, respectively), or locked onto a selected target via the new target (TGT) button or keybinding (default <kbd>O</kbd>), similar to a main screen. Enabling the target-locked camera also displays a list of target info similar to the Science station.

To target targetable objects, click/tap the radar or use Weapons keybindings. To navigate, manipulate the steering wheel rotation dial, use Helms keybindings, or use new keybindings that rotate the target heading on the steering wheel (default <kbd>Q</kbd> <kbd>E</kbd>).

- Adds "Cockpit View" to the 1 Player/Extras screen list when FEATURE_3D_RENDERING is enabled. This view uses existing control bindings for Helms/Weapons.
- Adds a new navigational control that uses a rotation dial to set and communicate a target heading, with new keybindings.
- Adds new keybindings for view manipulation specific to Cockpit View, in order to avoid helm/weapons keybinding conflicts.

Potential areas for improvement:

- Adding keybinds or UI controls for shield and beam frequency calibration.
- Adding conditional UI and/or keybinds for beam system targeting, appearing when a beam-targetable object is in target-lock view.
- Improving target selection and targeting keybind behavior, or allowing for target filtering on radar (ie. showing/selecting only enemies, only friendlies, only stations).
- Adding a keybind or UI control to toggle a larger short-range radar view, or a long-range radar view, similar to the main screen's radar view options.
- Adding a keybind or UI control to drop WP1 at the ship's current position, in lieu of a more advanced waypoint management system, so that single-pilot ships can call for supply drops or reinforcements from stations to their current position.
- Adding a keybind or UI control to toggle hiding the controls.

Scanning, hacking, and probe management seem out of scope for this view. This also needs testing with scenarios that make heavy use of custom functions to ensure they're still visible and usable, and with joystick input to ensure turning left/right via joystick still works with the changes made to helm controls.


aBlueShadow:
I really like it!

Just one thing though. Imo it feels off that the TURN_LEFT/TURN_RIGHT hotkeys control the steering wheel instead of just rotating the ship.
That is a very different behavior from the other stations, and especially from a cockpit type console I would expect a very direct control. The Joystick controls are already leaving the steering wheel alone, which I would expect.
So having a direct control without changing the steering wheel would also be more consistent.

oznogon:
Keyboard turn behavior beyond the UI is identical to Helms; modify `commandTargetRotation` by 5 degrees. The difference is it's reflected in a UI element that Helms doesn't have.

Helms:

```
    if (key.category == "HELMS" && my_spaceship)
    {
        if (key.hotkey == "TURN_LEFT")
            my_spaceship->commandTargetRotation(my_spaceship->getRotation() - 5.0f);
        else if (key.hotkey == "TURN_RIGHT")
            my_spaceship->commandTargetRotation(my_spaceship->getRotation() + 5.0f);
    }
```

Cockpit view:

```
        if (key.category == "HELMS")
        {
            if (key.hotkey == "TURN_LEFT")
            {
                target_rotation -= 5.0f;
                my_spaceship->commandTargetRotation(target_rotation);
                steering_wheel->setValue(target_rotation - view_rotation);
            }
            else if (key.hotkey == "TURN_RIGHT")
            {
                target_rotation += 5.0f;
                my_spaceship->commandTargetRotation(target_rotation);
                steering_wheel->setValue(target_rotation - view_rotation);
            }
        }
```

If anything the existing Helms behavior is inconsistent, because holding down left and right hotkeys repeatedly interrupts the 5-degree `commandTargetRotation` on each input — it only appears to turn the ship smoothly while you hold down the key, but visualizing that behavior via the steering wheel exposes it as jittery and inconsistent, especially on clients.

oznogon:
> The Joystick controls are already leaving the steering wheel alone, which I would expect.

Also, I've fixed this, because the wheel should indicate the target ship heading. Joystick input also updates the target heading differently from keyboard input.

oznogon:
Current view from a Piranha at 4:3 and 16:9

Screen Shot 2021-01-17 at 12 22 26 PM

Screen Shot 2021-01-17 at 12 22 06 PM

aBlueShadow:
`Keyboard turn behavior beyond the UI is identical to Helms; modify commandTargetRotation by 5 degrees. The difference is it's reflected in a UI element that Helms doesn't have.`
hm.. the rotation is not identical at all when I tested it. The problem is that the steering wheel will soon get way ahead of the actual rotation.
So when you stop pressing the key on helms, the ship will stop turning, but not on the cockpitview.

oznogon:
Edited my comment with additional details about the differences in observed behavior and causes. I reassert that the behavior of pressing the key is identical; the difference is in how repeated input is acted upon.

To clarify, hotkeys don't fire events by detecting that the key is held down. It's firing constantly repeating, separate keypress events while the key is down. That's a known limitation of the EE hotkey implementation.

Again, with Helms-style input and the steering wheel UI only visualizing what Helms is doing under the hood:

```
        if (key.category == "HELMS")
        {
            if (key.hotkey == "TURN_LEFT")
            {
                LOG(INFO) << "Unique TURN_LEFT keypress event";
                target_rotation = my_spaceship->getRotation() - 5.0f;
                my_spaceship->commandTargetRotation(target_rotation);
                steering_wheel->setValue(target_rotation - view_rotation);
            }
            else if (key.hotkey == "TURN_RIGHT")
            {
                LOG(INFO) << "Unique TURN_RIGHT keypress event";
                target_rotation = my_spaceship->getRotation() + 5.0f;
                my_spaceship->commandTargetRotation(target_rotation);
                steering_wheel->setValue(target_rotation - view_rotation);
            }
```

https://user-images.githubusercontent.com/19192104/104855748-63226e80-58c3-11eb-8c1c-837cc34d77a8.mov

This can be hacked around by handling input without using the hotkey system, but would prevent the keys from being remappable.

Either I implement it to match the Helms behavior bug-for-bug, resulting in internally inconsistent behavior (how much rotation will I get this keypress? why?)

or I implement it to reflect the command behavior, resulting in different behavior from Helms (every input is consistent, but it isn't broken in the same manner)

or I implement a different keyboard input system just for Cockpit View to behave more like joystick input (keys aren't remappable)

aBlueShadow:
Maybe we should do a poll for that, but I would expect that people who are using hotkeys for turning, will rotate by keeping them pressed(like in basically any other game), rather than pressing them multiple times. So I think the main concern won't be inconsitency per key press (you could call it "inertia"), but rather if those constant command interruption will cause problematic overhead

> or I implement it to reflect the command behavior, resulting in different behavior from Helms (every input is consistent, but it "looks" like it behaves the same as Helms)
I am not exactly sure what the implications of this would be

oznogon:
> > or I implement it to reflect the command behavior, resulting in different behavior from Helms (every input is consistent, but it "looks" like it behaves the same as Helms)

> I am not exactly sure what the implications of this would be

It's what's already implemented in this PR, the behavior you don't like.

oznogon:
Ideally keyboard rotation on Helms would work like this, instead of setting the target rotation:

```
            if (key.hotkey == "ROTATE_LEFT")
            {
                my_spaceship->commandTurnSpeed(-1);
            }
            else if (key.hotkey == "ROTATE_RIGHT")
            {
                my_spaceship->commandTurnSpeed(1);
            }
```

This is the same as what the joystick is doing, except the joystick is always reporting its axis value, so when the axis is 0, the turn speed is 0.

Because the hotkey system only reports key-down events, and not whether the key is held down or when it's released, I can't reset `commandTurnSpeed` to 0 when the key is released. So the ship starts turning when the key is pressed and continues turning after the key is released.

The workaround would be to implement SFML input, ie. `isKeyPressed`, here. But I need to figure out a way to get keycodes from the hotkey system to retain remappability.

oznogon:
Trying a different tack. I added a function to `HotkeyConfig` to get the SFML Key associated with a hotkey so they can be used outside of `onHotkey`, then implemented TURN_LEFT and _RIGHT in `onDraw` with `isKeyPressed` instead.

The behavior is now much closer to joystick input, using `commandTurnSpeed()` and updating it only if the turn speed is actually changing. I don't like adding more logic in `onDraw`, but it didn't respond when I added and put it in `onKey`, but I'm not sure whether I botched it.

Now:

- Holding down TURN_LEFT and TURN_RIGHT keys rotates the ship directly.
- Holding TURN_LEFT and _RIGHT to rotate is smoother, both on the server and client.
- TURN_LEFT and _RIGHT now allows for finer control than even the Helms implementation, since the rotation is based on turn speed and no longer results in a minimum of 5 degrees turn from the ship's current rotation.
- Rotating the target heading indirectly — "turning the wheel", my previous implementation — is now a separate keybind `TURN_WHEEL_LEFT` and `_RIGHT`.

My main concern is whether having this logic in onDraw will interfere with joystick rotation input. This is where I'd really prefer to handle events or manage state in something limited to keyboard handling.

oznogon:
> > I don't like adding more logic in onDraw
> 
> Gui's elements have an `onUpdate()` as well, wouldn't that be more suited for logic?

Yeah, but my point was more that I'd expect key down/pressed/up to be handled in `onKey` or when the keyboard is in use, rather than having to handle it every draw or update or tick, etc.

Specifically I'm concerned about how constantly checking and possibly resetting `commandTurnSpeed` to 0 might conflict with joystick input, which has its own axis-related event. (I don't have a joystick, so I can't test that.)

PR #1279, opened 2021-01-16
Enable translation of gm create ship/station buttons

StarryWisdom:
partial fix for #1256 (point 3 and 4)

### DIFF
--- a/src/screens/gm/objectCreationView.cpp
+++ b/src/screens/gm/objectCreationView.cpp
@@ -42,10 +42,14 @@ GuiObjectCreationView::GuiObjectCreationView(GuiContainer* owner)
     std::sort(template_names.begin(), template_names.end());
     for(string template_name : template_names)
     {
-        (new GuiButton(box, "CREATE_STATION_" + template_name, template_name, [this, template_name]() {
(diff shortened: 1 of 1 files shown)


src/screens/gm/objectCreationView.cpp
```diff
@@ -42,10 +42,14 @@ GuiObjectCreationView::GuiObjectCreationView(GuiContainer* owner)
     std::sort(template_names.begin(), template_names.end());
     for(string template_name : template_names)
     {
-        (new GuiButton(box, "CREATE_STATION_" + template_name, template_name, [this, template_name]() {
-            setCreateScript("SpaceStation():setRotation(random(0, 360)):setFactionId(" + string(faction_selector->getSelectionIndex()) + "):setTemplate(\"" + template_name + "\")");
-        }))->setTextSize(20)->setPosition(-350, y, ATopRight)->setSize(300, 30);
-        y += 30;
+        auto stationTemplate=ShipTemplate::getTemplate(template_name);
+        if (stationTemplate)
```
daid, line 46: This check looks odd, why wouldn't a template exist if it is in the list of templates?

muerteFR:
I would like to know if this PR is valid to be merged?  Thanks

PR #1282, opened 2021-01-17
Optional condensed GuiShieldsEnable with icon

oznogon:
Add a `showIconOnly()` function that adds `gui/icons/shields` as the button icon and limits the button text to only the frequency if in use or `ON`/`OFF` if not.

Allows for `GuiShieldsEnableButton`s as small as 130, 40. No change in default behavior; `icon_only` defaults to `false`.

Rearranges some localized strings, but doesn't change their meaning.

Default behavior / `(new GuiShieldsEnableButton(this, "SHIELDS_ENABLE"))->showIconOnly(false)`:

Screen Shot 2021-01-16 at 9 21 19 PM

Screen Shot 2021-01-16 at 9 19 49 PM

`(new GuiShieldsEnableButton(this, "SHIELDS_ENABLE"))->showIconOnly(true)`:

Screen Shot 2021-01-16 at 9 21 06 PM

Screen Shot 2021-01-16 at 9 22 00 PM

### DIFF
--- a/src/screenComponents/shieldsEnableButton.cpp
+++ b/src/screenComponents/shieldsEnableButton.cpp
@@ -11,17 +11,26 @@
 #include <string>
 
 GuiShieldsEnableButton::GuiShieldsEnableButton(GuiContainer* owner, string id)
-: GuiElement(owner, id)
+: GuiElement(owner, id),
+  icon_only(false)
 {
+    // Recommended minimum button size with default text size: 130, 40
+
+    // Shield toggle button.
     button = new GuiToggleButton(this, id + "_BUTTON", "Shields: ON", [](bool value) {
         if (my_spaceship)
+        {
             my_spaceship->commandSetShields(!my_spaceship->shields_active);
+        }
     });
     button->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
+
+    // Progress bar during calibration.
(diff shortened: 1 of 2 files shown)


src/screenComponents/shieldsEnableButton.cpp
```diff
@@ -31,20 +40,46 @@ void GuiShieldsEnableButton::onDraw(sf::RenderTarget& window)
...
             if (gameGlobalInfo->use_beam_shield_frequencies)
-                button->setText(tr("{frequency} Shields: {status}").format({{"frequency", frequencyToString(my_spaceship->shield_frequency)}, {"status", shield_status}}));
+            {
+                if (icon_only)
+                {
+                    button->setText(shield_frequency);
+                }
+                else
+                {
+                    button->setText(string("{frequency} {status}").format({{"frequency", shield_frequency}, {"status", shield_status}}));
```
daid, line 69: This seems to have changed from `"{frequency} Shields: {status}"` to `"{frequency} {status}"`?

oznogon: Yes. The logic of whether to add "Shields: " to the string is in L56.

daid: Ah, I missed that. That still looks odd then. In case a translation needs "{frequency} {status} shields" that is now no longer possible.

oznogon: I don't understand the problem, so I'll just make it more verbose so it changes less

PR #1287, opened 2021-01-23
Revise Relay shared radar checks

oznogon:
On Relay radar, reveal the short-range radar range for friendlies,
instead of the hardcoded 5U.

Allow a ship's own short-range radar range to still be visible on
Relay even if the ship belongs to a faction that is hostile to
itself.

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -246,9 +246,10 @@ void GuiRadarView::drawNoneFriendlyBlockedAreas(sf::RenderTarget& window)
                 // ... or is a scan probe from our ship ...
                 || (P<ScanProbe>(obj) && P<ScanProbe>(obj)->owner_id == my_spaceship->getMultiplayerId()))
             {
-                float r = 5000.0 * scale;
+                float r = 5000.0f * scale;
 
-                // Unmask short-range radar around it
+                // Unmask the short-range radar radius around friendly PlayerShips.
+                // (We have to specify our own ship because we can belong to a faction hostile to us!)
                 if (P<PlayerSpaceship>(obj) && (obj->isFriendly(my_spaceship) || obj == my_spaceship))
                 {
                     r = P<PlayerSpaceship>(obj)->getShortRangeRadarRange() * scale;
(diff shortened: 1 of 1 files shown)


src/screenComponents/radarView.cpp
```diff
@@ -232,25 +232,30 @@ void GuiRadarView::drawBackground(sf::RenderTarget& window)
...
-            {
-                circle.setPosition(worldToScreen(obj->getPosition()));
-                window.draw(circle);
-            }
-            P<ScanProbe> sp = obj;
-            if (sp && sp->owner_id == my_spaceship->getMultiplayerId())
+            // If an object is a ship or station that is friendly, or is our ship...
+            if (((P<SpaceShip>(obj) || P<SpaceStation>(obj)) && (obj->isFriendly(my_spaceship) || obj == my_spaceship))
+                // ... or is a scan probe from our ship ...
+                || (P<ScanProbe>(obj) && P<ScanProbe>(obj)->owner_id == my_spaceship->getMultiplayerId()))
```
oznogon, line 247: Consolidates the two if statements into one and removes the otherwise unused P<ScanProbe> assignment.

src/screenComponents/radarView.cpp
```diff
@@ -232,25 +232,30 @@ void GuiRadarView::drawBackground(sf::RenderTarget& window)
...
             {
+                float r = 5000.0 * scale;
+
+                // Unmask short-range radar around it
+                if (P<PlayerSpaceship>(obj) && (obj->isFriendly(my_spaceship) || obj == my_spaceship))
```
oznogon, line 252: Instead of the fixed 5U radius, which remains the default, check for a short-range radar range on friendly PlayerSpaceships and use that instead.

src/screenComponents/radarView.cpp
```diff
@@ -591,39 +597,58 @@ void GuiRadarView::drawObjects(sf::RenderTarget& window_normal, sf::RenderTarget
...
+        for(P<SpaceObject> obj : space_object_list)
         {
+            // If the object can't hide inside a nebula, show it.
             if (!obj->canHideInNebula())
+            {
                 visible_objects.insert(*obj);
+            }
 
-            if ((!P<SpaceShip>(obj) && !P<SpaceStation>(obj)) || !obj->isFriendly(my_spaceship))
+            // If an object is a ship or station that is friendly, or is our ship...
```
oznogon, line 620: Same changes here; consolidate the if statements and set the radius for querying and revealing objects to short-range radar radius on PlayerSpaceships.

src/screenComponents/radarView.cpp
oznogon: After re-reviewing the Discord discussion in November around this, I'll move short-range radar range to ShipTemplateBasedObject so it can be customized on any ship type, not just PlayerSpaceships.

PR #1288, opened 2021-01-24
Add remappable voice chat keys

### LINKED ISSUE #930: Allow remapping voice chat keys
Voice chat keys are [hardcoded](https://github.com/daid/EmptyEpsilon/blob/f998fea2e720c850c5515c83addb6a10fc06c9e7/src/main.cpp#L296-L297) to <kbd>~</kbd> and <kbd>Backspace</kbd>. Allow them to be remapped as hotkeys, since not all keyboards have tilde/backtick or backspace keys.

oznogon:
Add HOTKEY.BASIC.VOICE_CHAT_ALL and HOTKEY.BASIC.VOICE_CHAT_SHIP
for broadcasting voice to the server and player's ship, respectively.
Set default hotkey bindings to their current hardcoded keys of
Backspace and Tilde, respectively.

The hotkey menu in Options doesn't have access to the
NetworkAudioRecorder object in `main()` where the keys are bound,
so they're put in the BASIC hotkey category, which can't be rebound
via the hotkeys menu. However, the keys can be rebound in
`options.ini`.

Resolves #930.


daid:
Good enough for now I guess.

PR #1291, opened 2021-01-25
Make missiles scriptable in LUA

Tsht:
Allows to create missiles in script.
Added setters and getters to specific missile functions (lifetime, target, size)
Missiles can be created using a derived class.

Category modifier conversion put in MissileWeaponData class


src/missileWeaponData.cpp
```diff
@@ -37,4 +37,30 @@ string getMissileSizeString(EMissileSizes size)
...
+        case MS_Large:
+            return 2.0;
+        default:
+            return 1.0;
+    }
+}
+
+const EMissileSizes MissileWeaponData::convertCategoryModifierToSize(float size)
+{
+    if(0.5 == size)
```
daid, line 57: Might want to do something like `if (std::abs(size - 0.5) < 0.1)` to allow for floating point inaccuracies.

Tsht: Makes sense even if 0.1 might be a little overkill :) Don't see it upgrading to 0.1 increments though...

PR #1296, opened 2021-01-28
Removed glBegin/End on Beam effects and explosions

gcask:
Removing fixed functions - starting with glBegin/glEnd in the beam fx.

Keeping in mind that the end game is to remove the fixed pipeline entirely, that means a lot of the `gl_*` variables used in the shaders won't be available.
To that end, added versioned shaders (`gl_ModelViewProjectionMatrix` is not a thing in es2 shaders, but it is in gl2), as to not conflict with existing ones.
And started a few utilities

Moved over to the shader pipeline in the simplest way possible.

### DIFF
--- a/resources/shaders/basic.frag
+++ b/resources/shaders/basic.frag
@@ -1,0 +1,9 @@
+#version 120
+uniform vec4 color;
+uniform sampler2D textureMap;
+varying vec2 fragtexcoords;
+void main()
+{
+	gl_FragColor = texture2D(textureMap, fragtexcoords.st) * color;
+	gl_FragColor.rgb *= color.a;
+}
--- a/resources/shaders/basic.frag
+++ b/resources/shaders/basic.frag
@@ -1,7 +1,10 @@
 #version 120
+
 uniform vec4 color;
 uniform sampler2D textureMap;
+
 varying vec2 fragtexcoords;
+
 void main()
 {
 	gl_FragColor = texture2D(textureMap, fragtexcoords.st) * color;
--- a/resources/shaders/basic.frag
+++ b/resources/shaders/basic.frag
@@ -7,6 +7,6 @@ varying vec2 fragtexcoords;
 
 void main()
 {
-	gl_FragColor = texture2D(textureMap, fragtexcoords.st) * color;
-	gl_FragColor.rgb *= color.a;
+    gl_FragColor = texture2D(textureMap, fragtexcoords.st) * color;
+    gl_FragColor.rgb *= color.a;
 }
--- a/src/glObjects.cpp
+++ b/src/glObjects.cpp
@@ -1,0 +1,33 @@
+#include "glObjects.h"
+#if FEATURE_3D_RENDERING
+#include <GL/glew.h>
+#include <type_traits>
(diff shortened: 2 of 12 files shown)


gcask:
If that's to your (and the CIs!) liking, then it's rinse and repeat for the rest, deprecate the old shaders and voila.

resources/shaders/basic.frag
```diff
@@ -0,0 +1,9 @@
+#version 120
+uniform vec4 color;
```
daid, line 2: I'm going to be annoying, but I like my whitespace lines. So after `#version` and before `void main()` as it gives a clear visual clue what things belong together.

src/glObjects.cpp
```diff
@@ -0,0 +1,33 @@
+#include "glObjects.h"
+#if FEATURE_3D_RENDERING
+#include <GL/glew.h>
+#include <type_traits>
+static_assert(std::is_same<uint32_t, GLuint>::value, "GLuint and uint32_t are *NOT* the same: troubles!");
```
daid, line 5: More whitespace lines, before static_assert and after it.

src/glObjects.h
```diff
@@ -0,0 +1,49 @@
+#ifndef EMPTYEPSILON_GLOBJECTS_H
+#define EMPTYEPSILON_GLOBJECTS_H
+#include "featureDefs.h"
+#if FEATURE_3D_RENDERING
+#include <array>
+#include <cstdint>
+namespace gl
```
daid, line 7: whitespace before the namespace... well, you'll start to get it I guess :)

daid (commented):
Overall looks good, just some general style comments.

General comment, I think all other code is indented with two spaces, while your new files are intended with tabs.

gcask:
Checked a few existing files - seems to be 4 spaces, so lined up with that.

daid:
> Checked a few existing files - seems to be 4 spaces, so lined up with that.

The diff on github looks still like tabs...

gcask:
https://raw.githubusercontent.com/daid/EmptyEpsilon/e56a078087fa0c5c41e21d75542b585693a6c530/resources/shaders/billboard.frag

That's 4 spaces (you can select them!)

gcask:
ah looks like I missed the cpp/h, be there in a sec ;)

gcask:
Alright now that should do it.

daid:
I think there is still a ton of room for gl code improvements. But small steps and a good start :)

For example the beam effects that update every frame depending on the camera position could be handled in a vertex shader. And it could all be converted to VBOs. But one step at a time :)

Also, if you are up for an extra challenge, for a whole bunch of 3D models I should have bumpmaps/normal maps that are currently not in the repo nor used by the shaders.

gcask:
Yup, that's the plan.

Currently working my way up removing the old shaders, then the fixed pipeline.

Then we can talk improvements, top of my head:
-Particles too should just really be points (and sample texture based on `gl_PointCoord` in the frag shader). They use a lot of fillrate for very little data.
- the whole rendering pipeline to be revisited, currently it switches states too much, things need to be more persistent.
- I think ebo should make a comeback for models (again, less data).
- the starbox should really just be a cube texture.

I'm working my way through it, while trying to stick to the one PR/one change (you're about to have nebulas, explosions, wormhole, blackhole).

PR #1297, opened 2021-01-29
More factions

Xansta:
Sixteen factions: all enemies to each other. Designed for more PvP scenarios


daid:
I'm going to reject this one. I understand the usecase, but I rather not see the extra clutter for any other game that does not need these extra factions.

PR #1299, opened 2021-01-29
Roam near position if no target position is set

### LINKED ISSUE #355: Bug : roaming ordering ships go to 0,0
When ships have order to roaming, they first go to 0,0 and, then, their destination is random (I think). I think It will be better to set random destination from the beginning.

oznogon:
If no order_target_position is set, a ship's orders are set to AI_Roaming, and
no targets are within 50U, ships following roaming orders will move towards 0,0
because that's the default value if no target position is unassigned.

When getting orders to roam, instead get the ship's current position at that time,
and then continuously move to random points within a 30U radius of that position.

This significantly changes the behavior of roaming ships more than 50U away from an
enemy, which will no longer fly to coordinates 0,0. Scenarios that rely on this behavior might be affected. To reproduce an equivalent to the old behavior, order
`flyTowards(0,0)` before roaming.

Fixes #355.

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -332,7 +332,7 @@ void ShipAI::runOrders()
             // 30U of this ship's current position.
             else
             {
-                roamToNewRandomLocation(30000.0f);
+                roamToNewRandomLocation(owner->getPosition(), 30000.0f);
             }
         }
         // - Retreat to a station, or roam to find a station, if out of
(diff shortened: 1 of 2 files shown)


daid:
Problem you might get now, is that ships just wander off towards infinity (and beyond) not sure what the best fix is...

oznogon:
@daid What's the intended behavior of roaming? It looked like it was meant to roam around the entire map, but if the point is to stay within a 30U radius of a given point, it would be straightforward enough to center then random coordinates around whatever the ship's current position is when it enters roaming.

daid:
>  It looked like it was meant to roam around the entire map

Yes, that is sort of the intention, only issue is, what is the "entire map". I kinda randomly picked this roaming size.

PR #1303, opened 2021-01-30
Billboard quads pass

gcask:
Converting a few quads, mostly billboards.

If it looks like a bunch of copy pasta that's... because it is sadly.
But I'm  fighting fire with fire, the actual cleanup and benefits are reaped at a later stage.


gcask:
On hold, needs #1308 for headless support.

daid:
Yeah, it looks very copy pasta, but that's not your fault :)

Looking at it, I think we could wrap the sf::Shader into a class that will hold the attribute values as well, as the shader+attribute indices seems to be a very common setup.

PR #1307, opened 2021-01-30
Script ScanProbe speed

oznogon:
Add `ScanProbe:getSpeed()` and `ScanProbe:setSpeed(float)` to set a given ScanProbe's speed. The default value of 1000 = 1U/second.

This also refactors probe movement, since speed values of >30000 would cause the probe to "flap" around the target coordinates, since its movement per tick was too fast for the probe's radius to make contact with the target.

### DIFF
--- a/src/spaceObjects/scanProbe.cpp
+++ b/src/spaceObjects/scanProbe.cpp
@@ -131,13 +131,31 @@ void ScanProbe::update(float delta)
 
     // The probe moves in a straight line to its destination, independent of
     // physics and at a fixed rate of speed.
-    if ((target_position - getPosition()) > getRadius())
+    sf::Vector2f diff = target_position - getPosition();
+    float movement = delta * probe_speed;
+    float distance = sf::length(diff);
+
+    // If the probe's outer radius hasn't reached the target position ...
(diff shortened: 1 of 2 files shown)


src/spaceObjects/scanProbe.cpp
```diff
@@ -40,7 +49,8 @@ REGISTER_SCRIPT_SUBCLASS_NO_CREATE(ScanProbe, SpaceObject)
 
 REGISTER_MULTIPLAYER_CLASS(ScanProbe, "ScanProbe");
 ScanProbe::ScanProbe()
-: SpaceObject(100, "ScanProbe")
+: SpaceObject(100, "ScanProbe"),
+  probe_speed(1000.0f)
```
daid, line 53: probe_speed better be replicated now as well, or clients will simulate that wrong and probes with a non-default speed will look janky.

oznogon:
Linux CI failure is unrelated:

```
E: The repository 'https://apt.postgresql.org/pub/repos/apt bionic-pgdg Release' no longer has a Release file.
Error: Process completed with exit code 100.
```

Issue #1305, opened 2021-01-30
Crash on ExplosionEffect


oznogon:
Ruled out some issues in #1304 but hit a similar crash, with the shaders in the expected location, on a headless run, when a Basic scenario had an asteroid collision on launch that created an ExplosionEffect. This also affected Windows, where #1304 did not.

Version is built from the master branch of both EmptyEpsilon and SeriousProton. This isn't reproduced in the 2020.11.23 release.

macOS crash report:

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [44617]

VM Regions Near 0:
--> 
    __TEXT                 000000010b425000-000000010bcbc000 [ 8796K] r-x/rwx SM=COW  /Users/USER/Documents/*/EmptyEpsilon.app/Contents/MacOS/EmptyEpsilon

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   ???                           	000000000000000000 0 + 0
1                                 	0x000000010b9729ac ExplosionEffect::ExplosionEffect() + 1212
2                                 	0x000000010b9852d2 Asteroid::collide(Collisionable*, float) + 674
3                                 	0x000000010baa2f32 CollisionManager::handleCollisions(float) + 2498
4                                 	0x000000010bb449a9 Engine::runMainLoop() + 681
5                                 	0x000000010b42fc00 main + 31040
6   libdyld.dylib                 	0x00007fff6c0af3d5 start + 1
```

Windows EmptyEpsilon.RPT:

```
-------------------

Error occurred on Friday, January 29, 2021 at 22:22:24.

EmptyEpsilon.exe caused an Access Violation at location 00000000 DEP violation at location 00000000.

Registers:
eax=00000000 ebx=05152148 ecx=02105d38 edx=02210000 esi=0157e830 edi=005ebc9c
eip=00000000 esp=0157e73c ebp=0157e848 iopl=0         nv up ei pl zr na po nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00010246

AddrPC   Params
00000000 00000001 00877C88 3F8284E7
[...]

PR #1308, opened 2021-01-30
#1305 : Detect overall GL availability

gcask:
For #1305 : Skip shader initialization if OpenGL is not available.

Detecting this via glew variables, since using sf::Shader::isAvailable() would force GL initialization, something we probably want to try and avoid in headless contexts.


daid:
I've also seen `sf::Shader::isAvailable` return true while the GPU didn't support shaders, so I wouldn't depend too hard on it :)

PR #1309, opened 2021-01-31
lua functions to customize artifact radar traces

aBlueShadow:
This adds 3 functions for artifacts:

setRadarTraceIcon()
setRadarTraceScale()
setRadarTraceColor()

Originally made with asteroid mining in mind, so you could make artifacts that look like asteroids, but are either big "motherlodes" that should not explode on contact, or small chunks that could be cut from the bigger ones, and then be collected (maybe for the latter, one might still want to make it look slightly different from normal asteroids). 
But there would also be a lot of other applications for that.
artifacts

### DIFF
--- a/src/spaceObjects/artifact.h
+++ b/src/spaceObjects/artifact.h
@@ -11,6 +11,11 @@ private:
     float artifact_spin=0.0;
     bool allow_pickup;
     ScriptSimpleCallback on_pickup_callback;
+
+    string radar_trace_icon="RadarBlip.png";
+    float radar_trace_scale=0;
+    sf::Color radar_trace_color = sf::Color(255, 255, 255);
+
 public:
     Artifact();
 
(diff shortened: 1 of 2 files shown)


src/spaceObjects/artifact.h
```diff
@@ -11,6 +11,11 @@ class Artifact : public SpaceObject, public Updatable
     float artifact_spin=0.0;
     bool allow_pickup;
     ScriptSimpleCallback on_pickup_callback;
+
+    string radar_trace_icon="RadarBlip.png";
+    float radar_trace_scale=0;
```
daid, line 16: default of zero?

aBlueShadow: Yep, I used the value of zero to set the standard scaling behavior. See artifact.cpp line 84

daid: ah, missed that :)

PR #1310, opened 2021-02-02
Particle instancing

gcask:
Using instancing technique for the particle system (retaining gl2/es2 compatibility, so poor person's instancing :) )

The particle shader now supports instancing 64 particles per draw call.

The current system uses 96B per particles (RGB color, size, 4 vertices, 4 texture coords). Because it's a fixed function pipeline, it's hard to estimate how many draw calls are actually involved (it's likely the API batches somewhat).

However on my machine, testing at large scale (Battlefield - Huge) did suffer greatly from the particles (just removing the draws and it was a lot smoother).

Using the instancing trick gets us there by reducing both the number of draw calls (one call for 64 particles), as well as the memory traffic: Each particle now eats only 32B (RGB color, size, 4 vertex IDs, 1 position), essentially fitting 3x the amount of particle in the same space.

This translates into a nice framerate boost, and less memory used.

Additionally, plugged some extra processing time: the particles array is now kept partitioned, so finding a free particle is an (amortized) O(1) operation (amortized because allocations may happen).
Upon updating, we always ensure live particles are kept at the front of the array.
This should help especially after a spike in particles, when things die down and the particle count gets orders of magnitude lower.

Similar techniques could be applied to asteroid fields.


oznogon:
Battlefield, Huge variant, on macOS 10.14 with SP on master. Missing/glitchy missile and engine trail particles:

https://user-images.githubusercontent.com/19192104/106591994-cdfab900-6503-11eb-8378-ae0ba3e69ba0.mov

What's the expected behavior?

gcask:
Well the expected behavior is obviously unchanged behavior.

Couple follow up questions:
 * Are both videos taken on the Huge battlefield variant? (The radar looks a lot less busy on the second video...)
 * Are you missing trails everywhere or only in the huge battlefield?

gcask:
Mmmh...

I think I understand the why now. I might've have forgotten about the projection and model view matrices taking up uniform space. (oops)

gcask:
I've tested it in my Debian VM (which has a crappy MESA driver) and it works good there with the latest update of the PR.

If you still run into an issue, try in Debug and look for shader compiler error (they should pop up in the console while transitioning to the ship selection screen).

gcask:
This last one should to it 🤞 
I fixed the triangle winding, and changed all the shaders to be in version 120 (I can deal with fixing them up to es2 later).

gcask:
@oznogon Everything was working for you with the latest, right?

oznogon:
@gcask 👍

src/glObjects.cpp
```diff
@@ -43,6 +54,22 @@ namespace gl
             glDisableVertexAttribArray(attrib);
     }
 
+    ScopedTexture::ScopedTexture(uint32_t target, uint32_t texture)
+        :target{ target }, texture{ texture }
+    {
+        // ES2 only supports 2D textures and cubemaps.
+        glGetIntegerv(target == GL_TEXTURE_2D ? GL_TEXTURE_BINDING_2D : GL_TEXTURE_BINDING_CUBE_MAP, &previously_bound);
```
daid, line 61: Most openGL guides say to avoid `glGet*` as they can be slow, no idea if this is the case for this one.

gcask: That's not wrong, this is a safety net - I'm trying to minimize side effects.
On the same vein, you don't want to `Bind(GL_NONE)` all the time to "keep things clean" - it's a safety net as well.

So  far I've only seen gains  or no change (Oz on mac only reported a "more stable" fps) which means those few calls are not on the hot path (yet. When they are, that means the rendering is in an awesome shape 😁 )

Ideally you know your flow and you only rebind when necessary.
Which is something I want to get to... But first things first :)

Once all the 3D render code goes through `gl::Scoped*`, it becomes trivial to do client-side tracking.

PR #1313, opened 2021-02-02
Star background as a cubemap

gcask:
Moved the starbox to a cubemap.
Added a RAII object for textures.

This takes care of several fixed API functions.

### DIFF
--- a/resources/shaders/starbox.vert
+++ b/resources/shaders/starbox.vert
@@ -1,0 +1,22 @@
+#version 100
+
+// Program inputs
+uniform mat4 projection;
+uniform mat4 model_view;
+uniform float scale;
+
+// Per-vertex inputs
+attribute vec3 position;
+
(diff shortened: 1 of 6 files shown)


resources/shaders/starbox.vert
```diff
@@ -0,0 +1,22 @@
...
+// Per-vertex outputs
+varying vec3 texcoords;
+
+void main()
+{
+    // For texcoords, undo flipping the Z axis and rotation around X axis
+    // from code.
+    texcoords = vec3(position.x, position.z, -position.y);
+
+    // We get rid of the translation component, so the box gets always centered around the camera.
```
daid, line 20: A different way to rotate without translation is to set the 4th vec4 value to 0.0 instead of 1.0. Not sure if it will make a difference. Won't make a difference in this case, but might be nice to know for other things.

PR #1325, opened 2021-02-06
Allow scripting Collisionable physics on SpaceObjects

oznogon:
Expose `Collisionable` functions `setCollisionPhysics`, `setCollisionRadius`, and `setCollisionFriction` in `SpaceObject`.


daid:
I don't like this at the SpaceObject level, changing certain objects physics properties can get things really screwy. Also it doesn't account for network replication, so only the server knows the new physics properties.

oznogon:
I tested it on clients, which behaved no differently than with existing objects with the same physics properties. 🤷  Will find a better way

PR #1327, opened 2021-02-07
Use radar ranges instead of hardcoded values for AI targeting and firing

oznogon:
Use `getShort` and `getLongRangeRadarRange()` values when calculating AI actions, instead of hardcoded values. Some values are now modified based on those ranges by a fixed amount (ie. short-range + 3U instead of 8U). 

Assuming a ship's defaults aren't modified from 5U short/25U long, only these ranges behaviors should change:

- Roaming will now seek the best targets within long-range radar range, rather than 8U.
- New roaming destinations will be generated from a radius equal to the long-range radar range, rather than 30U.
- Default AI missile attacks will consider a radius equal to short-range radar range, rather than 4.5U.

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -17,6 +17,10 @@ ShipAI::ShipAI(CpuShip* owner)
     weapon_direction = EWeaponDirection::Front;
 
     update_target_delay = 0.0;
+
+    short_range = owner->getShortRangeRadarRange();
+    long_range = owner->getLongRangeRadarRange();
+    relay_range = long_range * 2.0f;
 }
 
 bool ShipAI::canSwitchAI()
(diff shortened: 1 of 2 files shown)


src/ai/ai.cpp
```diff
@@ -17,6 +17,10 @@ ShipAI::ShipAI(CpuShip* owner)
     weapon_direction = EWeaponDirection::Front;
 
     update_target_delay = 0.0;
+
+    short_range = owner->getShortRangeRadarRange();
+    long_range = owner->getLongRangeRadarRange();
+    relay_range = long_range * 2.0f;
```
daid, line 23: getting this once means if a script changes it on the fly the AI won't be aware of it.

Issue #1319, opened 2021-02-03
[fixed-pipeline] remove glBegin/glEnd pairs


gcask:
- [x]  3D reticule (#1314)
- [x] Starbox (#1313)
- [x]  space objects (black holes, planets, explosions...) (#1303) 
- [x] Particles (#1310)
- [x] spacedust (#1329)
- [x] nebulas game info (#1335, #1331)

Debug features:
- [x] collision boxes (viewport3d) (#1332)
- [x] engine emitters (rotationModelView) (#1332)

PR #1328, opened 2021-02-07
Shader-based spacedust

gcask:
#1319 - Shader-based spacedust.

It's pretty close to where I would like it to be, but still have a few shortcomings (full disclosure, I _suck_ at maths).

It mimicks the existing system by putting points on a sphere around the ship, and passing that ship through it.

Currently it has the following limitations:
  * The pattern is fixed - that's easy to overcome by adding a "generation" factor based on time to increase the illusion of randomness.
  * There is a slight bias towards the center (clustering around the ship). Probably want a non-linear scale for the radius (logarithmic falloff maybe)?
  * Currently, because I'm using the velocity to compute the offset, the look and feel isn't too great during velocity changes (esp. large ones, like coming out warp).
 
 I believe what I want is actually to have a "rolling" sphere center that the ship passes through along its current direction.
My issue is mirroring the existing code where that sphere has a fixed point in worldspace and the ship passes through it (and when the ship goes too far out, it just rolls back in front of it).
Maybe that's something that would be more easily computed on the code side.

Any help welcome :)

As for performance, this cuts down on ~2k draw calls (now it's just one for all particles), as many push/pop matrices, random generation (we just need it to "look good")

### DIFF
--- a/resources/shaders/spacedust.frag
+++ b/resources/shaders/spacedust.frag
@@ -1,0 +1,9 @@
+#version 120
+
+// Shader constants
+const vec4 color = vec4(0.7, 0.5, 0.35, 0.07);
+
+void main()
+{
+    gl_FragColor = color;
+}
--- a/resources/shaders/spacedust.vert
+++ b/resources/shaders/spacedust.vert
@@ -1,0 +1,70 @@
+#version 120
+
+// Shader constants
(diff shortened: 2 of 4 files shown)


gcask:
fwiw this was me having some fun with it - I'm likely going to offer an alternative that matches more closely what the current code does :)

Issue #1333, opened 2021-02-08
compile_script_docs.py can't find certain functions


oznogon:
Given this code in `compile_script_docs.py`:

```
            with io.open(filename, "r", errors="ignore") as f:
                for line in f:
                    if line.startswith('#'):
                        continue
                    res = re.search('([a-zA-Z0-9]+)::([a-zA-Z0-9]+)\(([^\)]*)\)', line)
                    if res != None:
                        self._function_info.append((res.group(1), res.group(2), res.group(3)))
                    res = re.search('^class ([a-zA-Z0-9]+)', line)
                    if res != None:
                        current_class = res.group(1)
                    if current_class is not None:
                        res = re.search('^ *([a-zA-Z0-9 \:\<\>]+) +([a-zA-Z0-9]+)\(([^\)]*)\)', line)
                        if res != None and res.group(2) != '':
                            self._function_info.append((current_class, res.group(2), res.group(3)))
```

The `ScienceDatabase::getKeyValues` function doesn't match:

```
std::map<string, string> ScienceDatabase::getKeyValues()
{
...
```

--

`ScriptStorage` functions also don't match:

```
class ScriptStorage : public PObject
{
public:
    ScriptStorage()
    {
    ...

    void set(string key, string value)
    {
```

`ScriptStorage` is an odd case where it's defined in the `cpp` file with no header.

oznogon:
Appears to be fixed by @edave64's refactoring.

PR #1334, opened 2021-02-08
Split ScriptStorage into a header and code

oznogon:
Add some code docs and edit scripting docs.

### DIFF
--- a/src/scriptDataStorage.h
+++ b/src/scriptDataStorage.h
@@ -1,0 +1,23 @@
+#ifndef SCRIPT_DATA_STORAGE_H
+#define SCRIPT_DATA_STORAGE_H
+
+#include "scriptInterface.h"
+#include <unordered_map>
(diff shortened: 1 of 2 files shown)


oznogon:
Regardless of whether this PR is merged, the result indicates that the structure of `ScriptStorage` causes #1333 with regard to `ScriptStorage` functions.

src/scriptDataStorage.h
```diff
@@ -0,0 +1,23 @@
...
+
+    void set(string key, string value);
+    string get(string key);
+
+private:
+    string scriptstorage_path;
+    std::unordered_map<std::string, std::string> data;
+};
+
+static P<ScriptStorage> storage;
```
daid, line 21: ... that's wrong, a static variable like that shouldn't exist in a header file, means every cpp file that includes it get it's own private copy of it.

Issue #1318, opened 2021-02-03
[fixed-pipeline] Remove fixed lighting


gcask:
This touches the object* shaders.

Currently the lights are static, so they could be baked as constants into the shaders.

PR #1340, opened 2021-02-10
Removing fixed lighting

gcask:
#1318: Taking care of fixed pipeline lighting functions.

Took the shortest possible route (which incurs a little duplication).

I can file an issue about moving them back into uniforms once we got rid of the fixed functions.

Ideally I want to batch up the uniforms update for the matrices - lights could be done in the same pass/fashion (but first, the fixed functions need to go!)


daid:
Moving of the shaders caused this to be in conflict.

Also, with the Planet() object can be configured to look like a star, so could be nice to configure the light point based on that at some point. If a scenario spawns a star.

gcask:
Interesting - locally git was able to do the forward merge (so no conflicts).

I'm with you 100% about the potential for lighting improvements that this opens :)

We could also have ships with actual lights, space lighthouses, pulsars... Cool stuff.
That's why I mentioned turning them into uniforms.

But managing more lights and dynamic ones is a whole other can of worms, and feels outside this scope :/

PR #1341, opened 2021-02-11
Fixed triangle winding of starbox top face

gcask:


oznogon:
LGTM

for context, this fixes this behavior when looking straight up with the cinematic cam in unlocked mode

Screen Shot 2021-02-10 at 4 17 16 PM

src/screenComponents/viewport3d.cpp
```diff
@@ -104,7 +104,7 @@ GuiViewport3D::GuiViewport3D(GuiContainer* owner, string id)
             2, 6, 4, 4, 0, 2, // Back
             3, 2, 0, 0, 1, 3, // Left
             6, 7, 5, 5, 4, 6, // Right
-            7, 3, 1, 3, 1, 5, // Front
+            7, 3, 1, 1, 5, 7, // Front
```
gcask, line 107: If you look at the cube's picture (above): 7, 3, 1 and 3, 1, 5 actually overlap... and leave the 7, <center>, 5 triangle open.

*Ooops.*
7, 3, 1 followed by 1, 5, 7 is the same pattern as all the other faces and properly makes the two non-overlapping triangles of the quad :)

PR #1351, opened 2021-02-19
Remove client arrays

gcask:
This removes the client arrays in the object* shaders.

There's a lot - I recommend going the changes commit by commit to get a more "logical" view of the changes.

Because it meant generalizing the need to be able to pass in attributes and setting up uniforms (a necessary step to completely remove the fixed pipeline), I ended up adding a 'shader registry': a centralized way to access the game's shaders.

Additionally, thanks to how OpenGL shader attribute binding works, I was able to ensure all similarly named attributes from different programs end up having the same locations - meaning we can now bind vertex attrib arrays once for multiple shaders.

While now only leveraged in the explosion fx, we could enable common arrays (like the position ones) once before the render loop.

Lastly, now that nothing goes through any fixed pipeline (aside the matrices), we're free to simply do a state reset on the SFML window, instead of a full blown push/pop. (I did not notice any difference doing so, but lower spec system might).


daid:
Some unrelated change for the CI sneaked into this?

gcask:
Gah! Silly me, it's based off of the wrong branch 🤦 
It was bound to happen at some point :)

I'll close this a recreate one which is correctly based off of this repository.

PR #1353, opened 2021-02-23
Being able to set a reverse speed and reverse acceleration

Tsht:
Be able to set a reverse speed and reverse acceleration on ship and templates.
Reverse speed of 0 is supported.
You can tweak reverse speed on GM tweak screen.
By default, reverse speed is equal to front speed (same for acceleration), old scripts are compatible.


src/shipTemplate.cpp
```diff
@@ -414,6 +418,19 @@ void ShipTemplate::setSpeed(float impulse, float turn, float acceleration)
     impulse_speed = impulse;
     turn_speed = turn;
     impulse_acceleration = acceleration;
+
+    if((impulse == 0.0f) && (impulse_reverse_speed == 0.0f)) // hack supposing we keep existing invariant, to handle template copies
```
hemmond, line 422: Does it mean, that it will set reverse speed only if I call `template:setSpeed(0 , 5, 1)`? Because you are comparing against input argument (which I doubt will be a zero). And in constructor, you have `impulse_reverse_speed = 500.0` so I have no idea how this hack could work.

src/spaceObjects/spaceship.cpp
```diff
@@ -56,10 +56,14 @@ REGISTER_SCRIPT_SUBCLASS_NO_CREATE(SpaceShip, ShipTemplateBasedObject)
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, setSystemCoolant);
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, getImpulseMaxSpeed);
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, setImpulseMaxSpeed);
```
hemmond, line 58: I believe here you are not backwards compatible. If someone in his mission script changed speed and acceleration of some ship, it would now change only for forward speed and it will keep reverse speeds intact, would it not? 

Just want it to be clarified (so it can be specified in documentation, that there were change in the behavior).

src/shipTemplate.cpp
Tsht: It means it's pure madness I don't even understand how I could test templates with this... Just correcting :( So sorry.

src/spaceObjects/spaceship.cpp
Tsht: The main idea is : if only using setSpeed, then forward speed is always the same as reverse speed
If set reverseSpeed, then reverseSpeed is always inherited and not overwritten by a new "setSpeed".
Still, using an optional in setSpeed with two optional arguments would be better

src/spaceObjects/spaceship.cpp
```diff
@@ -295,9 +295,24 @@ void SpaceShip::applyTemplateValues()
     energy_level = max_energy_level = ship_template->energy_storage_amount;
 
     impulse_max_speed = ship_template->impulse_speed;
-    impulse_max_reverse_speed = ship_template->impulse_reverse_speed;
+    if(ship_template->impulse_reverse_speed != -1)
+    {
+        impulse_max_reverse_speed = ship_template->impulse_reverse_speed;
```
hemmond, line 300: Does it mean, that if I use the `StarShip:setSpeed()` it will only modify Forward impulse speed (even though I didn't change reverse speed)? 

If so, I would propose changes to have these functions exported to LUA: 
- `StarShip:setSpeed(speed)` => `impulse_max_reverse_speed = impulse_max_speed = speed`
- `StarShip:setForwardSpeed(speed)` => `impulse_max_speed = speed`
- `StarShip:setReverseSpeed(speed)` => `impulse_max_reverse_speed = speed`

That way it will keep the same functionality (and honestly, it would make sense). 

It we merge this commit, mission script like this will work differently (and strangely):
```ps = PlayerShip():setTemplate("Atlantis")  -- code from template: atlantis_variant:setSpeed(90, 10, 20)
ps:setSpeed(120)```
This would set Atlantis forward speed to 120, but it will keep reverse on 90. And this WILL be confusing to mission authors. They did not change Reverse speed, but the speeds are different. And it could break some mission scripts of theirs. 

DTTO on ship_acceleration.

Comment (Tsht) [reply src/spaceObjects/spaceship.cpp:300]: Waiting for acceptance or not of std::optional so SetSpeed could just be amended.
For the getter, not fond of deprecated methods but I can't still decide if having a x,y = getImpulseMaxSpeed() which would break formatter strings or multi affectation like x,y,z = getImpulseMaxSpeed(),10 is really worse. 
Maybe an optional getImpulseMaxSpeed(front) (by default) and getImpulseMaxSpeed(rear) or a getImpulseMaxSpeeds() ?
I will come with additional proposal.

Comment (Tsht) [reply src/spaceObjects/spaceship.cpp:300]: For now, for me optional arguments solves the setter issues
**On shiptemplate**
setSpeed have a variable number of args.
In lua, 

- setSpeed(forward speed, turn rate, forward acceleration) will set all speed and all acceleration to the requested value

- setSpeed(forward speed, turn rate, forward acceleration, reverse speed) will set forward speed and reverse speed as required, acceleration is the same in both way (don't think we always want to set different acceleration...)

- setSpeed(forward speed, turn rate, forward acceleration, reverse speed; reverse acceleration) will set forward speed, reverse speed, acceleration and reverse acceleration as required

Please note there is no reverse turn rate, I don't think its useful but if added, this would mess logic of argument order.

** On spaceship **
- setMaxImpulseSpeed(speed) will set max forward speed and reverse speed with the same value
- setMaxImpulseSpeed(speed, reverse speed) will set max forward speed and max reverse speed as required

Remains getMaxImpulseSpeed
Here are the proposals :

1. getImpulseMaxSpeed() will return 2 values, like getVelocity 
Pros : 
- you can discard second value
-  it will not break this kind of scenarion : x = getImpulseMaxSpeed() considering forward speed is the wanted speed.
Cons : 
- will break this kind of scenario : x,t = getImpulseMaxSpeed(),OtherFunctionReturningT() (also in string.format("%f %f %f",getImpulseMaxSpeed(), OtherFunctionReturningT())

2. getImpulseMaxSpeed() will return only forward speed. getImpulseMaxSpeed(forward) will return forward speed. getImpulseMaxSpeed(reverse) will return reverse speed.
Pros :
- Won't break anything (except for scripts which won't take reverse speed into account...)
Cons :
- Ugly
- Not so simple to understand why getImpulseMaxSpeed would be forward by default... 

3. deprecated getImpulseMaxSpeed() which returns forward speed, or maybe null if reverse speed is different.
As hemmond proposed, but with only one method for the new function.
getImpulseMaxSpeeds() which will return both speed (forward and reverse)
Pros :
- Won't break anything (except for scripts which won't take reverse speed into account...)
Cons :
- Deprecated function

Comment (hemmond) [reply src/spaceObjects/spaceship.cpp:300]: I agree with your setters, I think it is future-proof enough. And messing up logic in ship template methods (if this case happens in the future) is smaller pain than messing it up in functions that are used throughout the mission scripts (and therefore read a lot more by a lot more people). 

About your proposals about getters: 

1. Option 1 is no-go, unless Daid says otherwise. It will break missions. 
2. I can life with option 2 with selectors "forward", "reverse", "both" and empty (which means "forward"). Description why it is not so simple to understand can be explained in documentation (for backwards compatibility, please do not use it in the future). 
3. I agree in principle with option 3, but one letter difference will confuse a lot of mission creators. It is easy to miss that letter and then be confused why it returns different number of values than expected. 
4. My proposal (option number 4): 

Add methods `getImpulseMaxForwardSpeed()`, `getImpulseMaxReverseSpeed()`, `getImpulseMaxBothSpeeds()` and deprecate function `getImpulseMaxSpeed()`. I know `getImpulseMaxBothSpeeds()` is ugly, but if you drop it, you fall to the same trap as in option 3 - confusion when missing a letter.

Comment (Tsht) [reply src/spaceObjects/spaceship.cpp]: Went with 3 with your amelioration proposal on name.

File: src/spaceObjects/spaceship.cpp [hunk 4]
```diff
@@ -54,12 +54,28 @@ REGISTER_SCRIPT_SUBCLASS_NO_CREATE(SpaceShip, ShipTemplateBasedObject)
...
+    ///If called with only one argument, sets forward and reverse speed to equal values.
+    ///If called with two arguments, first one is forward speed and second one is reverse speed.
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, setImpulseMaxSpeed);
+    ///Get multiple results, first one is forward speed and second one is reverse speed.
+    ///ex : forward,reverse = getBothImpulseMaxSpeeds() (you can also use select or _ to get only reverse speed)
+    REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, getBothImpulseMaxSpeeds);
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, getRotationMaxSpeed);
     REGISTER_SCRIPT_CLASS_FUNCTION(SpaceShip, setRotationMaxSpeed);
+    ///[Deprecated] Gets forward impulse acceleration.
+    ///Consider using getBothAccelerations as forward and reverse acceleration can be different.
```
daid, line 70: I don't think this one needs to be depricated. Remembering my lua right, if you now do:
```
local acc = ship:getBothAccelerations()
```
you just get the forward acceleration, so the getAcceleration implementation can just be replaced with the getBothAccelerations implementation, and getBothAccelerations can be dropped. (yay, less code)

(same for getting speed)

daid (commented):
Pull request is in conflict now as well, should be something simple I think.

src/spaceObjects/spaceship.cpp
Tsht: Yes I wanted to go with this :)
For me it would break this kind of usage
x,t = getImpulseMaxSpeed(),OtherFunctionReturningT() 
or 
string.format("%f %f %f",getImpulseMaxSpeed(), OtherFunctionReturningT())

But I'm not sure it represents even any ever written line of code as it's (for me) broken an unspecified lua code.
I would also prefer going with what you propose.

Tsht: Just to be clear, latest commit includes this proposal so there is no more getBothXXX methods, methods are returning values which are to be discarded (in legacy case, it is) or not, directly in the lua.

daid: Yeah, I think this is good, the chance of breaking anything is really small. And not seeing anything like it in the current scenarios.

PR #1355, opened 2021-03-03
Cz translation

hemmond:
I have updated most of important CZ tranlation. I am still missing ship and faction descriptions, most of the ship names and ship classes.


aBlueShadow:
> I am still missing ship and faction descriptions, most of the ship names and ship classes.
I'd advise against translating the ship names currently, due to #1080

PR #1363, opened 2021-03-19
Static waypoint labels

hemmond:
Waypoint labels are now static. After waypoint removal, the rest of the waypoints retains their number. When adding waypoint, first free number is used as label of the waypoint. 

I recommend squashing commits into one during merge.


daid:
This breaks how the LUA API is used:
https://github.com/daid/EmptyEpsilon/blob/master/scripts/comms_ship.lua#L52

hemmond:
Could `PlayerSpaceship:getWaypointLabel(int index)` then be introduced to LUA, which will return proper label? Stock LUA scripts will then still work as expected, custom scripts will have "different" numbers but will not crash. Could that be accepted?

PR #1374, opened 2021-03-27
Addition translation context for ships

muerteFR:
Added translation contexts for ships : 
"ship", "shipP", "class" and "subclass"


scripts/shipTemplates_StarFighters.lua
```diff
@@ -133,7 +133,7 @@ template:setTubeSize(2, "small")
 template:setTubeDirection(1, 1):setWeaponTubeExclusiveFor(1, "HVLI")
 template:setTubeDirection(2,-1):setWeaponTubeExclusiveFor(2, "HVLI")
 
-variation = template:copy("ZX-Lindworm"):setLocaleName(_("ZX-Lindworm")):setModel("LindwurmFighterBlue"):setType("playership")
+variation = template:copy("ZX-Lindworm"):setLocaleName(_("shipP", "ZX-Lindworm")):setModel("LindwurmFighterBlue"):setType("playership")
```
daid, line 136: "ship**P**"

muerteFR: Yes, I put the context ShipP(layer) for the 18 player ships. To clarify more.
I tested this PR locally with ninja update_locale and everything works fine except for an issue already present on version 23112020: player ship names are not translated. (classes and subclasses yes but no the ship names)

muerteFR: And It's not just with fr.po, it's general. This is also not a problem with accents.

daid: shipP, why not just playerShip then?

muerteFR: Because I thought that the context shouldn't be too long.
Then I replace them in playerShip. :-)

muerteFR: Done

muerteFR: @daid , is there a contraindication to validating this PR? Thanks

PR #1375, opened 2021-03-28
Spacedust could use unitialized data

gcask:
Happens when the ship is initially at full-stop *and* the ship is around the (0, 0) world position - the positions in the buffer have not been initialized and will remain so until the ship moves enough.

Moreover, ensuring the data type sent from cpu to the shaders lines up for the signs.


gcask:
cc @StarryWisdom

PR #1378, opened 2021-03-28
Cz translation - SCI DB bugfix

hemmond:
Removed most potentialy fragile database index keys in CZ translation.


hemmond:
There is still problem with Borderline Fever/Chaos of war (something with fighter_db local variable - going to debug today)

hemmond:
Problems are fixed and this is ready to merge. CZ translation (while still not entirely complete) should not break other things.

PR #1385, opened 2021-03-31
CMake 3.8

gcask:
* Switched to using SeriousProton's new target.


CMakeLists.txt
```diff
@@ -1,571 +1,413 @@
...
-      execute_process(COMMAND sed -i "s/<Windows.h>/<windows.h>/g" c/discord_game_sdk.h WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/discord")
-    endif()
-  endif()
-  
-  target_link_libraries(${EXECUTABLE_NAME} "${DISCORD_LIB_PATH}/discord_game_sdk.dll.lib")
-  target_include_directories(${EXECUTABLE_NAME} PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/discord/c/")
-elseif(ANDROID)
+target_link_libraries(${PROJECT_NAME}
+    PUBLIC
+        seriousproton glm::glm opengl
```
daid, line 336: I remember having issues with linking seriousproton as library before. Something with static initialization order not going right. But it's a vague memory...

gcask: You did, but I haven't encountered any issues...

I also use SP as a library for my SDL branch and nothing came up there either.

Currently triple checking (I did Windows, Debian, RPi already but I can do a second pass)

gcask: Triple checking for the win: I think I saw the issue you recall.

It's not about static initialization order, so much as [LTO bringing mayhem](https://stackoverflow.com/q/27372667) with MinGW.

It required a little finesse, but I have something that's working :) (requires additional tweaks on SP's target setup).

PR #1388, opened 2021-04-01
Android CI target

gcask:
Builds the APK through github's CI.


gcask:
This shares some commonality with #1159 but sandbox itself to github actions, and does not do artifact publishing - it's only a CI build check.

PR #1390, opened 2021-04-02
warning fixes

gcask:
* typecasting
* inheritance by dominance


gcask:
Will do the same as for SP, and provide separate PRs.

PR #1393, opened 2021-04-04
Scenario waves.lua and update beacon.fr.po

muerteFR:
1. Addition of 2 missing _().
2. Create waves.fr.po
3. Update Beacon scenario French translation.
4. Adding accents.
5. Correction of the meaning of sentences.


scripts/scenario_03_waves.lua
```diff
@@ -122,7 +122,7 @@ end
 
 function spawnWave()
     waveNumber = waveNumber + 1
-    getPlayerShip(-1):addToShipLog("Wave " .. waveNumber, "red")
+    getPlayerShip(-1):addToShipLog(_("Wave ") .. waveNumber, "red")
```
daid, line 125: The right way to add these markers is:
`string.format(_("Wave %d"), waveNumber)` so translators can decide of the number needs to come first or not.

muerteFR: Like that?

muerteFR: Done

PR #1394, opened 2021-04-10
Wrapper for addCustomButton, Info, Message

hemmond:
Since we have a lot of problems with different scenarios only showing mission specific stuff (custom messages/custom buttons) on 6/5 crew screens, the missions usually can become unplayable in 4/3 because mission does not show this info on this screens. This wrapper can allow creators to replace current lines with minimal changes to show the same thing on all stations of the same operator (for example Engineering and Engineering+). 

I would like to get this merged before I start changing stock missions to this wrapper (and before I start using this module in other modules). 

Any tips and suggestions to improve this module are welcome. 

[Demo mission to check functionality](https://github.com/hemmond/EE-Scenario-snippets/blob/main/scenario_00_customElements.lua)

### DIFF
--- a/scripts/utils_customElements.lua
+++ b/scripts/utils_customElements.lua
@@ -1,0 +1,163 @@
+-- Name: utils_customElements
+-- Description: Wrapper upon different stations, so mission author can add button/information to station Operator, 
+--- abstracting the position through which the operator fulfills his duties. 
+--- This module should remove multiple boilerplate to accomodate 6/5 and 4/3 station categories. 
+--- Operator in context of this module means player fulfilling specific duties aboard the ship
+--- (for example Engineer can use "Engineering" or "Engineering+"). 
+
+--- Module API description: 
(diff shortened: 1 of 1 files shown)


hemmond:
1) Every scenario can change default mapping of stations to make sure it shows where author wants it to show (and I expect, that in more complex scenarios, this will be used to - for example - drop AltRelay from "Relay" category). 
2) CustomButton and CustomInfo can be also on AltRelay, (at least both will have the same info - AltRelay just lacks comms, otherwise it is the same, so I expect the same data), CustomMessage by default pops on all stations and first one to close it will close it for all other stations. This is enabled by default and can be disabled by calling  `customElements:closeAllMessagesUponClose(false)`. 

I intend to start replacing on Xansta's missions - and with his assistance, so we might tweak the categories later on (when we find out, that we change "Relay" category too often). Later on, I hope I will get mission authors to help me with the replacing, so it done how it was supposed to be done.

PR #1405, opened 2021-04-23
Coherent winding order

### LINKED ISSUE #294: Text on space stations textures are mirrored

gcask:
This harmonizes the winding order to be counter-clockwise across the board (I checked the nebulas, wormholes, beams, engine trails, particles, asteroids, stations, black holes, planets, stars... all render OK).

Meshes do need to be rotated around the Z axis because of how EE massages the matrices around for the 3D pass (and how the entire world appears slanted 90 degrees overall, but that is a story for another time maybe :) )

fixes #294.

### DIFF
--- a/packs/pack_gen.py
+++ b/packs/pack_gen.py
@@ -29,10 +29,10 @@ def convertObj(filename):
 		for i in xrange(2, len(face)):
 			for n in [0, i, i-1]:
 				v = vertices[face[n][0] - 1]
-				vt = uvs[face[n][2] - 1]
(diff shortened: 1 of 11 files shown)


daid:
>  (and how the entire world appears slanted 90 degrees overall, but that is a story for another time maybe :) )

Assuming this is rotation 0 is to the right, so heading 90? Not really fixable without breaking API.

PR #1408, opened 2021-04-25
Add angle() function to utils.lua

hemmond:
Complementary function for `angleToVector()`, with simillar API as `distance(a,b,c,d)` function, `angle()` returns bearing between first object/position and second object/position.

### DIFF
--- a/scripts/utils.lua
+++ b/scripts/utils.lua
@@ -5,7 +5,7 @@
 -- These functions should be as generic as possible, so they are highly usable.
 
 -- Given enough information, find the distance between two positions.
--- This function can be called four ways:
+-- This function can be called in four ways:
 --
 -- distance(obj1, obj2)
 --   Returns the distance between two objects.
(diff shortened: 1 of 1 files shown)


daid:
Due to EE's coordinate system being screwy, it's for me unclear if this function returns a heading or a rotation vector.

Example: `player:setRotation(angle(player, station))` or `player:setHeading(angle(player, station))`?

hemmond:
It returns rotation vector (0° is "to the east"). I could rename it to `angleRotation()` and add `angleHeading()` for heading (0° is "to the north"), so the utils.lua provides both functions. I can also try to document this "dual angling" in the documentation of both functions (and rename function arguments to `heading_angle`/`rotation_angle` in `utils.lua` depending on what angle it accepts). Is that OK?

hemmond:
So, I have implemented `angleRotation()` and `angleHeading()` instead of original `angle()` function along with its respective documentation comments. I have also introduced `angle_is_heading` optional variable to `setCirclePos()` and `vectorFromAngle()`. It should now provide all functions for both numeric use-cases (rotation angle and heading).

PR #1409, opened 2021-04-25
Basic scenario enhanced GM controls

hemmond:
scenario_00_basic.lua GM functions was improved:

- Xansta's Improvements of GM menu (enemy waves in sub-menu)
- Wave positioning can be switched between random positioning (behavior that was in basic scenario before) and between GM positioning (where GM clicks on position where the wave should be spawned). 
- Victory button now asks for confirmation before declaring victory. Information that players have won is then shown in GM screen instead of other GM buttons (GM buttons does nothing after victory anyways). 

This pull request is dependent on [Pull Request #1408](https://github.com/daid/EmptyEpsilon/pull/1408) as angle() function is needed from utils.lua

### DIFF
--- a/scripts/scenario_00_basic.lua
+++ b/scripts/scenario_00_basic.lua
@@ -27,7 +27,7 @@ local stationList
 
 --- Add an enemy wave.
 --
--- That is, create it and add it to `list`.
+-- That is, create enemy wave and add all its ships to `list`.
 --
 -- @tparam table list A table containing enemy ship objects.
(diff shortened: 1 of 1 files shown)


hemmond:
Rebased against current repository since there were changes in this file in commit [0e50318](https://github.com/daid/EmptyEpsilon/commit/0e50318db8a4c253f46df9c900998cf7a0fea666). Fixed so it now uses current version of `utils.lua` (uses function `angleRotation()` instead `angle()`). 

Can be merged right after #1408.

PR #1412, opened 2021-04-28
Being able to list android files (remove slashes)

Tsht:
This is to merge at the same time as the Serious Proton pull request daid/seriousproton#109.
Android needs directories not to have trailing slash.
Instead of making operation on paths for android (removing slashes), use the same convention as Serious Proton 2.
DirectoryProviders should not provide trailing slash, and they are added on each provider if needed.
L109 is kept for not changing the behavior as linux and windows won't care if there is an extra slash (and only want to address Android issue here) (can be cleaned up after)


daid:
With the changes I proposed in SeriousProton, is this still needed?

Tsht:
> 
> 
> With the changes I proposed in SeriousProton, is this still needed?

Nope, I came to the same conclusion after making the change, except maybe for skipping home path on android, but this won't change anything even without. I guess we can skip this PR, for one backslash erasure I don't think this will be a performance issue :)

Tsht:
Not needed after all.

Issue #758, opened 2020-02-23
Radar display bug on android and possible workarounds


aBlueShadow:
A common display bug on android  is the radar bug. There is no grid nor distance circles, the radar shows a black background, and nebulae are not displayed. When this bug is present, it affects all stations with a circular radar, so i.e only engineering, relay and derived stations are fully playable. I am not sure how common this bug is, it is surprisingly rarely mentioned when someone is talking about android, and daid said once on his deviced everything is displayed fine - but it was present on every device I was able to test it on so far.

![Screenshot_1](https://user-images.githubusercontent.com/25465934/75119053-cbe77380-567f-11ea-84c0-085748778e1e.png)

The obvious difference between relay and the other stations is, that the others are circular, so I changed one to rectangular mode, and suddenly it works:
![Screenshot_20200223-035458](https://user-images.githubusercontent.com/25465934/75119108-46b08e80-5680-11ea-893e-c0bcc2eb9eb1.png)

Looking a bit into the code, it seems that the rendertexture does not work correctly. Instead of using black to cut out a circle, the blend mode seems to be ignored, drawing just a black disc on top of other elements. 

https://github.com/daid/EmptyEpsilon/blob/277f98c1326005d029dd127118970b1b4169b4e9/src/screenComponents/radarView.cpp#L123

I have no idea how to fix that (if possible), and as I currently don't have a virtual android setup on my pc, testing is kinda tedious.
A quick workaround would be to add an options.ini switch to change all radars to rectangular. Not a pretty option, but could be beautified a bit, e.g. by changing the background or drawing a rectangular line around the radar. Also, the officer would be able to see a bit more than usual, but I don't think that is a big deal and could be mitigated by letting that option only affect android.

Thoughts?

daid:
A problem there is SFML's API being an total shitstorm on the blend modes. I didn't know it when I made that code, but it acts like there is a lot of flexibility, but in reality that only works if a specific OpenGL extension in available:
https://github.com/SFML/SFML/blob/519e1f95727b353124fe5c891bc3d1743289463a/src/SFML/Graphics/RenderTarget.cpp#L581

Quite potentially, that is causing the problems.

Note that in your 2nd screenshot, you don't have any areas blacked out by the nebulae fog of war.

oznogon:
Related SFML forum thread from 2016: https://en.sfml-dev.org/forums/index.php?topic=20969.0

[...]

PR #1413, opened 2021-04-29
Stencil-based radar

gcask:
Switched the radar to use a stencil buffer.
It still renders into its own texture, but just one instead of three.

Understanding this is a core piece, the more eyes the better :)


gcask:
fixes #758

daid:
Well... here goes nothing...

PR #1422, opened 2021-05-03
Create scenario_07_gftp.en.po

muerteFR:


muerteFR:
It can be validated Daid :-)

PR #1423, opened 2021-05-05
UI language selection

Tsht:
This is to have the option to change the language in the UI, without modifying the option file by hand.
It adds a language option in Options/Interface Options (see screenshot).
Language cycle through every language available in resources/locale
It works on Android. Tested only on linux and android.

Language is applied only when pushing "back" button (refreshing the page in realtime seems tedious, and may not be a good idea as any click on the arrows could make the interface not understandable).
Language choice is saved if "save" is chosen.
Default language is the one in language file, or "en" if none available. (or falls back to first available if no "en", if no language at all should just propose a black box)

image

Bonus french translation included.


Tsht:
Convert to draft, thought of P<X> as auto release instead of thinking to call destroy instead ^^;

src/menus/optionsMenu.cpp
```diff
@@ -169,6 +169,37 @@ OptionsMenu::OptionsMenu()
...
 
+    //Select the language
+    (new GuiLabel(interface_page, "LANGUAGE_OPTIONS_LABEL", tr("Language (applies on back)"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
+    
+    //Don't wan't scripts/locale. Also, compatible with android assets.
+    DirectoryResourceProvider* lang_prov {new DirectoryResourceProvider("resources/locale")};
+    std::vector<string> languages {lang_prov->findResources("*.po")};
+    lang_prov->destroy();
+
+    std::for_each(languages.begin(), languages.end(), [](string &str)
```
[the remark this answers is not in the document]
Tsht: Agree, carried on by the remove_if into old C++ I guess.
Also added a sort on valid languages, as files are not necessarily sorted by alphabetical order.

src/menus/optionsMenu.cpp
```diff
@@ -173,8 +173,10 @@ OptionsMenu::OptionsMenu()
     (new GuiLabel(interface_page, "LANGUAGE_OPTIONS_LABEL", tr("Language (applies on back)"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
     
     //Don't wan't scripts/locale. Also, compatible with android assets.
-    DirectoryResourceProvider* lang_prov = new DirectoryResourceProvider("resources/locale");
+    DirectoryResourceProvider* lang_prov {new DirectoryResourceProvider("resources/locale")};
```
gcask, line 176: is there a reason you're not using a variable on the stack directly?

Tsht: Yes, see p.h line 6 to 14.

src/menus/optionsMenu.cpp
```diff
@@ -169,6 +169,38 @@ OptionsMenu::OptionsMenu()
         destroy();
     }))->setSize(GuiElement::GuiSizeMax, 50);
 
+    //Select the language
+    (new GuiLabel(interface_page, "LANGUAGE_OPTIONS_LABEL", tr("Language (applies on back)"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
+    
+    //Don't wan't scripts/locale. Also, compatible with android assets.
+    DirectoryResourceProvider* lang_prov {new DirectoryResourceProvider("resources/locale")};
+    std::vector<string> languages {lang_prov->findResources("*.po")};
```
daid, line 177: Why not `findResources("locale/*po")`? that saves constructing the resourceprovider.

Tsht: This would not search and return scripts/locale/*po scenario specific translations as well ?

Tsht: I don't think it's better to list more files to filter out after... But another possibility would be to have the global langage files in another directory ? So it's possible to search them with another hierarchy.

daid: But you would already filter those out at line 186?

PR #1429, opened 2021-05-13
Fix blackhole blend function

gcask:
BlackHole was changing the active blend function without restoring the default one. This lead to issues in some circumstances.


gcask:
L67 (slightly before the change) is where `BlacHole::draw3dTransparent()` changes the blend function, wrecking havoc with anything that comes after (the blend function is set only once in the 3d viewport).

PR #1431, opened 2021-05-16
Make discord available to more platforms

gcask:
Make discord available to more platforms (ie unix-like).

Requires daid/SeriousProton#110


gcask:
I have checked Windows and Linux are working fine (checked linux package installation).
It's likely macOS will require an iteration or two, I'll have to find volunteers :)

PR #1433, opened 2021-05-17
Use deferred translation for GMInfo.

### LINKED ISSUE #1417: Text cannot be translated in GM
I reopen the issue regarding some entries in GM which cannot be translated because the tags cannot be added as is.
These are issues 7 and 8 of this issue: https://github.com/daid/EmptyEpsilon/issues/1256

Oznogon had responded regarding these issues. The question is: Is it possible to modify this to allow translation?

"Where to translate in the GM console always: the indications at the top right? "Callsign", "Shield", "Hull", "position", "Orders" Some are already translated as "Time elapsed" ... Still in GM, the orders at the bottom right are translated but at the top right they are not : ("roaming", Idle", "Defend location" etc."

This is the selection_info object in GameMasterScreen:
EmptyEpsilon/src/screens/gm/gameMasterScreen.cpp Lines 295 to 334 in 0a40f22 :

"The information comes from SpaceObject::getGMInfo(), which is used only by the GM screen. For ships, the pertinent override is in ShipTemplateBasedObject, which returns an unordered_map of non-localized strings, and are keyed on those strings as well (some of which are hardcoded in English):"

EmptyEpsilon/src/spaceObjects/shipTemplateBasedObject.cpp Lines 255 to 266 in 5c37189 :
"selection_info relies on at least one hardcoded English-language string as a key (selection_info["Position"])."

gcask:
Use deferred translation for the GMInfo map.

Using deferred translation allow the map to still rely on the english keywords, while providng translators a way to translate them.

The translation is only done at the last minute, on the Key/Value display.

fixes #1417 .

### DIFF
--- a/src/screens/gm/gameMasterScreen.cpp
+++ b/src/screens/gm/gameMasterScreen.cpp
@@ -321,7 +321,7 @@ void GameMasterScreen::update(float delta)
 
     if (targets.getTargets().size() == 1)
     {
-        selection_info["Position"] = string(targets.getTargets()[0]->getPosition().x, 0) + "," + string(targets.getTargets()[0]->getPosition().y, 0);
+        selection_info[trMark("gm_info", "Position")] = string(targets.getTargets()[0]->getPosition().x, 0) + "," + string(targets.getTargets()[0]->getPosition().y, 0);
     }
 
     unsigned int cnt = 0;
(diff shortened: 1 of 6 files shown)


src/screens/gm/gameMasterScreen.cpp
```diff
@@ -321,7 +321,7 @@ void GameMasterScreen::update(float delta)
 
     if (targets.getTargets().size() == 1)
     {
-        selection_info["Position"] = string(targets.getTargets()[0]->getPosition().x, 0) + "," + string(targets.getTargets()[0]->getPosition().y, 0);
+        selection_info[trMark("gm_info", "Position")] = string(targets.getTargets()[0]->getPosition().x, 0) + "," + string(targets.getTargets()[0]->getPosition().y, 0);
```
gcask, line 324: @muerteFR Any preference for a context name? Couldn't really remember which convention you went with...

Seeing as you made the bulk of the translation work, let me know if it's not in line with the rest.

muerteFR: It seems very clear for me :-) Thanks for asked me :-)

Issue #707, opened 2019-11-12
[Request] - GUI Notification On Client For Authentication Failure


JonathanILevi:
When the client's connection is refused because of a version mismatch it should show the user with a graphical pop up.  As stated in #244 autodiscovery does not show version mismatched servers.  But attempting to connect can happen when manually typing in IPs (which some players use a lot).

#244 should be fixed first.

I expect to get around to fixing these 2 issues at some point, just wanted to make them officially known, and potentially discussed.

oznogon:
PR daid/SeriousProton#39 logs the version mismatch issue to the client logs:

```
[INFO]  GameClient: Connected, waiting for authentication
[INFO]  Server version 20191101 does not match client version 20200111
```

A GUI notification might require a little more detail to be sent back from SeriousProton's multiplayer_server on an auth failure, but it might still be straightforward to add a GUI notification that points at the log for details.

oznogon:
Auth flow is generally:

1. Client requests to connect, sends the client version.
2. Server checks the client version against the server version. If it matches, or if either is 0, proceed. Else, log an error on the server and disconnect.
3. Server receives the request and checks if a password is set.
    1. If a password is set, check if it's correct.
        1. If so, handle the new client, _etc._
        2. If not, send a new auth request. _No rejection is sent, just a new auth request._
    2. If no password is set, handle the new client, _etc._ 

So to handle a failed password check, either:

- the client has to recognize that it is receiving repeated auth requests, or
- the server's reply has to communicate the failure

To handle a failed version check, either:

- the client has to recognize that it was authenticated but not connected, or
- the server has to send a reply before disconnecting

JonathanILevi:
I was thinking to have the server respond to the client on the form of failure (the latter of the two options).  I think it would be the simplest way at leaves out complicating guess work.  I have not gotten to fixing this yet so if you do different that would be fine by me.

JonathanILevi:
The log you showed is from the server though, isn't it?

oznogon:
I don't think I have a different idea, I just don't have a solid plan to get it into the GUI.

The log is from the client as a result of PR daid/SeriousProton#39.

JonathanILevi:
Oh, I didn't realize that the client receives the version number.  It seems silly that it does.  Also the fact that the logic for whether to authenticate happens on both sides and with different code smells bad but whatever.

[...]

PR #1435, opened 2021-05-18
Display disconnect reason in server browser

### LINKED ISSUE #940: Server Client Version Mismatch
When trying to connect a client to a server, if the client's version doesn't match the server's then the client will ask for the server password and, after a seemingly random interval, drop back to the server address screen.

The command line indicates the version mismatch, but the UI should probably do so as well.

gcask:
In the server browser, when the player fails to connect, display a reason on screen.

image

This is currently most useful to quickly spot version mismatch error (one that seems common).

requires daid/SeriousProton#111

fixes #940


gcask:
Also fixes #707

Issue #1003, opened 2020-08-04
Windows Installer


JonathanILevi:
Some users find it difficult or confusing to use an app from a zip folder.  It also comes off unprofessional or different which can give some users concern of the legitimacy or safety of the app.  It can also be annoying for users unsure where to put the extracted folder.  These are particularly problems for users not comfortable with folder structures, which are more common since the spread of smartphones.

@daid would you be open if I setup a system, that fits into your current build process, to create an .msi installer for the releases to replace (or reside along size) the current .zip download on the website?  It shouldn't be too complicated to create an msi, and without using some junky freeware.

daid:
> It shouldn't be too complicated to create an msi

Automated? From linux?

JonathanILevi:
I didn't remember how you built EE, so that was one of my questions.

I think it should be automatable on Linux, I think it is just a zip file with specific datafiles.

JonathanILevi:
That, or maybe create an executable which downloads and installs the latest version from the website.

NinjaSpectre:
Perhaps just create a self-extracting archive (SFX) with 7-zip.  The 32-bit Windows installer 7-zip provides is just a SFX itself so it should be extractable on Linux.  Then use the Windows binary 7z.sfx with the Linux 7z command.

`7z a -mhe=on -p -sfx7z.sfx EmptyEpsilonInstaller.exe EmptyEpsilonFolder`

JonathanILevi:
That does sound like an option @NinjaSpectre.  Can self-extracting archives install into the Program Files folder?  (An uninstall would also be advantageous)

I found this Stack Overflow question: https://stackoverflow.com/a/55089464/2521261

There might be a better way, as the meta-data does not need to be changed.  Several other places on the subject used Wine.

I'll fiddle with it and see what I can figure out.  I'll probably create an Msi on Windows first and then either replicate the process on Linux, or modify the Msi on Linux.

gwaland:
I'd tried to massage NSIS to create an installer package since NSIS is already supported in cmake and cpacker, but I've run into a lot of snags with it so not sure how viable it would be overall but would need someone with more knowledge than me in cpacker to get it going.

NinjaSpectre:
The archive will extract to whatever path the user enters.  The simple method I gave above only does that though... there are no "extras" (uninstaller, registry keys, start menu entries, etc.).  It is simply an archive extractor.
[...]

PR #1437, opened 2021-05-18
Use %appdata% on windows.

JonathanILevi:
When installing Empty Epsilon in the system Program Files folder (like #1003 would) it requires administrator privileges to write to the local folder.  Using the appdata folder also makes `options.ini` and such user specific, which is a good thing.

I have not tested this, if someone who already has a Windows development environment setup would do so, I would appreciate it.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -107,10 +107,11 @@ int main(int argc, char** argv)
 #ifdef CONFIG_DIR
     PreferencesManager::load(CONFIG_DIR "options.ini");
 #endif
+    PreferencesManager::load("options.ini");
     if (getenv("HOME"))
         PreferencesManager::load(string(getenv("HOME")) + "/.emptyepsilon/options.ini");
-    else
-        PreferencesManager::load("options.ini");
+    if (getenv("APPDATA"))
+        PreferencesManager::load(string(getenv("APPDATA")) + "/emptyepsilon/options.ini");
 
     for(int n=1; n<argc; n++)
     {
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -362,6 +374,14 @@ int main(int argc, char** argv)
             mkdir((string(getenv("HOME")) + "/.emptyepsilon").c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
 #endif
             PreferencesManager::save(string(getenv("HOME")) + "/.emptyepsilon/options.ini");
+        }else if (getenv("APPDATA"))
+        {
+#ifdef _WIN32
+            mkdir((string(getenv("APPDATA")) + "/emptyepsilon").c_str());
+#else
+            mkdir((string(getenv("APPDATA")) + "/emptyepsilon").c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
```
gcask, line 382: This feels like happy-go-lucky copypasta: which platform is this supposed to handle?

I don't think APPDATA really is a thing to worry about on anything but windows :)

gcask:
This is incomplete btw - there are references to `HOME` in `src/menus/optionsMenu.cpp` as well as `src/scriptDataStorage.cpp` that needs to be fixed up too (and tested).

gcask (changes requested):

src/main.cpp
JonathanILevi: Yeah, copypasta is what it was.  A more eloquent solution would be ideal.

src/main.cpp
```diff
@@ -107,10 +107,11 @@ int main(int argc, char** argv)
 #ifdef CONFIG_DIR
     PreferencesManager::load(CONFIG_DIR "options.ini");
 #endif
+    PreferencesManager::load("options.ini");
     if (getenv("HOME"))
         PreferencesManager::load(string(getenv("HOME")) + "/.emptyepsilon/options.ini");
-    else
-        PreferencesManager::load("options.ini");
+    if (getenv("APPDATA"))
```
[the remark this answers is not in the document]
JonathanILevi: Yeah, copypasta... `HOME` is also not a thing on Windows either...  I just extended the current solution rather than re-implementing a better one.

JonathanILevi:
> This is incomplete btw - there are references to `HOME` in `src/menus/optionsMenu.cpp` as well as `src/scriptDataStorage.cpp` that needs to be fixed up too (and tested).

Shoot, I got lazy and didn't look, I can add those.

PR #1438, opened 2021-05-19
macOS and win packages

gcask:
This provides the basics for generate DMG packages on macOS (the 'app bundle'... thingy) as well as generating Windows MSI installer via WIX (from #1003 ).

# macOS
It was actually pretty close: the forced install prefix was actually messing up the DMG pipeline, and the deployment target had to be an environment variable instead of a regular CMake one... and be defined *very* early.
Minor tweak relying on the `TARGET_BUNDLE_CONTENT_DIR` target property.
Lastly, the homebrew sfml version isn't really portable - I've only had success with the Frameworks version that SFML ships.
Should this PR pass, I'll update the wiki documentation with the new instructions.

# Windows
This one was a lot more painful that it should be, but it is working.
WIX has a lot of... needs you need to accomodate: the major version number cannot be greater than 256 (a wix limitation, not msi nor windows nor microsoft), so some hoop jumping required there.
There was some extra variables to set up to get a nice clean MSI, but now it is there.
Moreover, it does allow for different versions to be installed side-by-side.
The current caveat is if you install under the standard ProgramFiles tree, you will need to run as a privileged user - but this is getting addressed in #1437.
It could be prettier, but hopefully in a good enough shape to be used/redistributed by users who really want an installer - all that's required is having [WIX](https://wixtoolset.org/) installed on the dev machine (it is part of the standard git actions provisioning), and pass `CPACK_GENERATOR=WIX`.

It does not break the ZIP generation, and in fact you can even generate both at the same time.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -62,7 +62,7 @@ jobs:
         run: |
           mkdir -p _build_macos
           cd _build_macos
-          cmake ../EmptyEpsilon -G Ninja -DSERIOUS_PROTON_DIR=../SeriousProton
+          cmake ../EmptyEpsilon -G Ninja -DSERIOUS_PROTON_DIR=../SeriousProton -DCMAKE_INSTALL_PREFIX=.
           ninja
           ninja install
   windows:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,8 @@
 # Set minimum target macOS version: https://cmake.org/cmake/help/v3.8/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html#variable:CMAKE_OSX_DEPLOYMENT_TARGET
-# SFML targets 10.7, recent XCode version complains about anything below 10.9 (Mavericks, Oct. 2013).
+# SFML targets 10.7, however EE targets C++17 features only available in 10.15+ (Catalina, June 2019).
 # Must be first!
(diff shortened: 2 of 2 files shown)


gcask:
@JonathanILevi fyi.

JonathanILevi:
WiX is a Windows tool, is it not?  EE is cross compiled on a Linux machine, maybe `msitools`' reimplementation of WiX will work?

JonathanILevi:
This is a solution for #1003.

gcask:
> WiX is a Windows tool, is it not?

Yes, although that shouldn't be a problem for you right? If you want a windows installer, you probably run a windows system :)

> EE is cross compiled on a Linux machine, maybe `msitools`' reimplementation of WiX will work?

True for the current official releases.
But I will go on a limb and say that I don't think official release will care to maintain windows installers for a while. If ever.


See this like the macOS packages: provided as a convenience for some users, but not part of the "official" releases in any capacity (again, wild assumption)

JonathanILevi:
Why wouldn't the official release have an installer?!

daid:
> Why wouldn't the official release have an installer?!

Because they get cross compiler on linux by my CI server.

Maybe it is possible to run WIX in Wine and build the installer that way. I've done that before with NSIS to build software releases on windows.

JonathanILevi:
Yeah, that would be a reason.  I already have a solution which works on Linux though #1003.

PR #1439, opened 2021-05-19
Line up directory prefixes between package/source-tree install

gcask:
Because EE burns absolute path on compilation, we need to take some care ensuring they all line up during packaging.

This comes after noticing that leaving an empty install prefix would try to load from `/usr/local/...` while the package would be installed under `/usr/...`.

This comes from CMake having different logic handling between install-from-build-tree and install-from-package.

It also lines up the hardcoded defines with the GNU Install dirs, one less assumption.

This aims at ensuring all the root paths line up, and limit surprises.


gcask:
I've tested this building a debian package:
  * If you do not set `CMAKE_INSTALL_PREFIX`, the package will target `/usr/local` (CMake's default)
  * If you set `CMAKE_INSTALL_PREFIX=/usr`, the package will target `/usr` accordingly.

Both packages installed and ran successfully.

PR #1440, opened 2021-05-19
Use resetGLStates

gcask:
The current pop/push generate warnings (because they're backwards), as well as issues on some graphics hardware (older gen GeForce among other).

We have been checking this with @StarryWisdom & friends - no issues reported.

It actually fixed a few glitches on some graphics card.

Note that a clever mind actually [suggested exactly this a while back](https://github.com/daid/EmptyEpsilon/commit/3f4524bc389be38fc03c35e8e3b64377db4e414e#commitcomment-34589720)...


daid:
Most likely this will cause some new issues :D SFML is the gift that keeps on giving (in a bad way)

PR #1441, opened 2021-05-20
#1162: Display reputation and clock in comms window header

### LINKED ISSUE #1162: reputation can be obscured via comms
On a non wisdescreen aspect ratio the reputation number if covered up via open comms windows.

It is annoying as often  comms costs reputation and not being able to see how much you have is awkward.

at a ratio of 4:3 it is covered, at 16:9 it is visable, I did not check when it is first visable.

This may be as simple as left justifying the text - a majority of the label is visable just not the number that wants to be seen.

gcask:
Result:
image

This fixes #1162 by ensuring the information is always on-screen.

Note that I also rolled back the workaround from #1217, which only worked for the standard comms window.

### DIFF
--- a/scripts/comms_station.lua
+++ b/scripts/comms_station.lua
@@ -83,14 +83,12 @@ end
 -- @tparam SpaceStation comms_target
 function commsStationDocked(comms_source, comms_target)
     local message
-
     if comms_source:isFriendly(comms_target) then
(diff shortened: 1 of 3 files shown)


gcask:
Closing this after discussion on discord.
see #1442 for the continuation.

PR #1445, opened 2021-05-20
Ubuntu 18.04 / SFML 2.4 compatibility

gcask:
It has come to my attention that the build is broken on Ubuntu 18.04 (LTS).

This PR fixes the build, as well as the stencil-based radar on SFML 2.4 (since this version does not support creating stencil-backed render texture).

Note that since it means handling desktop, and desktop means 3D screen, there are the proper fixes there to handle the wide screen 3D viewport on the side of the radar.

### DIFF
--- a/src/gui/gui2_element.cpp
+++ b/src/gui/gui2_element.cpp
@@ -272,12 +272,14 @@ void GuiElement::updateRect(sf::FloatRect parent_rect)
 [[nodiscard]]
 bool GuiElement::adjustRenderTexture(sf::RenderTexture& texture)
 {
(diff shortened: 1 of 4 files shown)


gcask:
Related compatibility change in daid/SeriousProton#114

daid:
> Alternatively, I would be more than happy to ditch all this and just bump the SFML requirement to 2.5 in CMake directly 🙃

I'm considering it, as this doesn't become any more pretty...

gcask:
SFML 2.4 compatibility has been addressed by daid/SeriousProton#117.

The scissor fix has been split in #1447 .

PR #1446, opened 2021-05-20
#825 Configurable energy consumption rates

### LINKED ISSUE #825: Feature request: adjustable coolant flow speed
When engineering is managing coolant, it flows at a particular speed between systems as the level is adjusted. Can that flow speed be configurable per ship? Also, can that adjustment be made accessible to Lua?

Possible applications include making larger ships have slower flows and smaller ships have faster flows. Script writers could have damage slow the coolant flow or have refit rewards speed up the coolant flow.

Just an idea

gcask:
This fixes #825 by providing customizable energy consumption rates per type per subsystem per ship.

This means that heat, coolant, and power consumption can be configured at will on a ship:
image

And ditto for shields an warp drive:
image

(bonus: the fields will hide themselves if the ship has no warp or shield).

Everything is exposed to lua scripts and values are uncapped for maximum ~wreckage~ tweakability.

So yes, if you were wondering, you can have systems that produces energy, a warp drive can be set to recharge the ship, coolant can heat and heat can cool.

Note that all values are viewed from the _consumer_ perspective: a positive value will drain energy ("gives to" the consumer, "takes from" the producer), a negative value will produce energy ("takes from" the consumer, "gives to" the producer).

### DIFF
--- a/src/gameStateLogger.cpp
+++ b/src/gameStateLogger.cpp
@@ -389,7 +389,7 @@ void GameStateLogger::writeShipEntry(JSONGenerator& json, P<SpaceShip> ship)
                 system.write("power_rate_per_second", ship->systems[n].power_rate_per_second);
                 system.write("power_request", ship->systems[n].power_request);
                 system.write("heat", ship->systems[n].heat_level);
-                system.write("heatup_rate_per_second", ship->systems[n].heatup_rate_per_second);
+                system.write("heat_rate_per_second", ship->systems[n].heat_rate_per_second);
                 system.write("coolant_level", ship->systems[n].coolant_level);
                 system.write("coolant_rate_per_second", ship->systems[n].coolant_rate_per_second);
                 system.write("coolant_request", ship->systems[n].coolant_request);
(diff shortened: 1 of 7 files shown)


Xansta:
This is great! Could I have a Lua script example? Also, will the Lua functions show up on the script reference when it gets generated (I assume so, just want to verify)?

gcask:
> Could I have a Lua script example?

Sure! So now the coolant, heat and power rate are now tweakable per system.
They're exposed on the `Spaceship` objects through functions aptly named `getSystem*Rate` and `setSystem*Rate` (subsistute `*` with `Heat` or `Coolant` or `Power`).

They should work just like any other ship system function, ie `some_ship:setSystemCoolantRate("reactor", -1.f)`.

As mentioned in the PR description, the rate is measured from the consumer (ie the system): positive drains from the ship's energy banks (it "adds to" the system's), negative would turn the system into a producer ("giving back" to the ship's energy banks). And yes, that does mean coolant can heat up if you want it to.

Additionally, I've added functions to customize a ship's shield energy and warp power consumption: those are only available on a `PlayerSpaceship` object: `setEnergyShieldUsePerSecond(<new_value>)` / `getEnergyShieldUsePerSecond()` and `setEnergyWarpPerSecond(<new_value>)` / `getEnergyWarpPerSecond()`.

Note that for the shields it's all of them, you can't configure power consumption per shield.

Same as above - negative values should turn the shields / warp engines into dynamos.


> will the Lua functions show up on the script reference when it gets generated (I assume so, just want to verify)?

I assume so too! Just give me a ping if that's not the case :)

daid:
Yes, it will show up in the reference, the reference is build by just scanning the code.

PR #1447, opened 2021-05-26
Scissorbox for the radar when not using a RT

gcask:
When not using a Render Texture, it can lead to issues with the side main screen because the gl commands (ie clear) are not constrained to the area of the radar.

Using a scissor box fixes this.

split from #1445


daid:
Never knew this feature existed... do we actually still need the render textures?

PR #1453, opened 2021-05-30
Prevent Nukes and EMPs detonating twice on clients

StarryWisdom:
This was observed causing different clients to disagree on if a ship
was destroyed due to it having taken damage twice.


daid:
Great catch!

PR #1454, opened 2021-05-30
fix: Hotkey menu deleted joystick button mapping.

aBlueShadow:
As the (very cool) new hotkey menu is not aware of joystick mappings, joystick button codes added in options.ini got deleted in the mapping menu, as it assumed a wrong keycode.

This PR fixes the issue by checking for joystick button codes in addition to sfml key codes..

This is just a bugfix, joystick button mappings still have to be added manually in options.ini

### DIFF
--- a/src/menus/hotkeyMenu.cpp
+++ b/src/menus/hotkeyMenu.cpp
@@ -1,6 +1,7 @@
 #include <i18n.h>
 #include "engine.h"
 #include "hotkeyMenu.h"
+#include <regex>
 
 #include "gui/hotkeyBinder.h"
 #include "gui/gui2_autolayout.h"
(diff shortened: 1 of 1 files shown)


src/menus/hotkeyMenu.cpp
```diff
@@ -204,14 +205,15 @@ void HotkeyMenu::saveHotkeys()
     }
 
     // Read in all TextEntry values and update hotkeys
+    std::regex buttonIdExpression ("(\\[J([0-7])\\])?\\[B([0-9]|[12][0-9]|3[0-1])\\]"); // matches [J0] - [J7] (optional) and  [B0] - [B31]
```
daid, line 208: Raw string literals can most likely help making this a bit more readable: <https://www.geeksforgeeks.org/raw-string-literal-c/>

And I think  just testing for `[JB][0-9]+` might be good enough and simpler to read?

aBlueShadow: Interesting. Did not know about raw string literals, although in that case I am not sure if getting rid of two blackslashes helps that much. Anyways, you might be right about simplifying the pattern, should be be good enough for now. So I went with that option. Wether it is really more readable now or not, might be a matter of taste though ;-)

PR #1457, opened 2021-06-03
Colorize output when ninja is used

aBlueShadow:
While ninja is faster than make, the output is not colorized per default, that can make reading error messages more difficult.
This commit incorporates the workaround mentioned in the FAQ:
https://github.com/ninja-build/ninja/wiki/FAQ

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ endif()
 message(STATUS "EmptyEpsilon Version = ${PROJECT_VERSION}")
 
 # Enable colored output with Ninja
-if(CMAKE_GENERATOR MATCHES "Ninja")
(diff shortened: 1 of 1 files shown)


aBlueShadow:
One CI Job failed, but It does not look like it is really related to my PR. Is there a way to restart a job, in case it was some "hickup"?

aBlueShadow:
Ah, just saw all jobs were successful now! So it seems it really was just some hickup :-)

CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ endif()
 
 message(STATUS "EmptyEpsilon Version = ${PROJECT_VERSION}")
 
+# Enable colored output with Ninja
+if(CMAKE_GENERATOR MATCHES "Ninja")
```
daid, line 53: Should also have a check to see if the compiler is GCC, else you could pass this option to msvc.

aBlueShadow: Makes sense. Done.

PR #1459, opened 2021-06-03
Move GNUInstallDirs include after project/language are set

gcask:
Recent version of CMake (3.20) issues a warning that GNUInstallDirs cannot set some variable because it is called too soon (before languages are set).

Moving the include slightly later.


gcask:
The warning:
```
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.20/Modules/GNUInstallDirs.cmake:236 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:12 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

PR #1461, opened 2021-06-12
Radar fixes

gcask:
* Ship's short range is always visible
* Blackout when ship is inside nebula
* Visible objects correctly drawn

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -318,7 +318,6 @@ void GuiRadarView::drawNoneFriendlyBlockedAreas(sf::RenderTarget& window)
     {
         float scale = std::min(rect.width, rect.height) / 2.0f / distance;
 
-        float r = 5000.0 * scale;
         sf::CircleShape circle(0.f, 50);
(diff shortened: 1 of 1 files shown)


src/screenComponents/radarView.cpp
```diff
@@ -756,23 +776,10 @@ void GuiRadarView::drawObjects(sf::RenderTarget& window)
...
-    glStencilFunc(GL_EQUAL, as_mask(RadarStencil::InBoundsAndVisible), as_mask(RadarStencil::InBoundsAndVisible));
-    for (auto obj : visible_objects)
-    {
-        if (obj->canHideInNebula())
-        {
-            draw_object(obj);
-        }
-    }
-
-    // Second, draw all objects that can't hide.
```
gcask, line 770: This bit was interesting: the previous iteration had a complex path of using the 'alpha window' or the 'background' one depending on whether the object could hide in a nebula or not.

My initial understanding was that I had to honor that, but from a closer look, it appears the filtering is already done before (correctly handling friendlies etc) so... Looks like we can get away by just drawing the list of visible objects as-is.

PR #1474, opened 2021-06-26
CMake-driven config header

gcask:
Requires daid/SeriousProton#125 (there's a chance it'll compile, but the window title will be broken)


daid:
Could you check the conflict? I'm not 100% sure what happened there with the target name.

CMakeLists.txt
```diff
@@ -59,6 +59,8 @@ endif()
 # SeriousProton
 # Use SP provided json library.
 set(SERIOUSPROTON_WITH_JSON ON CACHE BOOL "" FORCE)
+# Set window title for SP.
+set(WINDOW_TITLE "${PROJECT_NAME}" CACHE STRING "" FORCE)
```
daid, line 63: Window title is no longer part of SP input, so can be removed. Yay! less code!

gcask:
looks like the merge was a little lost because now the defines apply only to EE, so they don't need to be applied on SP itself :) (I checked the defines by usage, they're all from EE cpp files)

Issue #1320, opened 2021-02-03
[fixed-pipeline] Remove transform fixed functions


gcask:
This essentially translates into bringing in [GLM](https://github.com/g-truc/glm), and replacing all the matrix transforms with it.

We can do it in a two-step process:

- [ ] Remove usage of predefined uniforms from all the shaders (passing the matrices via uniforms)
- [ ] Replace fixed functions with glm counterparts.

For the latter, there's not really any valid halfway point - it's mostly an all or nothing type deal (we'll see if there's a way to chunk it up some more)

Having the model and view matrices separate should help the code in many circumstances though.

PR #1477, opened 2021-06-27
getModelTransform() on space objects

gcask:
Part of #1320 : each space object is able to provide its (current) model matrix (transform in world space).

### DIFF
--- a/src/spaceObjects/beamEffect.cpp
+++ b/src/spaceObjects/beamEffect.cpp
@@ -192,6 +192,6 @@ void BeamEffect::setTarget(P<SpaceObject> target, sf::Vector2f hitLocation)
 
 glm::mat4 BeamEffect::getModelMatrix() const
 {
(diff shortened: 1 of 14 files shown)


src/spaceObjects/beamEffect.cpp
```diff
@@ -187,3 +189,9 @@ void BeamEffect::setTarget(P<SpaceObject> target, sf::Vector2f hitLocation)
     sf::Vector3f targetPos(target->getPosition().x, target->getPosition().y, 0);
     hitNormal = sf::normalize(targetPos - hitPos);
 }
+
+glm::mat4 BeamEffect::getModelMatrix() const
+{
+    auto position = const_cast<BeamEffect*>(this)->getPosition();
```
daid, line 195: guess the const cast is no longer needed now? :)

gcask: ha, indeed!

PR #1478, opened 2021-06-27
Remove a forgotten left over from the move of CMake MinGW Crash Logger code to SeriousProton project

FaramosCZ:


FaramosCZ:
Followup to https://github.com/daid/SeriousProton/pull/128

PR #1479, opened 2021-06-27
CMake: Explicitly set OpenGL version to LEGACY to deal with the CMake policy CMP0072

FaramosCZ:


FaramosCZ:
Solves the warning texts as seen here:
https://github.com/daid/EmptyEpsilon/runs/2927023349#step:5:23

```
CMake Warning (dev) at /usr/share/cmake/Modules/FindOpenGL.cmake:315 (message):
  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib64/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib64/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib64/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.
Call Stack (most recent call first):
  CMakeLists.txt:57 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found OpenGL: /usr/lib64/libOpenGL.so   
```

FaramosCZ:
I successful compiled EE with ```set(OpenGL_GL_PREFERENCE GLVND)``` too, but in this commit, I only want to resolve the warning.

Let the switch to the GLVND be decided & tested by someone else.

daid:
GLVND seems to be "vendor". Which I guess are the OpenGL libraries from nvidia or amd? As those might not always be available, defaulting to LEGACY sounds better.

gcask:
Yeah Legacy is fine, but not for the reason you think :)

It's more about how the context is retrieved (through glx or egl), not really tied to the manufacturer. Or not directly at least.

See the [FindOpenGL CMake docs](https://cmake.org/cmake/help/latest/module/FindOpenGL.html#linux-specific) for more info.

It impacts mostly systems using EGL (eg RPi) and requiring a GL3+ context.

For EE, either way we go is fine really, since it'll fallback on legacy if glvnd is not available.

PR #1480, opened 2021-06-28
Remove FindSFML.cmake form the EmptyEpsilon project

FaramosCZ:
- An updated version of this file is present in the SeriousProton project
 - This file is not called in the EmptyEpsilon project, but the one from SeriousProton project is called in it


FaramosCZ:
This PR requires https://github.com/daid/SeriousProton/pull/130 to be merged

PR #1482, opened 2021-06-29
Allow mod name without slash

aBlueShadow:
As adding a "/" at the end of the modfolder name isn't obvious, this adds an alternative syntax for the mod option. So you just have to enter `mod=modname` instead of `mod=modname/` in options.ini

Similar to f5cc8a5a2260cb1f797cbc66509373d95cc1c273 from #620, but backwards compatible, so the old syntax is still valid.


daid:
Pretty sure this does nothing, and was fixed a while back in SeriousProton:
https://github.com/daid/SeriousProton/blob/master/src/resources.cpp#L111

PR #1492, opened 2021-07-26
Index property for custom functions

aBlueShadow:
The new index argument for the add functions is optional, if empty, the index property will be set to zero.
As the entries are sorted by index, this is now a way to have better control over the display order  of custom functions.
This is especially useful if you want to update information, so you can avoid the last updated element jumping to the bottom.


daid:
Looks good, except for the name. `index` normally indicates something unique. I would name it `order` instead.

aBlueShadow:
Good Point. Done.

PR #1494, opened 2021-07-28
Relay tutorial: Removed comms capability of dummy ships, modified short range radar presentation.

hemmond:
* Comms window could not be closed because tutorial window was over the comms's cancel button if user hailed dummy ship instead of sending a probe.
* Removed 5U sensor range in tutorial text (since now it can be different per ship) and changed short range of one ship to demonstrate different sensor ranges.


aBlueShadow:
Keep in mind that the Relay part of the complete tutorial is still unchanged.

hemmond:
> 
> 
> Note: the Relay part of the complete tutorial is still unchanged.

Oh, thanks for notifying me. For some reason I thought that "all tutorials" is just concatenation of tutorials. I'll look into it tomorrow.

PR #1512, opened 2021-09-15
Add 43 _() scenario_81_pvp.lua

muerteFR:
Please @daid look at line 126. I'm not sure this is the best writing.
Thanks

### DIFF
--- a/scripts/scenario_81_pvp.lua
+++ b/scripts/scenario_81_pvp.lua
@@ -87,20 +87,20 @@ function init()
     -- Brief the players
     shipyard_human:sendCommsMessage(
         gallipoli,
-        [[Captain, it seems that the Kraylor are moving to take the Shangri-La station in sector F5!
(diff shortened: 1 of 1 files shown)


daid:
Not sure what would be wrong with line 126? But I'm also no native English speaker?

muerteFR:
> Not sure what would be wrong with line 126? But I'm also no native English speaker?
L226 I wanted to say. I was talking about the introduction of tags _().
string.format(args.request .. _(" (%d rep each)"), price),

daid:
Guess `string.format(args.request .. " " .. _("(%d rep each)"), price),` is slightly better, as the leading space isn't really part of the language.

muerteFR:
> Guess `string.format(args.request .. " " .. _("(%d rep each)"), price),` is slightly better, as the leading space isn't really part of the language.

Thanks, it's done

PR #1523, opened 2021-09-27
update some fr.po files

muerteFR:


daid:
Empty entries are removed in the dickens scenario?

muerteFR:
> Empty entries are removed in the dickens scenario?

No, like I said here https://github.com/daid/EmptyEpsilon/pull/1520 these are entries that are unnecessary for the story and after seeing with Xansta I removed them. (it served as a test)

muerteFR:
> > Empty entries are removed in the dickens scenario?
> 
> No, like I said here #1520 these are entries that are unnecessary for the story and after seeing with Xansta I removed them. (it served as a test)

And now I update the fr.po as we saw together the last time thanks to (python3 update_scenario_locale.py) :-)

PR #1525, opened 2021-09-27
Add german translations for basic scenario

PET2001:
form a german translation to that file


PET2001:
comment   this is not testet true the end in the gameplay   - just started it see the german text on the screen then put it up   
schoud be working
or   somebudy schould play it true   the end and in every condition -  maybe a group test

PET2001:
08-birth of atlantis also translatet to a german po   only short testet works

PR #1556, opened 2021-10-19
Fix errant if in Mesh::render()

gcask:
Would have devastating effects with meshes with over 64k vertices :)


daid:
surprising that we don't get an indentation warning...

PR #1589, opened 2021-10-31
Support for modulating the illumination map.

gcask:
With this, we  can do some tricks around the colors from blinking effects to color shifting etc...


gcask:
Requires daid/SeriousProton#200 for proper support (will compile without, but won't correctly load the info).

PR #1593, opened 2021-11-01
Cinematic and topdown keybindings

gcask:
* Add keybindings for cinematic and top-down views.
* Updated top-down controls to be framerate independent.


Xansta:
Will the cinematic view be in the next release? I have not seen it yet in the pre-release

daid:
> Will the cinematic view be in the next release? I have not seen it yet in the pre-release

Yes, I added it back after that build, simply missed that it was missing

PR #1595, opened 2021-11-01
Radar draw ordering

gcask:
This fixes an issue @Xansta and @StarryWisdom identified in the current release, where objects may have varying intensity because of inconsistent draw order.

This ensures all 'foreground' objects are always drawn on top of background ones, and offer the possibility to get more fine-grained control if required.

(Note that here, foreground/background is really a simple visual aspects and has no gameplay implications)


src/screenComponents/radarView.cpp
```diff
@@ -650,6 +650,14 @@ void GuiRadarView::drawObjects(sp::RenderTarget& renderer)
         break;
     }
 
+    std::vector objects_to_draw(std::begin(visible_objects), std::end(visible_objects));
+    std::sort(std::begin(objects_to_draw), std::end(objects_to_draw), [](const auto& lhs, const auto& rhs)
+    {
+        const auto lhsLayer = lhs->getRadarLayer();
+        const auto rhsLayer = rhs->getRadarLayer();
+        return lhsLayer < rhsLayer || (lhsLayer == rhsLayer && lhs->canHideInNebula() && !rhs->canHideInNebula());
```
gcask, line 658: First order by layers, and if in the same layers, put objects that can hide in nebulas "lower".

PR #1599, opened 2021-11-02
Ability to directly start a scenario and server creation screen

aBlueShadow:
This introduces a new preferencemanager option "server_scenerio" (not entirely sure about the name, maybe something like "quickstart" is better?)
It works similar to the headless version, but with a graphical window (does not query _startpaused_ though, maybe that could be added.)

The reason for this PR was that testing testing small adjustments/variations in code/scenarios/assets require multiple restarts, which can get tedious, especially with the new server gui. Often you can get around that by a combination of _headless_ and _autoconnect_ or a regular client, but e.G. you cannot enter the the GM screen that way.

### DIFF
--- a/src/menus/serverCreationScreen.cpp
+++ b/src/menus/serverCreationScreen.cpp
@@ -15,6 +15,7 @@
 #include "gui/gui2_panel.h"
 #include "gui/gui2_scrolltext.h"
 #include "scenarioInfo.h"
+#include "menus/mainMenus.h"
 #include "main.h"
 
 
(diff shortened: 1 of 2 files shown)


src/menus/serverCreationScreen.cpp
```diff
@@ -77,7 +78,10 @@ ServerSetupScreen::ServerSetupScreen()
     // Close server button.
     (new GuiButton(this, "CLOSE_SERVER", tr("Close"), [this]() {
         destroy();
-        returnToMainMenu();
+        if (PreferencesManager::get("server_scenario")=="")
```
daid, line 81: I'm not liking this hack. I think this feature is easier implemented here:
https://github.com/daid/EmptyEpsilon/blob/master/src/main.cpp#L331
Then you won't need this hack, and only open the scenario once on startup.

aBlueShadow: Okay, will change it. Thought it will be nice to have all those startup-related options at one place, but requiring that hack then kinda defeats the purpose. Doing that before the initial main menu call would also prioritise that feature when multiple options are set, but I guess that won't matter much.

PR #1604, opened 2021-11-05
Create scenario_55_defenderHunter.en.po

muerteFR:


daid:
I can see by the description that this was done before https://github.com/daid/EmptyEpsilon/commit/7405254af1ebf77f3bddb39e8eb8252e0392b667 which fixes the translation of descriptions. But they where generated wrong in the translation file (without newlines)

muerteFR:
> I can see by the description that this was done before [7405254](https://github.com/daid/EmptyEpsilon/commit/7405254af1ebf77f3bddb39e8eb8252e0392b667) which fixes the translation of descriptions. But they where generated wrong in the translation file (without newlines)

I tried again, still with update_scenario_locale.py but the result is the same. This does not matter because normally if you update from your repository, it should add the newlines in the description (like scenario_81_pvp.en.po for which I had not added the description because from the server it adds it when you refresh). The rest of this file looks correct with the newlines.

muerteFR:
@daid, do I have to do something else? Do I have to try again?

muerteFR:
I just re-compiled and this time it seems to have worked for the description. I don't know at all why though.

daid:
Magic 🧙

PR #1609, opened 2021-11-10
Update styles of scripting reference docs

jozan:
I need to read this documentation quite a lot so I removed some of the clutter and now the focus is on text. Probably it still retains that magical EmptyEpsilon feel I hope. I've been using this style override for a while now and decided to share this with everyone. What do you think @daid?

It looks something like this on Firefox 94.0.1 on Windows 10.

image

Old style for reference:
image


daid:
Looks good! Much more readable!

CI is failing due to python3 compatibility (mixed tabs and spaces is no longer allowed in python 3)

jozan:
Oh, which one do you prefer? Tabs or spaces? Currently it has both all over the file so I can fix them in one go.

daid:
I preferrer spaces myself.

jozan:
I ran it through black, which is python auto formatter. Not sure if you like the result of that but it was the quickest way to make it consistent. 乁( ◔ ౪◔)ㄏ

https://black.vercel.app

daid:
Looking good :) thanks

PR #1615, opened 2021-11-13
20_training1.lua Standardization of translation contexts

muerteFR:
- The tag was incorrectly positioned preventing translation of the description.
- Moved to the Basic category.

### DIFF
--- a/scripts/scenario_20_training1.lua
+++ b/scripts/scenario_20_training1.lua
@@ -75,7 +75,7 @@ end
 function commsInstr()
     if not instr1 and timer > 8.0 then
         instr1 = true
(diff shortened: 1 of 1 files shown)


daid:
I fixed a bug in the scenario description translation yesterday, the move of the `type` downwards should no longer be needed.

muerteFR:
Yes I saw but didn't have time to test. I made a commit of contexts here yesterday so as not forgetting, by the time I'm finishing my big current PR Gaps and big standardization of contexts. Then I will delete it.

muerteFR:
> I fixed a bug in the scenario description translation yesterday, the move of the `type` downwards should no longer be needed.

I put "type" back in its place and standardized the contexts.

PR #1618, opened 2021-11-15
Tweaks on the Discord icon of the community page

l-delpino:
This just ensures it displays properly (tested on Firefox, desktop version), plus it roughly aligns it to be inline with the text.


daid:
Thanks! I didnt have time yesterday to fix it properly.

PR #1621, opened 2021-12-01
03_waves.lua Update description translation & Standardization contexts

muerteFR:
I followed what you did with Waves for description translation.

### DIFF
--- a/scripts/scenario_03_waves.lua
+++ b/scripts/scenario_03_waves.lua
@@ -131,9 +131,9 @@ function spawnWave()
 
     -- Calculate score of wave
     local totalScoreRequirement  -- actually: remainingScoreRequirement
-    if getScenarioVariation() == "Hard" then
+    if getScenarioSettings() == "Hard" then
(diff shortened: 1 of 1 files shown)


scripts/scenario_03_waves.lua
```diff
@@ -3,8 +3,10 @@
 ---
 --- Spawn the player ships you want. The strength of enemy ships is independent of the number and type of player ships.
 -- Type: Basic
--- Variation[Hard]: Difficulty starts at wave 5 and increases by 1.5 after the players defeat each wave. Players are overwhelmed more quickly, leading to shorter games.
```
daid, line 6: If variations are replaced with settings, I would expect the call to `getScenarioVariation` to be replaced as well.

muerteFR:
even if it does not seem essential in this case where there is only this "Setting".

daid:
Well, the function doesn't know there is only 1 setting, so it requires the scenario setting name.

Issue #514, opened 2018-01-08
Problem with the name of the sector


tdelc:
With the last release, i notice a problem. You can see it when you scroll down the windows starting from the 0,0 coordinate.

Sectors A0 to A99 appear two times in a same row just one after the other.

![image](https://user-images.githubusercontent.com/13868149/34662502-71d5bc04-f451-11e7-8986-8038bac75a92.png)

oznogon:
PR #512 would probably change this behavior.

csibbitt:
"Sectors A0 to A99 appear two times in a same row just one after the other"

Yup. There is a bit more to this issue, too. We started to experiment with larger scale maps and immediately ran into this.

The reported problem is here: https://github.com/daid/EmptyEpsilon/blob/master/src/gameGlobalInfo.cpp#L232

Positive X simply counts up and seems fine for practical use. Negative X counts down from 99 to 0 then into the negatives, so overlaps with the Positive axis immediately and for 100 sectors. 

Here is a screenshot of the existing origin at F5 and the imposter origin, also called F5 - notice the numeric part goes negative instead of to 99 on the imposter
<img src="https://user-images.githubusercontent.com/5091372/146269928-219cf407-af8c-491d-9ecd-c44e57ec35df.png" alt="Old Origin" width="250"/>&nbsp;&nbsp;<img src="https://user-images.githubusercontent.com/5091372/146269982-0de7325c-3b69-4e94-860f-2dc7c95216d4.png" alt="Old Fake Origin" width="250"/>

This appears to have been a stylistic decision (preferring 99 to -1), but really limits the unambiguous play area around the origin.

There are also problems in the Y direction. Positive Y counts up without bound, through the capital letters, then [\]^_` then into lowercase letters and beyond, wrapping around the ascii table making very interesting results. Negative Y counts zz, zy, zw, ... zc, zb, y{, yz, yy, ...

Here is a screenshot of the existing glitches in the North and South
<img src="https://user-images.githubusercontent.com/5091372/146270125-10b1a998-43a8-45a0-85f9-47f6648effe4.png" alt="Old North" width="250"/>&nbsp;&nbsp;<img src="https://user-images.githubusercontent.com/5091372/146270184-05781242-bfa8-49d2-9859-b221196293a5.png" alt="Old South" width="250"/>

[...]

PR #1628, opened 2021-12-15
Improve sector naming

csibbitt:
Fixes https://github.com/daid/EmptyEpsilon/issues/514

"Sectors A0 to A99 appear two times in a same row just one after the other"

* Use negative numbers immediately when headed West instead of counting down
  from 99 first
* Standardizes the double-letter notation both North (lowercase same as
  existing behavior on master) and South (uppercase) which results in a much
  more usable play area
* I've reviewed all of the stock scenarios for any impact
  * Sector name in "Birth of the Atlantis" was the wrong case
  * As far as I can tell there are no other changes required
    * Most scenarios only use the unambiguous, non-glitched areas of the map

See the issue for more information.


csibbitt:
I have added another commit that contains a lua utility function call sectorToXY() which returns x,y coordinates for a named sector following the improved scheme laid out in this PR. Here is a test case with some screenshots:

```
require("utils.lua")
-- Test sectorToXY()
function init()
  -- Quadrants near origin
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("zx-3"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("zw4"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("D-3"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("E4"))

  -- Northern hemi-planes
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("xy-11"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("xx2"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("yd-11"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("ye2"))

  -- Southern hemi-planes
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("CY-1"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("CX12"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("DC-1"))
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("DD12"))
end
```
QuadrantSectorXYTest
NorthSectorXYTest
SouthSectorXYTest

aBlueShadow:
> Standardizes the double-letter notation both North (lowercase same as
existing behavior on master) and South (uppercase) which results in a much
more usable play area
However, this only works if font_bold is set to the new default (or another mixed-case font for that matter), as the old standard font is Uppercase only.

aBlueShadow:
Also, while I agree that the negative numberings seems logic, it is a bit inconsistent that the letters don't follow the same direction.
Although that might raise another question: In that case would  -A0 be adjacent to A0 or would we need a new origin sector like 00 to make it completly symmetrical?

Xansta:
I like the idea. Will the new release default to an upper and lower case font?

csibbitt:
> However, this only works if font_bold is set to the new default (or another mixed-case font for that matter), as the old standard font is Uppercase only.

It works on the current master branch. Is there a way to change the font from the default? I don't see it exposed in the options menu.

>Also, while I agree that the negative numberings seems logical, it is a bit inconsistent that the letters don't follow the same direction.
>Although that might raise another question: With negative letter, would -A0 be adjacent to A0 or would we need a new origin sector like ɛ​0 to make it completly symmetrical?

I'm not looking to raise any questions at all, this is not an aesthetic change, it's a bug fix. The negative numbers are already there, you just have to travel through duplicate-named sectors before reaching them. I'm not trying to add symmetry or enhance the style or feel of the game; heck, having a janky sector system that evolved over time and is "weird" can just be part of the lore AFAIC. This patch improves upon an obvious and highly limiting bug by changing as few things as possible and without introducing any new concepts whatsoever.  I think adding a center sector and redefining the origin would be a large change for little benefit in comparison. If the core team is motivated to re-vamp the sector system that's fine, but it's not my place to make that sort of creative change. In the meantime, it would help our scenario development plans to have a map that isn't bugged right near the start position.

csibbitt:
> I like the idea. Will the new release default to an upper and lower case font?

Currently the the master branch has upper/lowercase as shown in my screenshots. I'm hoping this patch will merge before someone decides to switch it back to uppercase-only :)

tdelc:
To give you some new ideas, this is the sector system made by Amir and used in our larp fork of EE. I didn't maintain version up to date, but I think that this feature could be added without a lot of work

https://user-images.githubusercontent.com/13868149/146396866-5d472aff-aad3-49ba-9b90-5bd26f23d76f.mp4

amir-arad:
#1628 512

csibbitt:
@tdelc I was looking at https://github.com/daid/EmptyEpsilon/pull/963 earlier, very nice work! I'm not trying to overhaul here, just fix. I'd be happy with either solution really.

aBlueShadow:
>>However, this only works if font_bold is set to the new default (or another mixed-case font for that matter), as the old standard font is Uppercase only.

>It works on the current master branch. Is there a way to change the font from the default? I don't see it exposed in the options menu.


it is the _font_bold_ option in options.ini in this case, and _font_regular_ for the mosts parts of the text.
For the linux version, the options on linux systems are stored in the user preference folder .emptyepsilon rather than the game directory like the windows version. So, a linux user that have EE installed previously will still have Uppercase if they don't actively change the font options.

> I'm not looking to raise any questions at all, this is not an aesthetic change, it's a bug fix. 

Fair enough. I was more like thinking aloud, like "if we do a potentially breaking change, we might consider if we want to change other things while we're at it." But thanks to _getSectorName()_, it's probably not that much of a breaking change anyways.

csibbitt:
@aBlueShadow Thanks for the explanation of the font preferences. I hadn't noticed options.ini before.

> a linux user that have EE installed previously will still have Uppercase if they don't actively change the font options

That is indeed a bit of a problem. All told, such users would STILL be in a better situation than before, with a much larger unambiguous play area, but it's certainly not ideal to have this option change the grid behavior. In the old code, there are actually 4 instances of the "F5" sector, and for those stuck in upper-case land, now there would only be two. :)

I hope this change is entirely non-breaking, and simply improves the situation without any visible change for players who have not already noticed the problems.

daid:
I know I've been silent, but I'm still alive ;-)

Let me first comment on a few basic things. The original sector labeling code never was made with "large scale maps" in mind. It was just a "it's good enough for now" kind of solution. So, addressing this is a really good thing! Thanks for putting in the effort.

About #963, I find that way too overcomplicated. It has a lot of customizable moving parts that need to be "just right" for things not to break in subtile ways, so that's why I never approved of it.


On this pull request. I think I rather see the `sectorToXY` function in C++, so it can be close to the function that makes a label from a position.
One other thing, I'm not sure if I like having lower and upper case letters. The old font only had uppercase, so it was "invisible" back then. But I'm not to concerned about that. I'm more concerned about consistent looks and player-to-player communication. `Player1: Sector U5. Player2: Upper or lower case U? Player1: Uh, lower case I guess?`
But I also don't have a better proposal...

csibbitt:
Thanks for the feedback, @daid . I'll take a shot at porting sectorToXY into the engine!

I think the consistency problem (notably the sticky linux default) can probably be taken care of with a release note. It may also intersect with this PR [https://github.com/daid/EmptyEpsilon/pull/1437] if they merge during the same release, so it requires a bit of care.

Luckily, the upper/lower equivalent sectors are actually very far apart from each other, so practically speaking it's unlikely to matter (and is still an improvement on the existing situation). I thought about it and decided to consider it a minor extra challenge when dealing with huge distances, but am certainly open to a better proposal should anyone else have one to add. I also think that if there is a change to be made there, it could come as a separate PR since it isn't this change that alters the font defaults. (ie. I didn't decide to use upper/lower, or even "reveal" the existing upper/lower with this PR, it happened elsewhere).

daid:
Excellent points :) I'm just merging this now, as it definitely makes things better.

csibbitt:
Thanks for the merge! I do have the port of sectorToXY() into C++ mostly complete, just need to finish up the equivalent testing and then I'll post it as a separate PR, coming soon!

PR #1632, opened 2021-12-20
Creation of scenario_49_allies.en.po & scenario_50_gaps.en.po

muerteFR:


daid:
Sorry that this took forever, I did this myself. I wanted to check out why the translation of settings wasn't working for a full 100% before I did this.

PR #1651, opened 2022-01-04
Port sectorToXY into c++

csibbitt:
>  I think I rather see the sectorToXY function in C++, so it can be close to the function that makes a label from a position . [1]

I put this through the same set of tests[2] as the original change.

[1] https://github.com/daid/EmptyEpsilon/pull/1628#issuecomment-1003970286
[2] https://github.com/daid/EmptyEpsilon/pull/1628#issuecomment-995412473

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -244,6 +244,56 @@ int getSectorName(lua_State* L)
 /// Return the sector name for the point with coordinates (x, y). Compare SpaceObject:getSectorName().
 REGISTER_SCRIPT_FUNCTION(getSectorName);
 
+glm::vec2 sectorToXY(string sector)
+{
+    constexpr float sector_size = 20000;
+    int x, y, intpart;
+
+    // Y axis is complicated
+    if(sector[0] >= char('A') && sector[1] >= char('A')){
+        // Case with two letters
+        char a1 = sector[0];
+        char a2 = sector[1];
+        try{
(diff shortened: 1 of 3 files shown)


src/gameGlobalInfo.cpp
```diff
@@ -244,6 +244,56 @@ int getSectorName(lua_State* L)
 /// Return the sector name for the point with coordinates (x, y). Compare SpaceObject:getSectorName().
 REGISTER_SCRIPT_FUNCTION(getSectorName);
 
+glm::vec2 sectorToXY(string sector)
+{
+    constexpr float sector_size = 20000;
+    int x, y, intpart;
+
+    // Y axis is complicated
+    if(sector[0] >= char('A') && sector[1] >= char('A')){
```
daid, line 253: This could crash if `sector.length() < 2`

csibbitt: Good catch, thanks! I'll try to get a fix in this evening.

csibbitt: Fix was tested with the following:
```
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("AA")):setCallSign("AA")
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("A")):setCallSign("A")
  PlayerSpaceship():setTemplate("Atlantis"):setPosition(sectorToXY("")):setCallSign("_blank_")
```

No crashes.

PR #1655, opened 2022-01-05
Translation update 100% scenario_49_allies.fr.po

muerteFR:


scripts/locale/scenario_49_allies.fr.po
```diff
@@ -0,0 +1,3547 @@
...
+msgctxt "buttonGM"
+msgid "Admiral"
+msgstr ""
+
+#: scripts/scenario_49_allies.lua:68
+msgctxt "buttonGM"
+msgid "Doomsday"
+msgstr ""
+
+#: scripts/scenario_49_allies.lua:408
```
csibbitt: I'm totally naïve about PO files, and poedit, etc, and was just wondering if/how these comments are auto-generated and whether anything depends on the line number mentioned in the comment? 

ie. If I were to edit the allies scenario so that this comms happened on another line, is there any need to update all the .po files as well? I'm assuming not, but wonder why they are mentioned in the first place

muerteFR: @daid and @Xansta will be able to answer you. I just take care of translating into French. ^^

aBlueShadow: Yeah, they are autogenerated, I think by the update_locale build target.
You assumed correctly. If the line number changes but not the translatable contend, the translation would still work.
However, they are still quite useful for translating. I use them every now and then, when I am unsure about the exact context of a translatable string. So I can look at the code (the poedit editor has a convenient function for this) which usually helps me to understand in what circumstance that text is used, so it gives me clues how I could translate it.

csibbitt: Thanks! So the line numbers could diverge, but would be fixed up during the build. Perfect, thanks!

PR #1658, opened 2022-01-15
Change icon installation path

NicolaSmaniotto:
The icons for custom programs should be installed in the appropriate subdirectories of the hicolor theme, not the root of the icons folder.

From the [specification](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html):

>In order to have a place for third party applications to install their icons there should always exist a theme called "hicolor" [1]. The data for the hicolor theme is available for download at: http://www.freedesktop.org/software/icon-theme/. Implementations are required to look in the "hicolor" theme if an icon was not found in the current theme.

I believe the current implementation only works due to fallback mechanisms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,5 +731,5 @@ include(CPack)
 
 if(UNIX AND NOT APPLE)
   install(FILES EmptyEpsilon.desktop DESTINATION "share/applications")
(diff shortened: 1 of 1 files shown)


daid:
Thanks, also for the effort of providing the documentation reference!

NicolaSmaniotto:
You are welcome!

PR #1639, opened 2021-12-21
update setting translation scenario_55_defenderHunter.lua

muerteFR:
@Xansta Do I also have to add the line :
difficulty = 1 --default (normal) like in Allies https://github.com/daid/EmptyEpsilon/pull/1624#issuecomment-989039970
although I have no errors on DefenderHunter?

PR #1659, opened 2022-01-19
Use getScenarioSettings in Defender Hunter

Xansta:
Per request in #1639 I have updated the scenario to use the new `getScenarioSettings()` function. I also separated difficulty into enemy ship strength and Murphy's law since `getScenarioSettings()` makes that possible and easy. I also added some additional enemy strength options: "Extreme" and "Quixotic" for the players for whom "Hard" is just not hard enough. I also added support for up to the maximum number of player ships playing simultaneously (currently 32)


muerteFR:
@Xansta , I have the impression that you made the changes without taking into account the update of the translation contexts that I did. If this PR is pushed, I would have done the work for nothing.

Xansta:
I tried

daid:
I also spotted a bug where you called _() with 3 strings instead of 2 in this change set.

muerteFR:
Where?

daid:
No, bug in this pullrequest

muerteFR:
Ok

PR #1661, opened 2022-01-25
Use getScenarioSettings in Defender Hunter

muerteFR:
I redid the PR proposed by @Xansta for the update of its scenario while waiting for it to resolve its problem of Fork not updated.

--> l7858


muerteFR:
Seen with @Xansta

PR #1674, opened 2022-02-02
Player Cruiser replacement

Xansta:
These suggested changes keep the 3D model in general use. The Player Cruiser can now only be deployed by the game master. This template, the Cruiser 2, is available for general use.

### model_data.lua
- Clone the existing battleship_destroyer_5_upgraded model and name it battleship_destroyer_5_upgraded_2
- Add two more visual beam source positions on the front of the ship where the guns stick out like the ones at the rear on the "wings"
- This model will be used in the Cruiser 2 player ship template
### corvette.lua
This template is similar to the original Player Cruiser template  
#### Differences:
- Original beams reduced from 10 damage per hit to 8 damage per hit
- Two additional beams added to correspond to the new visual sources on the updated model
  - arc: 60  
  - range: 800  
  - damage: 5  
  - Even though these beams are shorter range, since they're mounted further forward, they reach nearly as far as the other beams  
- Stronger shields (100,100 vs 80,80)
- Different speed (80, 12, 20 vs 90, 15, 20)
- One more repair crew (4 vs the default 3)
- Engineering ship layout more closely matches the overhead silhouette of the battleship_destroyer_5_upgraded model. The original rooms were identical to the Atlantis

### DIFF
--- a/scripts/model_data.lua
+++ b/scripts/model_data.lua
@@ -584,3 +584,25 @@ for type=1,5 do
         model:setIllumination("transport_space_ship_" .. type .. "/transport_space_ship_" .. type .. "_illumination.png")
     end
 end
+
+model = ModelData()
+model:setName("battleship_destroyer_5_upgraded_2")	--Cruiser PII
+model:setMesh("battleship_destroyer_5_upgraded/battleship_destroyer_5_upgraded.model")
+model:setTexture("battleship_destroyer_5_upgraded/battleship_destroyer_5_upgraded_color.jpg")
+model:setSpecular("battleship_destroyer_5_upgraded/battleship_destroyer_5_upgraded_specular.jpg")
+model:setIllumination("battleship_destroyer_5_upgraded/battleship_destroyer_5_upgraded_illumination.jpg")
+model:setScale(4)
+model:setRadius(200)
+
+-- Visual positions of the beams/missiletubes (blender: -X, Y, Z)
+model:addBeamPosition(-2, -31, -0.5)
+model:addBeamPosition(-2,  31, -0.5)
+model:addBeamPosition(35, -11, 0)
+model:addBeamPosition(35,  11, 0)
+model:addTubePosition(27, 0, -0.5)
+model:addTubePosition(27, 0, -0.5)
+model:addEngineEmitter(-33, 0, 0,  1.0, 0.2, 0.1, 16.0)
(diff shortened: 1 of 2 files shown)


daid:
The player cruiser served as a template for the Atlantis. But this might serve as a nice inbetween the Atlantis and the Phobos.

It does need a better name. Cruiser 2 is just meh.

daid:
Most of those sound good.

I'm also thinking that it's still quite close to the atlantis in "performance" maybe we need to adjust the missile storage a bit to make it more in-between the atlantis and the phobos?

Xansta:
It's got zero HVLIs (Atlantis and Phobos each have 20)

| Template | Homing | Nuke | Mine | EMP | HVLI |
| --- | --- | --- | --- | --- | --- |
| Phobos | 10 | 2 | 4 | 3 | 20 |
| Nacross (current) | 12 | 4 | 8 | 6 | 0 |
| Atlantis | 12 | 4 | 8 | 6 | 20 |
| Nacross (proposed) | 12 | 3 | 6 | 5 | 0 |

Nacross is an arbitrary name selection. Let me know if the name does not fall into the acceptable category

muerteFR:
I like Cruzeiro :-)

Xansta:
I plan to update the pull request to name the replacement Cruzeiro and change the missile load out as proposed (unless I hear otherwise)

aBlueShadow:
Reminds me of Draco inf the movie dragonheart "So instead of calling me 'Dragon' in your tongue, you'll call me 'Dragon' in some other tongue?" :-)

Xansta:
That's kind of how it came to me: what's another name for cruiser? I can't think of one, but Cruzeiro sounds like cruiser in Spanish or Portuguese. We said it in my last job working IT in the airline industry many times because it was one of those currencies we would use to test the boundaries of our numeric fields. Ticket prices denominated in Cruzeiros generated huge numbers. One of my colleagues was from São Paulo and she told me that the currency was so devalued that people would throw handfuls of 5 centavo coins into the street they were so worthless

Xansta:
I did not intend to close it, I just wanted to do another pull request.

Oh well. If anyone is _really_ interested, let me know and I'll resurrect it.

muerteFR:
You have to change branches. Create a branch for each PR. And then you can switch from one branch to another without losing your work.

PR #1683, opened 2022-02-09
add and modification of few translation tags about hotkey F1

muerteFR:


muerteFR:
...helpOverlay.cpp:22:42: error: ‘tr’ was not declared in this scope;

I don't know what I should do about this line 22. I don't see mistake.

muerteFR:
Ok. I thought about it but in the 2 other modified files crewStationScreen.cpp and MainScreen.cpp there are not either.

PR #1696, opened 2022-02-22
Satellites

aBlueShadow:
I made some assets that are all meant to be set in Earth (or other planet's) Orbit.

orbit

- Some standard satellite (yeah, we have the SensorBuoy MK2 models, but that's already used for probes)
- A debris version of that satellite (just distorted obj with same texture but no specular.)
- A military satellite
- Some unrecognizable debris. When two massive objects collide at orbital velocity, they will vaporize on impact and resolidify later. I did not find good footage of real space debris, so this is just my interpretation

satellites

The military "ANT 615" Satellite comes with a new ship template, the rest is intended to use as artifacts.


nallath (approved):
Changes look good. I haven't tested them yet.

PR #1698, opened 2022-02-22
Adding earth texture

aBlueShadow:
So far, Earth has been mentioned on very few occasions in the lore, now here it is!
earthimage
Based on Nasa imagery. I combined 
https://visibleearth.nasa.gov/images/57752/blue-marble-land-surface-shallow-water-and-shaded-topography and
https://visibleearth.nasa.gov/images/57730/the-blue-marble-land-surface-ocean-color-and-sea-ice
and adjusted the colors a bit.

P.S: Strictly speaking, I broke the naming scheme. So you might prefer if I rename it to planet-6 or planet-0.


daid:
The naming is good, as the rest are generated "random" textures, this is a very specific one. Anyone that sees it would say "yes, that's earth".

Not sure why you added so much saturation however, looks a bit overly saturated.

aBlueShadow:
Are you talking about the color of the water? as the rest has basically the same saturation as the original.

But you're right, we may just use the color as it is.

Originally I took an image from "blue marble next gen", where the water color is much darker. Later I find the older blue marble images more suitable, and just assumed they need a brighter water color as well. And as the water color of the iceless image was more uniform, I chose to combine them.

Here is how it looks without color modification:
earth

daid:
Looking good, think in your example the atmosphere color needs to be a bit darker, but the texture looks good.

PR #1700, opened 2022-02-23
Add 3 _() translation in scenario Beacon.lua

muerteFR:


Xansta:
While you're in there, can you please change 
J. J. Johnson on lifepod
to 
J. J. Johnson in lifepod

I fixed the first occurrence, but missed the second occurrence

muerteFR:
Doesn't matter,
Ok no pb

PR #1717, opened 2022-03-05
Script documentation improvements

edave64:
This PR adds the following changes:

- Show lua types, not C++ types
  - Take apart vector types into variadic or multiple paramaters
- Link to classes, enums and special types
- Sticky headers to make it always obvious where in the documentation you are (Removed the class prefix before methods and members)


edave64:
Odd, that exception doesn't occur on my system. I'll look into it

edave64:
Ok, forgot the SeriousProton updated comments. The build should work once that's merged. I also could make it so the python script survives situations like this, but also, it refusing to work if the lua functions aren't sufficiently documented can also be seen as a good thing.

daid:
Failing can be a good thing then yes, but the error is unclear in why it fails.

edave64:
Now it tells you exactly what the issue is, in what file the error is and what to search for to find the line and what to change :)

PR #1719, opened 2022-03-06
Birth of Atlantis Scenario - Handle nil Warp Jammers

Xansta:
Check for nil warp jammer before assessing distance

### DIFF
--- a/scripts/scenario_08_atlantis.lua
+++ b/scripts/scenario_08_atlantis.lua
@@ -977,15 +977,19 @@ function update(delta)
 
     -- If the player enters the Kraylor defense line, or engages a forward station, attack with all forces.
     for _, warp_jammer in ipairs(kraylor_defense_line) do
-        if distance(player, warp_jammer) < 6000 then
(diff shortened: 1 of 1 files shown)


scripts/scenario_08_atlantis.lua
```diff
@@ -977,15 +977,19 @@ function update(delta)
 
     -- If the player enters the Kraylor defense line, or engages a forward station, attack with all forces.
     for _, warp_jammer in ipairs(kraylor_defense_line) do
-        if distance(player, warp_jammer) < 6000 then
-            putKraylorDefenseLineOnFullOffense()
-        end
+    	if warp_jammer ~= nil and warp_jammer:isValid() then
+			if distance(player, warp_jammer) < 6000 then
```
daid, line 981: Tabs vs spaces

Xansta: Uh, I used tabs. Are you saying I should use spaces?

Xansta: I think I sorted it

PR #1729, opened 2022-03-15
better documentation for setShiedsMax

HagenRo:
I'm trying to write a scenario where one can upgrade the Playership. So this would be a nice feature.
did this in the browser but I think it should not break anything :)


daid:
But, you can already do this with `getShieldCount` + `getShieldsMax` and `setShieldsMax`

Untested, but something like:
```
function addShield(ship, charge)
  local shields = {}
  for n=1,ship:getShieldCount() do
    shields[n] = ship:getShieldsMax(n - 1)
  end
  ship:setShieldsMax(unpack(shields))
end
```

PR #1742, opened 2022-03-30
3 German translation files added

HagenRo:
for the new place_station_scenario_utility

for defenderHunter

for surrounded


daid:
@aBlueShadow can you check if this matches your style?

aBlueShadow:
I skimmed through it yesterday. I might have a more thorough later this day, however:
I created a shared document that combines a rough style guide and a translation glossary for frequent game terms. When HagenRo started his translation, I shared it with him with editing permission, so he can in turn add terms specific for some scenarios like the goods of xanstas trade subsystem.
So I am pretty confident it will fit my style overall :-)

PR #1743, opened 2022-03-31
better doc for onNewPlayerShip

HagenRo:
once in a while a wild doc addition appears

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -658,7 +658,8 @@ static int onNewPlayerShip(lua_State* L)
     return 0;
 }
 /// void onNewPlayerShip(ScriptSimpleCallback callback)
-/// Register a callback function that is called when a new ship is created on the ship selection screen.
+/// Register a callback function that is called when a new ship is created (on the ship selection screen or with the constructor in a lua script).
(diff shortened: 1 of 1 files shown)


HagenRo:
and as I'm confused by the callback interaction I'll ask it here.

why is this working:
`onNewPlayerShip(
    function(ship)
        initializePlayership(ship)
    end
)`

while this is not working:
`onNewPlayerShip(test)
test = function(ship)
    initializePlayership(ship)
end`

note there is no error occuring.

In my understanding there should be no difference between the anonymous function and the fuction reference.
What am I missing?

daid:
```
onNewPlayerShip(test)
test = function(ship)
    initializePlayership(ship)
end
```
Doesn't work because you use `test` before you assign it. So it's `nil` at that point, which is handled as "no callback"

```
function test(ship)
    initializePlayership(ship)
end
onNewPlayerShip(test)
```
Should work fine.

HagenRo:
I hate languages that don't load all functions befor execution :(

daid:
It's why the scenarios have an `init` function. You can define all your functions nicely in whatever order you want, and in the `init()` you do the `onNewPlayerShip(...)`

Issue #1737, opened 2022-03-22
Script reference ambiguity


Xansta:
There are several examples of this kind of ambiguity, but I'll just cite this one. For determining if shields are active, the current script reference looks like this:  
![Screen Shot 2022-03-22 at 9 27 19 AM](https://user-images.githubusercontent.com/35241832/159505140-3dd95da8-1a37-49ca-a9df-bd0c174ad6b9.png)  
It's a little ambiguous whether or not this is a stand alone function or one that is attached to something else. Admittedly, it kind of appears like it's attached to ShipTemplateBasedObject, but it's hard to tell when you look at the adjacent entries:  
![Screen Shot 2022-03-22 at 9 30 34 AM](https://user-images.githubusercontent.com/35241832/159505881-829bc5a6-b2d1-48ba-bee2-474960745957.png)
Also, it's hard to establish the relationship to ShipTemplateBasedObject just based on the formatting alone.

In the previous script reference, it was clear that the function should be part of a ship template based object:  
![Screen Shot 2022-03-22 at 9 32 31 AM](https://user-images.githubusercontent.com/35241832/159506304-4750fca0-23d4-42d5-ba8a-19145e00983d.png)

It's good to know that the function returns a boolean, but it would also be good to know that the function should be attached to a ship template based object. Could we have it look more like this:
`boolean ShipTemplateBasedObject getShieldsActive()`

aBlueShadow:
> Could we have it look more like this:
boolean ShipTemplateBasedObject getShieldsActive()

That wouldn't make much sense IMO, you would need the : as seperator, otherwise the relationships of the types _boolean_ and _ShipTemplateBasedObject_ won't be clear.

btw. It is still clear that getShieldsActive() is member of ShipTemplateBasedObject, as it's in the ShipTemplateBasedObject category, but I agree that it was more obvious in the older version.

aBlueShadow:
I am also not sure what you mean by "when you look at the adjacent entries". The parent is determined by the category (in this case ShipTemplateBasedObject), not by adjacent entries.

Are you referring to "ShipTemplateBasedObject setImpulseSoundFile"?
That has nothing to do with the category, ShipTemplateBasedObject is just the type of the return value (like boolean in the other example). As the description says "Returns the object it was called on.", so that's why the return value is incidentally the same data type as the class of the function

daid:
[...]

PR #1747, opened 2022-04-03
Print the class name with every member (EmptyEpsilon#1737)

edave64:
Addresses this: https://github.com/daid/EmptyEpsilon/issues/1737
Changes this:

ModelData setRadius
To this
ModelData ModelData:setRadius

If you really want that. I personally still don't like it.


daid:
It might look uglier, but I do think it is clearer this way

PR #1748, opened 2022-04-03
Documentation: Remove all non-font external references

edave64:
Even while being local, the scripting documentation is currently completely unreadable when offline, because it has multiple online dependencies. 

This removes all external references except for the fonts, including _all_ javascript (It was all unused), jquery-ui css and http://daid.github.io/EmptyEpsilon/main.css

There isn't actually a lot that these contributed, the new css is only 6 short lines longer. (Compared to 368,39kB in the previous dependencies)

The output is pixel perfectly identical with how it was before, except in offline mode where it now looks substancially better.

Originally I also wanted to turn the section \<div\>s into \<section\> elements, but turned out to actually have css implications, so now they are still divs.

### DIFF
--- a/compile_script_docs.py
+++ b/compile_script_docs.py
@@ -388,8 +388,6 @@ class DocumentationGenerator(object):
 
     def generateDocs(self, stream):
         stream.write('<!doctype html><html lang="us"><head><meta charset="utf-8"><title>EmptyEpsilon - Scripting documentation</title>')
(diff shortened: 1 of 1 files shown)


daid:
Ah, good, this script has a lot of history in it. I'm glad you take some effort to clean up some of the more annoying bits that are the result of that history :)

PR #1753, opened 2022-06-05
CinematicView: add button to cycle through all ships

Piglit:
When cycling is active, the cinematc view switches to the next ship every 30 seconds.
This is intended to give spectators of the game a more varying impression.


daid (commented):
Changes look good, except for tabs instead of spaces for indenting on some bits

Piglit:
Fixed the tabs instead of spaces.

PR #1759, opened 2022-08-18
Warp Jammer range getter, hull getter and setter.

hemmond:
Not sure if I should register `WarpJammer`'s `hull` attribute to member replication with `registerMemberReplication`. 

But since Hull is nowhere to be seen (yet, but there is getter so LUA script can get the value), I am not sure if it should be added there or not. 

Also when I was debugging this code (with demo LUA missions) I added some clarifications to LUA bindings elsewhere I had to test to know what it is really doing (sorry for mixing it into one commit).


daid:
You would only need a `registerMemberReplication` if clients need to know about it. Lua runs on the server, so there is no need right now. If hull was visible to science, there would be a need

hemmond:
Okay, then I have nothing more to add to it for now (and so it is ready for merge). I might someday in the future on add it to Science's screen by separate PR (or someone else could do it).

Issue #1763, opened 2022-09-06
Segfault when placing terrain


Piglit:
I get a segfault when running the following script:

```lua
function init()
    Planet()
    for cnt = 0, 15 do
        Mine()  -- also breaks when using Asteroid()
    end
end
```

The segfault occures after selecting game master screen or selecting any ship station.
It does not always occure; in about one out of three tries, the segfault does not happen.
When using a seperate client, the segfault happens on the client. The server keeps on running, as long no crew station is selected there.

I tested it on Debian 11 using your master branch (from today).
I think the same segfault happens when running the deliverAmbassador scenario, when createObjectsOnLine() is called.

Issue #1760, opened 2022-08-22
segfault on science screen


Chasethechicken:
EmptyEpsilon sometimes randomly crashes, when the science or operations screens are open.
I compiled from master (commit 61f8666) on Fedora 36.

Running using valgrind gives:
```
==37468== Invalid read of size 8
==37468==    at 0x5693DC: GuiRadarView::drawObjects(sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x56DABD: GuiRadarView::onDraw(sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x59747F: GuiContainer::drawElements(glm::vec<2, float, (glm::qualifier)0>, sp::Rect, sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x59749B: GuiContainer::drawElements(glm::vec<2, float, (glm::qualifier)0>, sp::Rect, sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x59749B: GuiContainer::drawElements(glm::vec<2, float, (glm::qualifier)0>, sp::Rect, sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x59749B: GuiContainer::drawElements(glm::vec<2, float, (glm::qualifier)0>, sp::Rect, sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x59749B: GuiContainer::drawElements(glm::vec<2, float, (glm::qualifier)0>, sp::Rect, sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x58C314: GuiCanvas::render(sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x66A38E: RenderLayer::render(sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x66A313: RenderLayer::render(sp::RenderTarget&) (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x64A1D5: Engine::runMainLoop() (in /usr/local/bin/EmptyEpsilon)
==37468==    by 0x461C47: main (in /usr/local/bin/EmptyEpsilon)
==37468==  Address 0xb2507d8 is 8 bytes before a block of size 4,800 alloc'd
[...]

PR #1766, opened 2022-09-15
fix segfault in radarView sort

Piglit:
issue: #1760
issue: #1763


daid:
Wait, it crashes because the results of the compare function are inconsistent sometimes? :|

PR #1772, opened 2022-11-02
Translation 95% scenario_74_omicron.fr.po

muerteFR:
Translated at 6%


muerteFR:
@daid can be validated as it is because I should update the fr.po before finalizing the translation.

PR #1773, opened 2022-11-02
clock displays formatted time

Piglit:
Relay/GM/ShipLogs show mission time in hh:mm:ss format, instead of seconds.

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -229,6 +229,16 @@ void GameGlobalInfo::destroy()
     MultiplayerObject::destroy();
 }
 
+string GameGlobalInfo::getMissionTime() {
+    unsigned int seconds = gameGlobalInfo->elapsed_time;
+    unsigned int minutes = (seconds / 60) % 60;
+    unsigned int hours = (seconds / 60 / 60) % 24;
+    seconds = seconds % 60;
+    char buf[9];
+    std::snprintf(buf, 9, "%02d:%02d:%02d", hours, minutes, seconds);
+    return string(buf);
+}
+
 string getSectorName(glm::vec2 position)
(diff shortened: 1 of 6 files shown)


daid:
@Xansta @StarryWisdom are you using the time in scenario communication anywhere? As this will impact that.

Xansta:
I use `getScenarioTime()` in several scenarios and the code assumes it's seconds elapsed since the start of the scenario. At a glance, it looks like this is a new function `getMissionTime()` which shows up formatted as hh:mm:ss. Am I reading this wrong?

Piglit:
This is correct.
I guess I should add a function for lua that gets the fromatted string like getMissionTime() does for the c++ code.
And/or rename it to getScenarioTimeFormated to avoid confusion.

daid:
> I use `getScenarioTime()` in several scenarios and the code assumes it's seconds elapsed since the start of the scenario. At a glance, it looks like this is a new function `getMissionTime()` which shows up formatted as hh:mm:ss. Am I reading this wrong?

My concern is that you are displaying this time somewhere to the user, and if we change the clock on relay to this new format, they can no longer match this time up.

Xansta:
I don't screen scrape the time from the relay screen. The `getScenarioTime()` is a somewhat recent addition or I only started using it somewhat recently. When I wish to put up a time measurement on one of the consoles via Lua using a custom widget, I usually want a countdown timer, so I've got to run some kind of calculation anyway. 

I'd suggest that the value returned by the Lua function `getScenarioTime()` remain as it is today (number of seconds since the start of the scenario). Having a formatted time option is useful. I suggest another Lua function provide that, like `getScenarioTimeFormatted()`, or that it be added as a second return parameter to `getScenarioTime()`

PR #1777, opened 2022-11-02
shipSelectionScreen: remember last chosen stations

Piglit:
in station selection screen commandSetCrewPosition() has to be called in addition to setting the buttons values. Otherwise the ready button is disabled.


daid:
This seems to be the wrong way to fix this, I'm not sure why the button is disabled and why it gets enabled by this. But it it's most likely some roundabout side effect.

PR #1778, opened 2022-11-03
improve ship password handling

Piglit:
* hide passwords in password entry fields with asterisks (affects server and ship passwords on clients)
* remember the last used ship password in running client session
* enter key in ship password entry dialog acts like pressing ok

### DIFF
--- a/src/playerInfo.h
+++ b/src/playerInfo.h
@@ -45,6 +45,7 @@ public:
     uint32_t main_screen_control = 0;
     int32_t ship_id;
     string name;
+    string last_ship_password;
 
     PlayerInfo();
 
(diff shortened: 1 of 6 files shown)


src/playerInfo.h
```diff
@@ -45,6 +45,7 @@ class PlayerInfo : public MultiplayerObject
     uint32_t main_screen_control = 0;
     int32_t ship_id;
     string name;
+    string last_password;
```
daid, line 48: there is a possible server password, gm password and ship control code, it's now not clear which this is used for.

Piglit: renamed to last_ship_password

PR #1787, opened 2022-11-09
fix scenario_74_omicron.lua

muerteFR:
Fix After destroying the final objective, the game does not end. Seen by Xansta.


daid:
Thats quite a few changes for a fix, this is aligned with Xansta?

muerteFR:
Yes, it's several small fixes seen with Xansta, it was he who gave me the bits of code to modify as I reported the problems to him.
As well as the French translation updated as and when.

PR #1791, opened 2022-11-26
Fixed URL for EE website in script_refernce.

hemmond:
It bugged be, that emptyepsilon.org is no longer redirecting in its default form, fixed URL in script_reference.html to where the EE will redirect anyways.  

Actually, the web itself can redirect if there is `www` in www.emptyepsilon.org but then SSL certificates bugs you. 

Also, Beacon of Light also lists `EmptyEpsilon.org` in description (and therefore in translation files), but I am not going to fix that: 
`The Beacon of Light scenario, built from the series at EmptyEpsilon.org.`

But the series that once was on the website is probably no longer there. Maybe time to update the mission description as well? 

Not setting it as draft. If this gets merged, I will create an issue with Beacon of Light series description since I am not sure how to approach that fix.


daid:
I never owned the domain and lost contact with who did. So all references should go to github.

PR #1794, opened 2022-12-03
Fix inability to repair systems

csibbitt:


csibbitt:
Thanks for this scenario, it was a fun one! Full disclosure, we didn't test this fix yet.

muerteFR:
Hello Thank you for your feedback. This is always very useful for authors.
Feel free to do so here : https://discord.gg/mXuJsC8sed
It's Xansta the author, I'm just testing a bit and translating into French. 
If we talk about the same thing, not being able to repair certain systems completely is wanted, some stations allow to repair systems that the crew cannot repair. @Xansta?

csibbitt:
I guess I should explain the bug we saw.

In our case, the sensor system and the self destruct system (lol!) were damaged, and when we visited a station it offered to repair them for us, but the price in the messages were Nil, like:

"Repair scanning system (Nil Rep)"

When we select that option, the comms reply was just '?' (which usually indicates a script exception) and terminated the comms, which is not what is supposed to happen. (Our sensors were also not repaired)

This line appears in my server log:

[CRITICAL]: Callback function error: [string "scenario_74_omicron.lua"]:3727: bad argument #1 to 'takeReputationPoints' (number expected, got nil)

You can see, this is because there was no price defined: https://github.com/daid/EmptyEpsilon/blob/EE-2022.10.28/scripts/scenario_74_omicron.lua#L3727

PR #1818, opened 2022-12-31
Translation 88% scenario_30_brokenglass.fr.po

muerteFR:


muerteFR:
@daid This translation can be validated as it is. I'll finish it later because the script needs a lot of changes regarding the translation. Changes referenced on USN in ee-scenario-feedback https://discord.com/channels/206597597455056898/945710809382793236

Issue #1817, opened 2022-12-31
[GameGlobalInfo] Scenario setting defaults not loaded on headless, server_scenario options


oznogon:
Launching a scenario without interacting with the scenario selection/options UIs, for instance by setting the `headless` or `server_scenario` options, doesn't appear to provide a way to set any `scenario_settings`, nor does it load any defined default values.

This causes scenarios that rely on scenario settings, like No. 00 _Basic_ and No. 74 _The Omicron Plague_, to fail trying to access them:

```
INFO: Load script: scenario_74_omicron.lua
     -----     Scenario: Omicron     -----     Version 0.0.1     -----
Lua 5.3
ERROR: LUA: init: [string "scenario_74_omicron.lua"]:84: attempt to index a nil value (field '?')
```

This may be related to `headless` and `server_scenario` invoking `gameGlobalInfo:startScenario` directly from `main.cpp`, while the Server Creation Screen loads and clears `scenario_settings` before invoking `startScenario`.

_Omicron_ L84:

```
enemy_power =   enemy_config[getScenarioSetting("Enemies")].number
```

In context, _Omicron_ tries to avoid this by checking `if getScenarioSetting == nil`, but it's unclear whether that works here. L84 is in that `if` statement's `else` outcome, which suggests it shouldn't be invoked if getScenarioSetting is actually `nil`:

```
function setVariations()
        if getScenarioSetting == nil then
                difficulty = 1
                adverseEffect = .995
                coolant_loss = .99995
                coolant_gain = .001
                starting_rep = 20
                enemy_power = 1
                rescue_time_limit = 20
                destroy_time_limit = 20
        else
                local enemy_config = {
                        ["Easy"] =              {number = .5},
                        ["Normal"] =    {number = 1},
                        ["Hard"] =              {number = 2},
                        ["Extreme"] =   {number = 3},
                        ["Quixotic"] =  {number = 5},
                }
L84 ->          enemy_power =   enemy_config[getScenarioSetting("Enemies")].number
```

oznogon:
FYI @Xansta

oznogon:
From the Discord: The other problem is that `if getScenarioSetting == nil` is true only if the EE version doesn't support the `getScenarioSetting()` function, even if no scenario settings are set.

So changing the condition in `setVariations()`:

```diff
function setVariations()
-         if getScenarioSetting == nil then
+         if (getScenarioSetting == nil or getScenarioSetting("Enemies") == "") then
```

[...]

PR #1829, opened 2023-01-02
[scenario_00] Initialize default scenario settings

oznogon:
When run with `headless` or `server_scenario` options, default scenario settings aren't set, breaking the scenario. Initialize using the defaults manually, then apply the settings if they exist.


oznogon:
Related to #1817

oznogon:
This seems to override scenario settings.

PR #1831, opened 2023-01-02
[scenario_00] Initialize default scenario settings

oznogon:
When run with `headless` or `server_scenario` options, default scenario settings aren't set, breaking the scenario. Initialize using the defaults manually, then apply the settings if they exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Build fixes
   - `update_locale` script target fixed
 - Scenario fixes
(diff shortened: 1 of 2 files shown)


oznogon:
If this seems like a solution, we should implement it in all official scenarios. The alternative would be to provide the defaults (maybe from ScenarioInfo?) to the scenario.

oznogon:
Closing in favor of solving by picking up defaults on scenario load.

PR #1837, opened 2023-01-06
[#1816] Add ShipWindow to ECrewPositions and autoconnect

### LINKED ISSUE #1816: Autoconnect for Main Screen and Ship Windows
We're currently running a completely overdone setup:
We have three clients set up as Ship Windows in offset angles, with three projectors creating a giant panoramic viewport.
I've read into the options.ini preferences file, but unfortunately there doesn't seem to be an option to autoconnect as Ship Window or dedicated Main Screen.
It's extremely tedious having to manually re-setup every single non-player station every time the ship blows up, especially for the Ship Windows, with their fiddly angle selection.
It would be great if the autoconnect function could be extended to those extra screens. If possible, with predefined viewport angle for the windows.

oznogon:
- Adds the Ship Window extra view as crew position 16/"shipwindow".
- Adds the `ship_window_angle` option, for now used only on autoconnect.

Example:

```
EmptyEpsilon autoconnect=16 ship_window_angle=45 ship_window_flags=1
```

Fixes #1816.


oznogon:
A few other options to do this, like setting the `shipWindow` crew position beyond `max_crew_positions`, or handling adding a new player option like `ship_window_autoconnect=45` to set autoconnect and the angle at the same time.

PR #1847, opened 2023-01-12
[SupplyDrop] Check and pass Player or SpaceShip onPickUp

### LINKED ISSUE #1802: SupplyDrop callback passes PlayerSpaceship even when collider is a CpuShip
If an `onPickUp` callback is set on a SupplyDrop object, the callback passes a PlayerSpaceship object to the callback function even if the SpaceShip picking it up is a CpuShip.

oznogon:
Pass the appropriate object to the `SupplyDrop:onPickUp()` callback if a non-player SpaceShip collides with the SupplyDrop.

Closes #1802.

### DIFF
--- a/src/spaceObjects/supplyDrop.cpp
+++ b/src/spaceObjects/supplyDrop.cpp
@@ -60,9 +60,18 @@ void SupplyDrop::collide(Collisionable* target, float force)
                 picked_up = true;
             }
         }
+
+        // If a callback is set, pick up the drop and pass the ship.
         if (on_pickup_callback.isSet())
         {
-            on_pickup_callback.call<void>(P<SupplyDrop>(this), player);
+            if (player)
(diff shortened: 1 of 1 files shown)


src/spaceObjects/supplyDrop.cpp
```diff
@@ -60,9 +60,18 @@ void SupplyDrop::collide(Collisionable* target, float force)
                 picked_up = true;
             }
         }
+
+        // If a callback is set, pick up the drop and pass the ship.
         if (on_pickup_callback.isSet())
         {
-            on_pickup_callback.call<void>(P<SupplyDrop>(this), player);
+            if (player)
```
daid, line 67: actually, you can just always pass ship, it doesn't really matter for the lua interfacing what kind of object it is.

PR #1848, opened 2023-01-12
[GameGlobalInfo] Initialize scenario settings on start

### LINKED ISSUE #1817: [GameGlobalInfo] Scenario setting defaults not loaded on headless, server_scenario options

oznogon:
Add `gameGlobalInfo->setScenarioSettings()` function, which takes
the scenario filename and an optional settings map. This always
initializes the setting defaults, then applies any passed settings
over them.

Rewrite `gameGlobalInfo->startScenario()` to take optional settings
and pass them to `setScenarioSettings()` prior to launching the
scenario.

Revise ServerScenarioOptionsScreen to collect settings and pass
them to `gameGlobalInfo->startScenario()` on scenario load, instead
of directly manipulating the `gameGlobalInfo->scenario_settings` map.

Fixes #1817.

### DIFF
--- a/src/gameGlobalInfo.cpp
+++ b/src/gameGlobalInfo.cpp
@@ -1,6 +1,7 @@
 #include <i18n.h>
 #include "gameGlobalInfo.h"
 #include "preferenceManager.h"
+#include "scenarioInfo.h"
 #include "scienceDatabase.h"
 #include "multiplayer_client.h"
 #include "soundManager.h"
@@ -186,6 +187,24 @@ void GameGlobalInfo::reset()
     {
         p->reset();
     }
+}
+
+void GameGlobalInfo::initializeScenarioSettings(string filename)
+{
+    // Parse the scenario metadata for default settings.
+    // Use this for non-interactive scenario loading, like headless and
+    // server_scenario game preferences.
+    ScenarioInfo info(filename);
+    LOG(INFO) << "Initializing settings for scenario " << info.name;
+
+    for(auto& setting : info.settings)
+    {
+        LOG(INFO) << setting.key << " scenario setting set to " << setting.default_option;
+        gameGlobalInfo->scenario_settings[setting.key] = setting.default_option;
+    }
+
+    // Set the scenario name.
+    gameGlobalInfo->scenario = info.name;
 }
 
 void GameGlobalInfo::startScenario(string filename)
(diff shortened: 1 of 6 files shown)


daid:
Looks good. One thing, the script function to load the same or new scenario, shouldn't that also run this function?

daid:
Thinking a bit more about this, couldn't just "startScenario" check if all settings have been set and else set a default for those that haven't been set?

oznogon:
Rewrote to initialize defaults always in startScenario(), then overwrite with passed settings.

PR #1861, opened 2023-01-17
[STBO] Move translatable term out of loop

### LINKED ISSUE #1695: [localisation issue] "Shield" at the top right cannot be translated in GM
The entry "Shield" in GM is one of the last words that is not translated. This corresponds to n7 of this old issue https://github.com/daid/EmptyEpsilon/issues/1256 and at this one https://github.com/daid/EmptyEpsilon/issues/1417#issuecomment-841918687

This has been partially resolved.
The other words could be translated except "shield"

shields

oznogon:
This appears to fix #1695, but I can't figure out why.


daid:
> I can't figure out why.
tr() vs trMark()

trMark is used to keep the sorting the same, but that doesn't work with formatting.

PR #1870, opened 2023-01-19
Translation 39% scenario_44_outpost.fr.po

muerteFR:


muerteFR:
It can be validated like that because there are still a lot of translations to do on it and it will take me some time.

PR #1872, opened 2023-01-21
[FactionInfo] Refactor

### LINKED ISSUE #530: REQUEST faction tools and relationships
- _table function **getAllFactionNames**()_ : This functions gets a table with all factions names.
- _table function **getAllFactionIds**()_ : This functions gets a table with all factions ids.
- _boolean function **isFriendly**(number factionId, number factionId)_ : This function returns true if the two factions are friendly (no need for spaceobject).
- _boolean function **isEnemy**(number factionId, number factionId)_ : This function returns true if the two factions are enemies (no need for spaceobject).
- _boolean function **setRelationship**(number factionId, number factionId, string relationship)_ : This function sets the relationship between two factions ("friendly","enemy","neutral") without using spaceobjects.
- _string function **getRelationship**(number factionId, number factionId)_ : This function gets the relationship between two factions ("friendly","enemy","neutral") without using spaceobjects.

oznogon:
Refactor FactionInfo to close #530.

- Add new global scripting function `getFactionInfos()` to return all FactionInfo objects.
- Convert `EFriendVsFoeState` enum to "friendly", "enemy", "neutral" in scripting and add to scripting docs
- Set factionInfo limit in a constant and use it instead of `factionInfo.size()`
- Add `getFactionId()` and expose to scripting
- Expose `getName()`, `getLocaleName()`, `getDescription()` to scripting
- Add `getRelationshipWith()` and `setRelationshipWith()` and expose to scripting.
  - `getRelationshipWith()` replaces `getState()` in internal usage.
- Add `setNeutral()` to reset a faction's friend/enemy mask toward another faction, and `resetAllRelationships()` to set all relationships to neutral.

This adds potential breaking changes:

- Renames `getFactionInfo()` to `getFactionInfoByName()`. This function isn't used in scenarios that ship with EE, but might be used by others.
- `findFactionId()` (not exposed to scripting) now returns 255 instead of 0 on an failed response, to avoid returning a valid faction ID unexpectedly.

### DIFF
--- a/compile_script_docs.py
+++ b/compile_script_docs.py
@@ -189,7 +189,7 @@ def translate_type(c_type, name):
     if res is not None:
         return VariadicType(translate_type(res.group(1).strip(), name))
 
-    if c_type in ('EAlertLevel', 'ECrewPosition', 'EMissileSizes', 'EMissileWeapons', 'EScannedState', 'ESystem', 'EMainScreenSetting', 'EMainScreenOverlay', 'EDockingState', 'ScriptSimpleCallback'):
+    if c_type in ('EAlertLevel', 'ECrewPosition', 'EFriendVsFoeState', 'EMissileSizes', 'EMissileWeapons', 'EScannedState', 'ESystem', 'EMainScreenSetting', 'EMainScreenOverlay', 'EDockingState', 'ScriptSimpleCallback'):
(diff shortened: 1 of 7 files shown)


oznogon:
Broken up into #1952, #1953, #1955, #1956.

PR #1879, opened 2023-01-23
[GameStateLogger] Replace dtoa_milo with std::to_chars

oznogon:
Proof-of-concept to replace dtoa_milo with C++17's std::to_chars. Probably a better way to do this, but it logs Battlefield on Huge without a measured drop in performance without causing daid/SeriousProton#54.

This could remove a dependency, but floating-point `to_chars` might not be fully implemented yet in all compilers' libc.

### DIFF
--- a/src/gameStateLogger.cpp
+++ b/src/gameStateLogger.cpp
@@ -1,8 +1,7 @@
 #include <memory>
 #include <time.h>
-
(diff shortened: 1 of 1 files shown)


oznogon:
As expected, macOS fails without a float implementation of std::to_chars. :-1:  :apple: :gun:

PR #1880, opened 2023-01-24
Translation 65% scenario_44_outpost.fr.po (420 entries remaining)

muerteFR:
And changes to similar entries in Omicron.fr.po


muerteFR:
I apologize if I do too much PR for the translation of a single scenario.
As there are many entries and it's not easy to translate (meaning, size...) and I modify similar entries on the other scenarios of Xansta (size too big) I prefer to push several PR rather than making multiple commits. If it's annoying let me know @daid  Thanks

daid:
Not a problem at all.

PR #1885, opened 2023-01-31
[GameStateLogger] Replace int-, float-to-string with to_chars

oznogon:
Use std::to_chars for int-to-string conversions, and nlohmann::detail::to_chars for float-to-string conversions, removing the less well-maintained dtoa_milo implementation.

nlohmann::detail::to_chars has a similar Grisu2 algorithm implementation as dtoa_milo but is actively maintained, and we already include it in SeriousProton for JSON keybind output.

std::to_chars float-to-string conversions perform better than either milo or nlohmann, but they don't have broad compiler support yet. As of now, MinGW and macOS Clang don't implement it. When std::to_chars support for floats is better, we can replace nlohmann::detail::to_chars with std::to_chars with few required changes (remove the need to manually add the /0 terminator).

### DIFF
--- a/src/gameStateLogger.cpp
+++ b/src/gameStateLogger.cpp
@@ -1,8 +1,14 @@
 #include <memory>
 #include <time.h>
 
-//We need a really fast float to string conversion. dtoa from milo does this very well.
-#include "dtoa/dtoa_milo.h"
+// std::to_chars for integers, and someday floats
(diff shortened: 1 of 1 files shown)


oznogon:
I know we're probably removing this feature, but I still wanted to see how CI deals with this.

PR #1886, opened 2023-02-02
lua minigame for micrometeorites point defense

aBlueShadow:
This is some plugin style lua script that provides an extremly simple "minigame" for the weapons console.
micrometeorite

It is basically an attempt to keep weapons on their toes outside of combat or scripted events. 
Every now and then, a micrometeorite is detected, and weapons has a short time windows to shoot it down by pressing the right button.
In can be en- and disabled per player ship. With shields on, it is always inactive, as this is a rough indicator of being in combat. In game, this could be explained by stating that micrometeorites are just absorbed by the shield.
It is also inactive when no weapons officer is present, as tactical and single pilot are busy enough.

I am not quite sure about the actual utility of this, but I figured publishing I might be useful anyways, at the very least as an example of modularization for scenario writers, so one might be encouraged to outsource systems so that they can be reused in other scenarios.
Shoutout to @hemmond, looking into the custom elements wrapper helped me a lot with OOP-izing the minigame.

I also added translation tags, although I am not sure if utility scripts are translatable at all. But as those tags did no harm either, I added them anyway

### DIFF
--- a/scripts/locale/utils_micrometeorites.de.po
+++ b/scripts/locale/utils_micrometeorites.de.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: \n"
+"PO-Revision-Date: 2023-11-20 11:57+0100\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: de_DE\n"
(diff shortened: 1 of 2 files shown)


aBlueShadow:
As I succesfully tested that the language tags do work indeed, I added a german .po file as well

aBlueShadow:
Some changes:

- Small fixes like the removal of debug messages.
- Lowered the volume so the beam sound is more subtle.
- Unicode characters are outside of the translation strings now, as those seemed to cause problems.
- Convenient updateAll() function, so you won't need to manually update for any ship with point defense.

DansDesigns:
This is really cool, is it possible to have this as an add-on? How would one add this to EE?

aBlueShadow:
@DansDesigns given that this PR did not get merged, the first step would be to download the utils_micrometeorites.lua file and save it into your scripts folder. That file is only needed on your server, not on the clients.
There is no way to have global addons, I think even as a mod it won't be possible to get it started automatically. Which means you will have to include it into the scenario where you want that feature, the description is in the "Usage" section in the file. If you want to have a convenient way to add/clear that minigame from selected ships, this howto would be slightly different:

1) At the top of the scenario script (after the description section), insert
`require('utils_micrometeorites.lua')`

2) Add those lines somewhere within the init() function:
```
    addGMFunction(_("Remove MMPD"), function() for _, obj in ipairs(getGMSelection()) do MicroMeteorites:remove(obj) end end)
    addGMFunction(_("Add MMPD"), function() for _, obj in ipairs(getGMSelection()) do MicroMeteorites:init(obj) end end)

```
3) Add this line somewhere within the update() function:
`MicroMeteorites:updateAll(delta)`


I also thought about preparing this library for usage with the Script() object, so you would only need a single line to include and activate it, but as daid does not like Script(), I hesitated to include something like that in this PR. But as this PR has been pending for more than 1 1/2 years now, it probably won't matter much anyways :-)

DansDesigns:
@aBlueShadow aha that's exactly what i'm looking for! thank you so much!

DansDesigns:
This is really great!
Only issue im having is that there is no visual indication in Weapons radar that there is an incoming meteorite..
I saw the code for the radar Artefact in utils_micrometeorites.lua, do you know why there is no visual marker?

aBlueShadow:
Well, there is no way to show objects exclusively on the weapons radar. If I would spawn the meteorites as a physical object, they would also appear on helms and science, potentially distracting them, and they have probably other stuff to do. 
The Artifact only serves as a temporary guide for the beam effect, as it needs a target to work. The reason why you don't see it is that it gets created and destroyed at the same frame.
A physical object would also complicate stuff, and it could be dodged, which kinda defeats the purpose of giving weapons more to do.
You could also explain it away by saying, as that the micrometeorites are so small, they don't show up on the radar(and on the main screen, where you only see the beam and the explosion) - only the "proximity sensors" on the point defense can detect them.

aBlueShadow:
Destroying the artifact in the same tick won't let the beam effect happen in ECS. So I gave it a very low lifetime instead.

PR #1902, opened 2023-02-20
[Asteroid] Refactor asteroid rendering on radar

### LINKED ISSUE #1893: missmatch between asteroid visual size and hitbox
Since the SDL change there seem to be a lot more occasions where the hitbox of asteroids seem wrong.
Its sadly somewhat hard to confirm what people really see when the collisions happen (as after the asteroid no longer exists)

This is also harder to identify than one would hope due to not being able to see easily the hitbox of ships

With some testing it looks like it is worse with larger sized asteroids, and much worse with asteroids that are larger than the gm screen can create.

Looking at the code - the 32 and 64 in this function look suspicious (though they may just be correcting for radius/diameter)
https://github.com/daid/EmptyEpsilon/blob/3489f30940ce2c5c1c3b3d1653d4805e94c99123/src/spaceObjects/asteroid.cpp#L83

I have not tested if this resolves the miss match observed.

oznogon:
Fixes #1893.

Replaces the sprite sizing logic for asteroids to use `getRadius() * 2` for the sprite width.

In master branch behavior, large asteroids (>150 radius) are drawn much smaller on radar than their radius. 10, 250, 500 radius from left to right:

image

image

With this PR, radius 10, 250, 500 from left to right:

image

image

With this PR in a debug build, which draws an extra circle outline using `getRadius()`:

image

image

### DIFF
--- a/src/spaceObjects/asteroid.cpp
+++ b/src/spaceObjects/asteroid.cpp
@@ -81,9 +81,6 @@ void Asteroid::drawOnRadar(sp::RenderTarget& renderer, glm::vec2 position, float
         setRadius(size);
 
     renderer.drawSprite("radar/blip.png", position, std::max(6.0f, (getRadius() * 2.0f) * scale), glm::u8vec4(255, 200, 100, 255));
-#ifdef DEBUG
-    renderer.drawCircleOutline(position, std::max(3.0f, getRadius() * scale), 2.0f, glm::u8vec4(255, 200, 100, 255));
-#endif // DEBUG
 }
 
 void Asteroid::collide(Collisionable* target, float force)
--- a/src/spaceObjects/asteroid.cpp
+++ b/src/spaceObjects/asteroid.cpp
@@ -80,10 +80,10 @@ void Asteroid::drawOnRadar(sp::RenderTarget& renderer, glm::vec2 position, float
     if (size != getRadius())
         setRadius(size);
 
-    float size = getRadius() * scale / 64.0f;
-    if (size < 0.2f)
-        size = 0.2f;
-    renderer.drawSprite("radar/blip.png", position, size * 32.0f, glm::u8vec4(255, 200, 100, 255));
+    renderer.drawSprite("radar/blip.png", position, std::max(6.0f, (getRadius() * 2.0f) * scale), glm::u8vec4(255, 200, 100, 255));
+#ifdef DEBUG
+    renderer.drawCircleOutline(position, std::max(3.0f, getRadius() * scale), 2.0f, glm::u8vec4(255, 200, 100, 255));
+#endif // DEBUG
 }
 
 void Asteroid::collide(Collisionable* target, float force)


src/spaceObjects/asteroid.cpp
```diff
@@ -80,10 +80,10 @@ void Asteroid::drawOnRadar(sp::RenderTarget& renderer, glm::vec2 position, float
     if (size != getRadius())
         setRadius(size);
 
-    float size = getRadius() * scale / 64.0f;
-    if (size < 0.2f)
-        size = 0.2f;
-    renderer.drawSprite("radar/blip.png", position, size * 32.0f, glm::u8vec4(255, 200, 100, 255));
+    renderer.drawSprite("radar/blip.png", position, std::max(6.0f, (getRadius() * 2.0f) * scale), glm::u8vec4(255, 200, 100, 255));
```
daid, line 83: I would expect a min as well, or a std::clamp in that case. As it's nice to always be able to see the asteroids, even when zoomed out.

The ECS branch "generalized" this with a RadarTrace component that has a min/max/scaled size properties, and there I already applied this for asteroids.

src/spaceObjects/asteroid.cpp
```diff
@@ -80,10 +80,10 @@ void Asteroid::drawOnRadar(sp::RenderTarget& renderer, glm::vec2 position, float
...
+#ifdef DEBUG
+    renderer.drawCircleOutline(position, std::max(3.0f, getRadius() * scale), 2.0f, glm::u8vec4(255, 200, 100, 255));
```
daid, line 85: I see how this was useful during development, but I rather not have it there by default.

src/spaceObjects/asteroid.cpp
oznogon: > I would expect a min as well, or a std::clamp in that case. As it's nice to always be able to see the asteroids, even when zoomed out.

`std::max` returns the larger of the two given values, so when the scale is low (zoomed out) or the asteroid is small, the sprite is always drawn at least 6 pixels wide.

This is visible in the screenshots; the 10-radius asteroid at 0.09 scale (max GM screen zoom) would be ~2 pixels, but is rendered 6 pixels wide, or at about the same size as the master branch.

I use 6 pixels because this is close to the result of the existing `if size < 0.2 then size = 0.2` gate; 0.2 * 32 = 6.4.

Science "Zoom 1.0x", labeled by asteroid radius:

image

"Zoom 6.0x": (scale ~0.09)

image

Relay "Zoom 1.0x": (scale ~0.01)

image

```
[INFO    ]: std::max(6.0f, (getRadius() * 2.0f) * scale): 6.00
getRadius(): 10.00
scale: 0.01
[INFO    ]: std::max(6.0f, (getRadius() * 2.0f) * scale): 6.00
getRadius(): 121.19
scale: 0.01
[INFO    ]: std::max(6.0f, (getRadius() * 2.0f) * scale): 6.00
getRadius(): 253.12
scale: 0.01
[INFO    ]: std::max(6.0f, (getRadius() * 2.0f) * scale): 9.00
getRadius(): 500.00
scale: 0.01
```

Adding `min` or using `clamp` would allow setting a limit on the maximum size of the asteroid sprite, but the radar sprite not accurately reflecting large asteroids' collidable size. I'm not sure what the purpose of an upper limit would be.

daid: ah, my mistake, I was thinking about it wrong :)

Issue #1903, opened 2023-02-21
player_ship:commandDock(station) triggers docking procedure even when out of range


hemmond:
When you trigger `player_ship:commandDock(station)` when the station is far away from you, it creates interesting side effect of automatically turning on impulse to 100% reverse no matter where behind you the station was. In my example it was around player ship's 145 degrees, where zero degrees is player ship's current heading.

I think the distance check should be not only in GUI but also in the EE logic itself. Either don't execute it at all, or when you do, then at least start the maneuver completely (full reverse and turn ship's aft section directly at target station). But this behaviour is confusing and weird. From my point of view, part of the code worked (impulse part) and part did not (steering part).

Oznogon's response from Discord:
> I think the distance check should be not only in GUI but also in the EE logic itself.

The `command` function just sends a packet that results in a call to SpaceShip:requestDock(), which should have a range limit:

```
    if (glm::length(getPosition() - target->getPosition()) > 1000 + target->getRadius())
        return;
```

So that might be broken.

oznogon:
Tried to reproduce this behavior on a Windows build off master and couldn't.

I added a log line to the above range check:

```diff
    if (glm::length(getPosition() - target->getPosition()) > 1000 + target->getRadius())
+   {
+       LOG(INFO) << "If I'm more than 1U from the docking target I shouldn't be here.\nglm::length(getPosition() - target->getPosition()): " << string(glm::length(getPosition() - target->getPosition()),0);
        return;
    }
    if (!canStartDocking())
        return;
+   LOG(INFO) << "Now I'm docking";
```

Launched the master-branch build into scenario_99_battlefield.lua with httpserver enabled and spawned a player ship at 0,0. I set the player ship's throttle to 100% forward and pointed it at a heading that would have to change to align for docking with the station. From the GM screen, I targeted the neutral station about 15U north.

Then I ran via exec.lua API:

```
player = getPlayerShip(-1)
station = getGMSelection()[1];
player:commandDock(station)
```

I confirmed in the log output that the `requestDock()` call occurred, and that it hit the range check and failed the condition, and that it returned early without initiating docking, all as expected:

```
DEBUG: Object count: 691 79
INFO: If I'm more than 1U from the docking target I shouldn't be here.
glm::length(getPosition() - target->getPosition()): 18272
DEBUG: Object count: 684 75
```

oznogon:
[...]

Issue #1904, opened 2023-02-21
Internally docked ship's label is still visible on carrier ship's helms radar


hemmond:
I've noticed, that you still see callsigns of all internally docked ships on Helms radar (positioned where your callsign would be). Not sure if this is desired behaviour. Big carriers might have problem with this.

Oznogon's Discord response: 
Looks like internal docking is kind of a hack; the ships still exist in-world at the carrier's coordinates, they're just not rendered on radars or 3D views. So it's probably an oversight.

oznogon:
`GuiRadarView::drawObjects` draws callsigns. Objects are determined to be visible if they're not obscured by fog-of-war (nebula obstruction on short-/long-range radar, outside of friendly/probe short-range radar on Relay). Then if callsigns are being shown on this radar type, and the object has a callsign, it's rendered:

```cpp
            if (show_callsigns && obj->getCallSign() != "")
                renderer.drawText(sp::Rect(object_position_on_screen.x, object_position_on_screen.y - 15, 0, 0), obj->getCallSign(), sp::Alignment::Center, 15, bold_font);
```

Could extend that check to confirm the object's docking status. `obj` is a SpaceObject, and while the `DockStyle` enum is in SpaceObject, there's no check for an object's current DockStyle except in the `docked_style` param on SpaceShips. So we could move `docked_style` up into SpaceObject or have a separate condition if obj is a SpaceShip.

A more complete solution might be for carriers to maintain a list of internally docked objects. RadarView could then exclude those from `visible_objects` instead of having `if (docked_style == DockStyle::Internal) return;` checks in SpaceShip `drawOnRadar` (and other similar checks for things like `SpaceShip:drawOnGMRadar`, `SpaceShip:draw3D`, and collision data). That'd also solve the problem of getting a list of all ships docked with a carrier.

PR #1909, opened 2023-02-22
[#1903] Don't try to dock with ships already docked with yourself

oznogon:
Building on #1906, also don't try to dock with a ship that's already docked with our ship.

For example, given Benedict callsign "Spock" initiated and successfully externally docked to another Benedict callsign "Kirk", "Spock" still appears to "Kirk" as a valid docking target.  If "Kirk" attempts to dock with "Spock", "Kirk" enters an endless loop of trying to rotate to dock with "Spock", which moves as "Kirk" rotates.

This could benefit from a potentially recursive check of dockable ships that are docked to ships that are docked to the player; it's still possible to enter an endless docking cycle in those more edge cases of 3+ ships chained together.

This PR also cancels docking if the docking target is moved >1U away from the docking ship.

### DIFF
--- a/src/screenComponents/dockingButton.cpp
+++ b/src/screenComponents/dockingButton.cpp
@@ -89,12 +89,22 @@ P<SpaceObject> GuiDockingButton::findDockingTarget()
 {
     PVector<Collisionable> obj_list = CollisionManager::queryArea(my_spaceship->getPosition() - glm::vec2(1000, 1000), my_spaceship->getPosition() + glm::vec2(1000, 1000));
     P<SpaceObject> dock_object;
-    foreach(Collisionable, obj, obj_list)
+
+    for (P<Collisionable> obj : obj_list)
     {
         dock_object = obj;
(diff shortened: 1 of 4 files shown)


daid:
I wonder if we should prevent docking if there are any ships docked externally on yourself. I think that should solve most issues, as it will prevent creating odd chains of ships docked to each other.

oznogon:
> I wonder if we should prevent docking if there are any ships docked externally on yourself. I think that should solve most issues, as it will prevent creating odd chains of ships docked to each other.

IIRC there are checks if a ship is docked with another ship, but not if a ship has other ships docked to it.

So we could prevent other ships from docking to a ship already docked with another ship, but right now I don't think we can prevent a ship that's already carrying another externally docked ship from docking with another ship. So chains would still be possible.

ie. Ship C docks with ship B -> ship B docks with ship A -> ship A doesn't know ship C is indirectly docked with it and could still try to dock with ship C

Similar to #1904/#1905, if ships stored data on which other ships are docked to them, we could control that behavior both ways.

PR #1911, opened 2023-02-22
[GameStateLogger] Replace float-to-string with to_chars

oznogon:
This removes the dependency on dtoa_milo, which has been problematic (daid/SeriousProton#54, daid/SeriousProton#48, daid/SeriousProton#224).

nlohmann::detail::to_chars uses a similar Grisu2 algorithm implementation as dtoa_milo and is actively maintained. Also, we already include nlohmann::json in SeriousProton for keybind output.

std::to_chars float-to-string conversions in charconv perform better than either milo or nlohmann, but they don't have broad compiler support yet. As of now, MinGW and macOS Clang don't implement it.

When std::to_chars support for floats is better, we can replace nlohmann::detail::to_chars with std::to_chars, with few required changes (remove the manually added /0 terminator).

If we use std:: or nlohmann::to_chars for floats, we can eventually also use it for integers to improve performance over sprintf.

### DIFF
--- a/src/gameStateLogger.cpp
+++ b/src/gameStateLogger.cpp
@@ -1,8 +1,11 @@
 #include <memory>
 #include <time.h>
 
-//We need a really fast float to string conversion. dtoa from milo does this very well.
-#include "dtoa/dtoa_milo.h"
+// SeriousProton provides nlohmann/json.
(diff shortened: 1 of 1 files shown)


daid:
Removing a dependency, I love it :D

PR #1920, opened 2023-02-27
Reverse proxy status in server creation page

Tsht:
If server is launched in reverse proxy mode (serverproxy=ips), show this as info instead of regular server screen.

### DIFF
--- a/src/menus/serverCreationScreen.cpp
+++ b/src/menus/serverCreationScreen.cpp
@@ -43,7 +43,6 @@ ServerSetupScreen::ServerSetupScreen()
     }
     else
     {
-    
         // Left column contents.
         // General section.
         (new GuiLabel(main_panel, "GENERAL_LABEL", tr("Server configuration"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
--- a/src/menus/serverCreationScreen.cpp
+++ b/src/menus/serverCreationScreen.cpp
@@ -69,7 +69,7 @@ ServerSetupScreen::ServerSetupScreen()
     string reverse_proxy_value;
     if((reverse_proxy_value = PreferencesManager::get("serverproxy")) != "")
     {   
-        GuiPanel* panel = new GuiPanel(main_panel, "TUTORIAL_DESCRIPTION_BOX");
+        GuiPanel* panel = new GuiPanel(main_panel, "SERVERPROXY_MESSAGE_BOX");
         panel->setSize(GuiElement::GuiSizeMax, 80);
         //Serverproxy (reverse proxy) is directly configured in options or command line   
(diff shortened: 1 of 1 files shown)


src/menus/serverCreationScreen.cpp
```diff
@@ -26,58 +26,75 @@ ServerSetupScreen::ServerSetupScreen()
...
+        (new GuiLabel(row, "NAME_LABEL", tr("Server name: "), 30))->setAlignment(sp::Alignment::CenterRight)->setSize(250, GuiElement::GuiSizeMax);
+        server_name = new GuiTextEntry(row, "SERVER_NAME", "server");
+        server_name->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
+
+        // Server password row.
+        row = new GuiElement(main_panel, "");
+        row->setSize(GuiElement::GuiSizeMax, 50)->setAttribute("layout", "horizontal");
+        (new GuiLabel(row, "PASSWORD_LABEL", tr("Server password: "), 30))->setAlignment(sp::Alignment::CenterRight)->setSize(250, GuiElement::GuiSizeMax);
+        server_password = new GuiTextEntry(row, "SERVER_PASSWORD", "");
+        server_password->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
```
oznogon, line 62: If the server password isn't shown or set here, where is it shown or set? `headless_password` appears to be set only if `headless` is in use, and I don't see another way to set a server password: https://github.com/daid/EmptyEpsilon/blob/bdaa9474bffefeb82ac4c40523edc51d37f865f0/src/main.cpp#L448

Tsht: never set a server password for reverse proxy, didn't even know it could be done. It must be the same for each server ? I will check this.

Tsht: ok I guess it's using a headless server combined with a reverse proxy ? I didn't know about this headless mode but indeed some options should be usable here in reverse proxy mode

src/menus/serverCreationScreen.cpp
```diff
@@ -65,6 +65,24 @@ ServerSetupScreen::ServerSetupScreen()
     server_port->setSize(GuiElement::GuiSizeMax, GuiElement::GuiSizeMax);
 
     (new GuiLabel(main_panel, "GENERAL_LABEL", tr("Server info"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50);
+
+    string reverse_proxy_value;
+    if((reverse_proxy_value = PreferencesManager::get("serverproxy")) != "")
+    {   
+        GuiPanel* panel = new GuiPanel(main_panel, "TUTORIAL_DESCRIPTION_BOX");
```
daid, line 72: TUTORIAL_DESCRIPTION_BOX?

PR #1884, opened 2023-01-30
Added translation tags to translate credits mainMenus.cpp

muerteFR:

PR #1927, opened 2023-03-02
[translation] Apply msgctxts from #1884 to en.po

### LINKED ISSUE #1922: [translation, regression] Main menu screen no longer shows translated texts.
When starting EE, main menu screen is in english no matter what language was selected. Other screens are in specified language, but main screen stays in english. Doesn't matter if EE was launched with different language or set in options, EE main menu screen is still in english. 

Observed on current GIT Master (commit: eb5135b83b41041bbfed5e63483c82f4a6e75bb8 ), issue is NOT present on latest release (2022.10.28).

oznogon:
#1884 added `mainMenu` translation contexts to main menu buttons, then applied those contexts to the French main.fr.po only, resulting in #1922. Running update_locale.py then removes the French contexts.

Merging this change, which copies the main.fr.po context changes in #1884 to main.en.po, then running update_locale.py to update contexts in other language .po files, should fix #1922.

### DIFF
--- a/resources/locale/main.en.po
+++ b/resources/locale/main.en.po
@@ -350,31 +350,78 @@ msgid "Last Session ({last})"
 msgstr ""
 
 #: src/menus/mainMenus.cpp:72
+msgctxt "Credits"
 msgid "Version: {version}"
 msgstr ""
 
 #: src/menus/mainMenus.cpp:74
+msgctxt "mainMenu"
 msgid "Your name:"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


oznogon:
Closing this PR because this should be done by `ninja update_locale` or an equivalent per the [wiki's translation guide](https://github.com/daid/EmptyEpsilon/wiki/Translation-and-Localization#adding-translations-of-game-text).

PR #1930, opened 2023-03-04
[GuiListbox] Add support for icons on GuiListbox entries

oznogon:
Similar to GuiButton, allow GuiListbox entries to have icons.

Depends on daid/SeriousProton#232.


daid:
How does this look? I get the feeling when I look at the code that the icons are right on the "buttons". But that wouldn't match the buttons with icons, which have the icons on the left.

oznogon:
image

PR #1934, opened 2023-03-04
[#1933] Validate impulse_request values

### LINKED ISSUE #1933: [AI] CpuShips with 0 max_impulse_speed can set infinite impulse_request values
Steps to reproduce

1. Enable game state logging (create a `logs` directory in the game executable directory)
2. Launch a server, start the Battlefield scenario, and open the GM screen.
3. Unpause the game.
4. Click "Create..."
5. From the cpuship list, select Defense Platform or Odin.
6. Place the ship in the middle of the battle.

Expected behavior:

The ships spawn and are logged by the game state logger. The game doesn't crash.

Observed behavior:

The ship spawns, and the game crashes upon the next tick of the game state logger.

```
Assertion failed: std::isfinite(value), file git/daid/SeriousProton/libs/Box2D/../nlohmann/json.hpp, line 17889
```

json.hpp L17889:

```cpp
    JSON_ASSERT(std::isfinite(value));
```

In context:

```cpp
@note The buffer must be large enough.
@note The result is NOT null-terminated.
*/
template<typename FloatType>
JSON_HEDLEY_NON_NULL(1, 2)
JSON_HEDLEY_RETURNS_NON_NULL
char* to_chars(char* first, const char* last, FloatType value)
{
    static_cast<void>(last); // maybe unused - fix warning
    JSON_ASSERT(std::isfinite(value)); // <-- HERE

    // Use signbit(value) instead of (value < 0) since signbit works for -0.
    if (std::signbit(value))
    {
        value = -value;
        *first++ = '-';
    }

#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfloat-equal"
```

Adding logging to GameStateLogger prints the value in question as `inf` or `-inf`, which trips nlohmann::json's `isfinite` check.

Defense platform:

```
WARN: GameStateLogger::write, key: impulse
value: inf
WARN: _f: inf
WARN: std::isfinite(_f): 0
```

Odin:

```
WARN: GameStateLogger::write, key: impulse
value: -inf
WARN: _f: -inf
WARN: std::isfinite(_f): 0
```

Valid impulse output for comparison:

```
WARN: GameStateLogger::write, key: impulse
value: 0.34
WARN: _f: 0.34
WARN: std::isfinite(_f): 1
```

This notably doesn't appear to affect these ships when they're created in situations where they aren't expected or want to move, such as an empty scenario script or via GM object creation in an empty scenario. In those cases, the impulse value is just 0:

```
WARN: GameStateLogger::write, key: impulse
value: 0.00
WARN: _f: 0.00
WARN: std::isfinite(_f): 1
```

Regardless of game state logging, ships likely shouldn't ever have infinite/-infinite impulse values.

oznogon:
#1933 shows that directly setting `impulse_request` on a `SpaceShip` can result in unusable or poorly defined values, such as infinity/-infinity from dividing by 0 impulse speed on Odins and Defense Platforms.

Add a `setImpulseRequest()` function that clamps the passed value and replace direct modifications of `impulse_request` with it.

Fixes #1933.

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -598,14 +598,21 @@ void ShipAI::flyTowards(glm::vec2 target, float keep_distance)
         if (pathPlanner.route.size() > 1)
             keep_distance = 0.0;
 
-        if (distance > keep_distance + owner->impulse_max_speed * 5.0f)
-            owner->setImpulseRequest(1.0f);
-        else
-            owner->setImpulseRequest((distance - keep_distance) / owner->impulse_max_speed * 5.0f);
-        if (rotation_diff > 90)
-            owner->setImpulseRequest(-owner->impulse_request);
-        else if (rotation_diff < 45)
-            owner->setImpulseRequest(owner->impulse_request * (1.0f - ((rotation_diff - 45.0f) / 45.0f)));
+        // setImpulseRequest only if impulse_max_speed is greater than 0.0
+        if (owner->impulse_max_speed > 0.0f)
+        {
+            owner->setImpulseRequest(0.0f);
+        } else {
+            if (distance > keep_distance + owner->impulse_max_speed * 5.0f)
+                owner->setImpulseRequest(1.0f);
+            else
+                owner->setImpulseRequest((distance - keep_distance) / owner->impulse_max_speed * 5.0f);
+
+            if (rotation_diff > 90)
+                owner->setImpulseRequest(-owner->impulse_request);
+            else if (rotation_diff < 45)
+                owner->setImpulseRequest(owner->impulse_request * (1.0f - ((rotation_diff - 45.0f) / 45.0f)));
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/ai/ai.cpp
```diff
@@ -599,13 +599,13 @@ void ShipAI::flyTowards(glm::vec2 target, float keep_distance)
             keep_distance = 0.0;
 
         if (distance > keep_distance + owner->impulse_max_speed * 5.0f)
-            owner->impulse_request = 1.0f;
+            owner->setImpulseRequest(1.0f);
         else
-            owner->impulse_request = (distance - keep_distance) / owner->impulse_max_speed * 5.0f;
+            owner->setImpulseRequest((distance - keep_distance) / owner->impulse_max_speed * 5.0f);
```
oznogon, line 604: This retains the potential division by zero, but now appears to be clamped by `setImpulseRequest()`. The condition here might be changed to prevent division by zero, but it doesn't seem to cause other problems, so I kept the existing behavior.

oznogon:
Added a check to confirm that impulse_max_speed is greater than 0 before attempting setImpulseRequest.

PR #1935, opened 2023-03-04
[GuiListbox] Implement themed hover behavior on entries

oznogon:
Pass hover coordinates from GuiCanvas so elements like GuiListbox can manage hover state, then use those coordinates to implement hover state on GuiListbox entries.

Master:

image

image

With this PR:

image

image


daid:
I would like to merge this, but there is now a conflict due to another merge, and I cannot quickly see how to fix the conflict.

It's not a perfect solution, but it will work for now. (In SeriousProton2, I get mouseMove events, and then you don't have the issue where if you hover over an item below a semi-transparent window that it will show as hovered)

PR #1938, opened 2023-03-11
Make the ship's log bigger

csibbitt:
Originally submitted as a naïve constant tweak that caused UI performance problems, UI usability problems, and network performance problems.

With @oznogon 's help and some additional work in the engine, these are all fixed. Changes in SeriousProton were rejected while it was migrated to the ECS codebase. Now that ECS has merged and the log code has been adjusted, this change is ready to go.

This patch depends on:
- https://github.com/daid/EmptyEpsilon/pull/2299
~- https://github.com/daid/SeriousProton/pull/239~
- https://github.com/daid/EmptyEpsilon/pull/2019
- https://github.com/daid/EmptyEpsilon/pull/2021
- https://github.com/daid/EmptyEpsilon/pull/2022

### DIFF
--- a/src/components/shiplog.cpp
+++ b/src/components/shiplog.cpp
@@ -9,9 +9,9 @@ void ShipLog::add(const string& message, glm::u8vec4 color)
 
 void ShipLog::add(const string& prefix, const string& message, glm::u8vec4 color)
 {
-    // Cap the ship's log size to 100 entries. If it exceeds that limit,
+    // Cap the ship's log size to 10,000 entries. If it exceeds that limit,
     // start erasing entries from the beginning.
-    if (entries.size() > 100)
+    if (entries.size() > 10000)
         entries.erase(entries.begin());
 
     // Timestamp a log entry, color it, and add it to the end of the log.


travishein:
👍

oznogon:
Running with a client in the server (single EmptyEpsilon.exe instance), 16-core Ryzen 9/RTX 3060, with a script that sent 100 entries upon scenario launch and one every 0.1 seconds after:

```lua
dock1 = PlayerSpaceship():setTemplate("Atlantis"):setPosition(-100, 0):setFaction("Human Navy")
i = 1
timeout = 0.1

for j = 1, 100, 1
do
    dock1:addToShipLog("Ack" .. i, "yellow")
    i = i + 1
end

function update(delta)
    timeout = timeout - delta
    if(timeout < 0)
    then
        dock1:addToShipLog("Ack" .. i, "yellow")
        i = i + 1
        timeout = 0.1
    end
end
```

By around 1,000 entries, the GuiSlider lozenge becomes too small to reliably use with touch:

image

By around 1,500 entries, the log begins to cause my to dip below 60fps on the Ship's Log screen.

By around 2,000 entries, the lozenge becomes too small to click.

image

By around 2,750 entries, the client is sending 500kb over the wire per frame.

By around 3,000 entries, FPS drops to below 30fps.

image

At 3,500 entries, CPU fans start to spin up to handle the load and FPS drops to under 25.

image

At 4,000 entries, FPS is at 21 and the lozenge is nearly obscured by the scrollbar arrow.

image

At 4,500 entries we start flirting with 1MB over the wire per second.

image

Resource Monitor confirms this.

image

At this point I open up a client over the network on Relay.

With the Ship's Log minimized on Relay, we still get 140+ fps.

With it expanded, fps drops to 15.

image

I had to stop around 7,000 because my CPU was pinned and reporting 90C temps. FPS was around 12 and system performance outside of EE was beginning to be impacted.

- GuiSlider needs a minimum touch-friendly lozenge size
- Viewing the Ship's Log with large numbers of entries should not impact client or server fps
- At some point for network replication, log entries needs to be paginated, cacheable, or truncated for clients

oznogon:
The ship log (the entire log, I think) is replicated to clients every frame. Not sure but I think this means it's replicated even to players who aren't using Relay/Ship's Log/any screen with access to the log? https://github.com/daid/EmptyEpsilon/blob/eb9ee8cf954977514bace1e74528008500fe7862/src/spaceObjects/playerSpaceship.cpp#L581

A dirty fix would reduce the replication frequency to 1.0 (1 second) or so. A better fix would send the entire log once to a new client and only send updates after that, but that would be a more dramatic implementation change.

oznogon:
Tried sending 10000 entries to the player upon launch, instead of sending entries constantly, and I'm not seeing the replication. Still not great if each new entry replicates the entire log. Either way, FPS with the ship's log visible drops to <9 with 10,000 entries and stays there.

image

@daid Does the log get replicated each frame or just on changes?

src/spaceObjects/playerSpaceship.cpp
```diff
@@ -1117,9 +1117,9 @@ void PlayerSpaceship::setRepairCrewCount(int amount)
 
 void PlayerSpaceship::addToShipLog(string message, glm::u8vec4 color)
 {
-    // Cap the ship's log size to 100 entries. If it exceeds that limit,
+    // Cap the ship's log size to 10000 entries. If it exceeds that limit,
```
oznogon, line 1120:
```suggestion
    // Cap the ship's log size to 1000 entries. If it exceeds that limit,
```

src/spaceObjects/playerSpaceship.cpp
```diff
@@ -1117,9 +1117,9 @@ void PlayerSpaceship::setRepairCrewCount(int amount)
...
     // start erasing entries from the beginning.
-    if (ships_log.size() > 100)
+    if (ships_log.size() > 10000)
```
oznogon, line 1122: Suggest reducing to 1,000, which is still a 10-fold increase, or less. Barring any changes to replication behavior, log rendering, or the GuiSlider behavior, this may be the practical limit without significantly degrading the log UI or performance. Naive testing suggests this still increases packet size each log update by ~20kB vs. 100 (vs. 200kB+ increase at 10000). Scenarios with noisier logs, like those with lots of CpuShips reporting order changes and status, might be more affected than others.

```suggestion
    if (ships_log.size() > 1000)
```

oznogon:
daid via Discord:

> yes, it's extremely inefficient, which is why it is so limited in size
> it's not just that the whole log gets sent, it's also checking every entry for changes each tick.
> ... I'm planning to change this on with the ECS code

Re: GUI performance issues with >1,500 entries:

> My guess, the minimized (view in Relay) just grabs the last line and displays that. But the expanded version first makes a "preparedfontrenderer" (call) for the all the lines, so it knows how much vertical space there is, and then draws a subsection of that.

oznogon:
Found another GuiScrollbar limitation: after 400 entries, the bottom entries in the log can't be viewed by scrolling, and the granularity of scrolling via scrollbar arrows is 1px/click, or too little to be useful. https://github.com/daid/EmptyEpsilon/issues/1940#issuecomment-1465271654

csibbitt:
@oznogon Thank you very much for testing this. I was expecting someone to explain the history of why it is like this, possibly with these sorts of details, but you did it with science!  It seems we have at least a UI issue and a performance issue to deal with before this could become a reality. I'll check out the discord discussion and see what I can do to help.

csibbitt:
Not sure if the CI system supports dependent changes, but this definitely won't even compile until https://github.com/daid/SeriousProton/pull/239 merges, and is not very usable until the other two required changes merge.

csibbitt:
With all of the changes in place, I was able to run the tests described here: https://github.com/daid/SeriousProton/pull/239#issuecomment-1763273428

csibbitt:
I've re-tested this on the new ECS code. After 1 adjustment (already merged), it worked just fine.

* Set the ShipLog entry limit to 10,000
* No longer need the SeqVector patch
* Scenario producing 10 logs/s
* Connected 2 new clients, full back replication worked
* Start and end of log work as expected before and after 10,000 on all 3 sessions
* Connected a 3rd client, full replication of 10,000 lines occurs
* Debug meter shows <=0.4 kbps per client (from the 10 logs/s)
* CPU and memory nominal and stable

image

What do you think, @daid - It's been two years, but I've never given up hope! 😆

daid:
images

PR #1943, opened 2023-03-13
[EE_GuiLIB] Align onDraw argument name definitions to implementations

oznogon:
Several GUI element defintions use "target" or "window" for onDraw function RenderTarget names, but implementations consistently use "renderer". Fix this mismatch by consistently using "renderer" in definitions.


daid:
This was just leftover from SFML times :)

Issue #1939, opened 2023-03-11
[GuiAdvancedScrollText] Large number of entries degrades render performance


oznogon:
PR #1938 revealed that a GuiAdvancedScrollText element with a large-enough number of entries (ca. 1,000-1,500 in my testing) sees performance degrade dramatically.

For me, an expanded Ship's Log view dipped from 100+ fps at 100 entries to under 60 fps at 1,500, 30 fps at 3,000, 20 fps at 4,500, and 9 fps at 10,000.

Relay's single-line log preview wasn't affected, staying at 140+ fps even with rapid updates, so despite the other performance issues in #1938 this seemed isolated to the scrolling view.

csibbitt:
This can be closed.

PR #1946, opened 2023-03-18
[GuiAdvancedScrollText] Improve performance

oznogon:
Re: #1939

This refactors GuiAdvancedScrollText to improve render performance.

- Font prep is moved to addEntry, and FontPreparedStrings are stored in the entries. This means fonts for an entry are prepared once upon the entry's addition, not each frame. This improves render performance, especially with large texts.
- Font prep is repeated across all entries only if the window width changes.
- Content outside the bottom of the scroll rect is ignored on the text-rendering step, rather than rendered and then clipped. This improves render performance when scrolling to the top of the text.
- max_prefix_width is calculated upon entry addition rather than each frame.

Testing on a variety of systems with a 10,000-entry log, this maintains:

- 20fps on a 12th-gen/16-thread i5 on Intel graphics, with power-saving enabled (previously: 2.8)
- 40fps on a 12th-gen/16-thread i5 on Intel graphics (previously: 4)
- 51fps on a 16-thread Ryzen 9/RTX 3060 (previously: 9)

Known issues:

- The latest-line view on Relay doesn't render correctly upon screen load. Expanding and minimizing the log view on Relay resolves this.
- Resizing the window with a large text is as expensive as before this PR, since this is now the only time the entire entries list is prepared at once.
- max_prefix_width is retained even if the largest prefix-width entry is removed from the log. This is potentially trivial to fix.

### DIFF
--- a/src/gui/gui2_advancedscrolltext.cpp
+++ b/src/gui/gui2_advancedscrolltext.cpp
@@ -24,16 +24,16 @@ unsigned int GuiAdvancedScrollText::getEntryCount() const
     return entries.size();
 }
 
-string GuiAdvancedScrollText::getEntryText(int index) const
+string GuiAdvancedScrollText::getEntryText(unsigned int index) const
 {
-    if (index < 0 || index >= int(getEntryCount()))
+    if (index >= getEntryCount())
         return "";
     return entries[index].text;
 }
 
-GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(int index)
+GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(unsigned int index)
 {
-    if (index < 0 || index > int(getEntryCount()))
+    if (index > getEntryCount())
         return this;
     entries.erase(entries.begin() + index);
     return this;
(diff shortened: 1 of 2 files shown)


oznogon:
Ran out of spare time on this, so I'm posting it up in case someone can improve on it.

src/gui/gui2_advancedscrolltext.cpp
```diff
@@ -44,40 +48,51 @@ GuiAdvancedScrollText* GuiAdvancedScrollText::clearEntries()
...
-    {
-        auto prepared = sp::RenderTarget::getDefaultFont()->prepare(e.prefix, 32, text_size, {0, 0}, sp::Alignment::TopLeft, 0);
-        max_prefix_width = std::max(max_prefix_width, prepared.getUsedAreaSize().x);
-    }
+    const bool is_resized = rect_width != rect.size.x;
+    if (is_resized) { rect_width = rect.size.x; }
 
     //Draw the visible entries
-    float draw_offset = -scrollbar->getValue();
+    float draw_offset = -scrollbar->getValue() + text_size + 12.0f;
```
oznogon, line 55: This is a bad kludge to realign text now that the positioning lives in the `Entry` rather than recalculated each onDraw.

oznogon:
#1947 was caught while doing this. An example of a potential fix involves checking the prefix text as well as the entry text, which requires adding a `getEntryPrefix()` function to GuiAdvancedScrollText. That would limit collisions to only entries with the same text at the same time in the ship's log. Another fix would be adding a UID to entries and checking that instead of the text.

PR #1958, opened 2023-03-25
[Planet] Update collision size on setPlanetRadius, setDistanceFromMovementPlane

### LINKED ISSUE #1957: [Planet] Collision radius isn't updated with object radius
Steps to reproduce:

1. Launch EE with scenario_10_empty.lua and the HTTP API enabled.
2. Open the GM station and select the large blue planet.
3. Run:

   ```lua
   planet = getGMSelection()[1]
   planet:setPlanetRadius(10000)
   ```
4. Create a ship and put it south of the resized planet.
5. Unpause the scenario.

Expected behavior:

- The planet collides with the ship and pushes it.
- On the GM screen and player radar, the planet's blue circle reflecting its collision radius is resized.

Observed behavior:

- The ship passes through and into the planet.
- The blue circle doesn't change in size.

image

Likely cause:

In planet.cpp this sets the collision radius:

```cpp
void Planet::updateCollisionSize()
{
    setRadius(planet_size);
    if (std::abs(distance_from_movement_plane) >= planet_size)
    {
        collision_size = -1.0;
    }else{
        collision_size = sqrt((planet_size * planet_size) - (distance_from_movement_plane * distance_from_movement_plane)) * 1.1f;
        setCollisionRadius(collision_size);
        setCollisionPhysics(true, true);
    }
}
```

but it only seems to be triggered in `update()` and only if `collision_size == -2.0f`, which is the default value upon instantiation.

```cpp
void Planet::update(float delta)
{
    if (collision_size == -2.0f)
    {
        updateCollisionSize();
        if (collision_size > 0.0f)
            PathPlannerManager::getInstance()->addAvoidObject(this, collision_size);
    }
```

oznogon:
When changing the planet radius with setPlanetRadius or z-coordinate with setDistanceFromMovementPlane, the planet's collision radius doesn't change. This updates the collision radius in setPlanetRadius and setDistanceFromMovementPlane.

Fixes #1957.

### DIFF
--- a/src/spaceObjects/planet.cpp
+++ b/src/spaceObjects/planet.cpp
@@ -201,6 +201,7 @@ Planet::Planet()
     registerMemberReplication(&orbit_target_id);
     registerMemberReplication(&orbit_time);
     registerMemberReplication(&orbit_distance);
+    registerMemberReplication(&collision_size);
 }
 
 void Planet::setPlanetAtmosphereColor(float r, float g, float b)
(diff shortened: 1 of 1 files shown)


daid:
Wont work correctly for multiplayer, clients keep the old collision size

oznogon:
Networked server (top) and client, default 2U radius:

image

10U radius:

image

Collision occurring on both:

image

The client doesn't get the new radius for the radar collision size but does get the planet radius in 3D views, and the collision occurs on the server side.

If the only thing the client radar uses to draw the circle is the collision radius, and the client already gets the planet radius and z-coordinate for 3D rendering, and the collisions happen on the server, then does the client really need the collision radius?

Can it just draw the planet radius, or derive the collision radius from the planet radius the same way as the server does from the planet radius?

oznogon:
Pushed a replication commit, and a local recalculation in drawOnRadar in [oznogon/1957-update-planet-collisionsize-locally](https://github.com/oznogon/EmptyEpsilon/tree/1957-update-planet-collisionsize-locally).

oznogon:
Closed for #1960.

PR #1965, opened 2023-04-09
[readme] Link ARM v8 Android build, note emulation

oznogon:
Add a link to the v8 build instructions, and note that the ARM builds will install and work in Android Emulator system images that support ARM ABIs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Official releases for Windows, Linux (as a .deb package), and Android (beta qual
 
 -   Windows releases are distributed as self-contained ZIP archives that don't need installation. You can expand the ZIP archive and launch EmptyEpsilon directly from the expanded folder.
 -   The .deb package requires freetype and SDL2 packages to be installed on your Linux distribution.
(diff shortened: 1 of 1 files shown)


oznogon:
Edited to clarify that this refers to the official APK, which is built only for ARM. If we start building official APKs that include x86, we can change that.

Added x86 build instructions to the wiki and linked to it as an option.

daid:
The usage of x86 android is very low:
https://stackoverflow.com/a/66039865

So I don't see a reason to have an official build for that. It's man use case is debugging with emulators on desktop I guess.

oznogon:
That link predates support for running Android apps on ChromeOS, [most of which are x86 or x86_64](https://www.chromium.org/chromium-os/developer-information-for-chrome-os-devices/). Due to the spotty support for Linux dev environments on Chromebooks and complexity of building, APKs are probably the most accessible way to get EE on a ChromeOS device.

While x86 Chromebooks can translate ARM code, it degrades performance and increases battery drain, and translating ARM v8 support is poor. Unity suggested the performance penalty of ARM translation was [25-30%](https://forum.unity.com/threads/when-to-enable-chrome-os-support-and-how-to-manage-aab-apk.1193830/). So [Google still suggests](https://chromeos.dev/en/games/optimizing-games-publishing) shipping APKs supporting ARM v7, v8, x86, and x86_64 ABIs for Chromebook support.

PR #1969, opened 2023-04-20
Fix underlinking by linking against stdc++

Flowdalic:


daid:
> No description provided.

No description provided.

PR #1970, opened 2023-04-27
Scenario 42 Shipyard Inspection

Xansta:
Introductory scenario for single player ship using either of two new player ship templates.
For anyone testing this, the client and the server need these files:
scripts/shiptemplates/corvette.lua
resources/radar/corsair.png


Xansta:
Any progress on evaluating this scenario, @daid ?
@muerteFR has started looking at it from a translation perspective.

PR #1972, opened 2023-05-05
Scenario 32 Planetary Devourer

Xansta:
A scenario celebrating the spirit of "May the 4th be with you."

Any number of player ships may join. The mission may be replayed often since the terrain differs each time the scenario runs. It takes about an hour to run.

The mission outline remains the same: prevent the dreaded "Planet Devourer" from destroying all five planets. The Kraylor have been secretly developing an "ultimate weapon" - a weapon strong enough to destroy an entire planet. Now they are testing it on innocent planets for maximum fear impact. The players are to assess the weapon and destroy it if possible. There are mission plot points to find the technical readouts or plans for the planet devourer, analyze the plans for a weakness, and exploit the weakness.


daid:
Kinda sad you didn't call it a "Murder Moon", even if "Planet Devourer" is a cool name, "Murder Moon" is even more a wink towards starwars.

daid:
With these large scenario scripts, is there a lot of copy&paste between the different scenarios? I am starting to wonder if there is some way to organize this better. 8000 line scripts are a bit of a red flag in code maintenance.

Xansta:
There is indeed a fair amount of copy and paste between scenario scripts. I abstracted some functionality related to stations into the place_station_scenario_utility.lua. There are other possibilities for reducing the size of scenarios through abstraction of common functionality, but I have not yet gone back to refactor along those lines.

PR #1979, opened 2023-05-16
added 100% _() scenario scavenger (3th part) and several Xansta's scenarios, create scavenger.en.po and fr.po

muerteFR:


muerteFR:
@daid ,this pull request can be pushed.

PR #1981, opened 2023-05-19
show playership descriptions on ship selection screen (server side)

aBlueShadow:
While not having the descriptions of playerships in the database is intentional, having them available outgame might still be a good idea.
And with the "new", leaner layout of the ship selection screen, there is enough empty space to show some additional text:
eedescription
It shows always the text of the ship that is currently selected on the ship spawn selector, so it is only visible on the server.

### DIFF
--- a/scripts/shiptemplates/OLD.lua
+++ b/scripts/shiptemplates/OLD.lua
@@ -387,7 +387,8 @@ template:setJumpDrive(true)
 template:setDockClasses(_("class", "Starfighter"), _("class", "Frigate"), _("class", "Corvette"))
 template:setSharesEnergyWithDocked(true)
 
-variation = template:copy("Ender"):setType("playership"):setLocaleName(_("playerShip", "Ender"))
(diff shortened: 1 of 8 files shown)


scripts/shiptemplates/OLD.lua
```diff
@@ -387,7 +387,8 @@ template:setJumpDrive(true)
 template:setDockClasses(_("class", "Starfighter"), _("class", "Frigate"), _("class", "Corvette"))
 template:setSharesEnergyWithDocked(true)
 
-variation = template:copy("Ender"):setType("playership"):setLocaleName(_("playerShip", "Ender"))
+variation = template:copy("Ender")--:setType("playership"):setLocaleName(_("playerShip", "Ender"))
```
daid, line 390: you removed the playership type here?

aBlueShadow: ah, dang! that was just meant as a temporary test to easyly compare both ships in the database, in preparation to write the description for the ender. Will fix it

PR #1982, opened 2023-05-20
New scenario "kessler"

aBlueShadow:
Save the global satellite network by catching space debris and prevent the Kessler Syndrome!

This is a beginner scenario initially created to be played by a crew of school kids. The officer(s) on relay&science or on operations should be able to read fluently.
It is a very combat-light scenario. Ideally, not a single ship should be destroyed and the players have no missiles whatsoever, as that would only create more space junk!


aBlueShadow:
Thanks to the extensive feedback from @muerteFR , I just added several small polishing fixes and additions.

PR #1985, opened 2023-05-21
Translation 100% scenario_47_scavenger.fr.po

muerteFR:


muerteFR:
@daid It's ok :-)

scripts/scenario_47_scavenger.lua
```diff
@@ -5268,7 +5268,7 @@ function handleDockedState()
 					addCommsReply(_("ridExuari-comms", "Add missile tubes"), function()
 						missileTubeUpgrade(comms_source)
 						setCommsMessage(_("ridExuari-comms", "Thanks for continuing to shoot down those Exuari. We've added some missile tubes to help you destroy the Exuari station"))
-						plot1_message = string.format(_("ridExuari-comms", "%s has asked for help against Exuari forces and has provided your ship with missile weapons to help you destroy %s"),first_station:getCallSign(),exuari_harassing_station:getCallSign())
+						plot1_message = string.format(_("ridExuari-", "%s has asked for help against Exuari forces and has provided your ship with missile weapons to help you destroy %s"),first_station:getCallSign(),exuari_harassing_station:getCallSign())
```
daid, line 5271: this context marker looks off? With the dash at the end.

muerteFR: No, it's just that I don't know if this text appears in the coms, the shiplogs... the 1st part of the context relates to the history of the scenario and the second indicates the place where the text appears. With these old Xansta scénario, it's harder for me to identify certain texts. anyway it will be necessary later, when xansta has the time, to adapt the settings to be able to translate them and I will also have to proofread the translation by testing this scenario. :-)

PR #1989, opened 2023-06-24
Enable auto connect for ship window

aBlueShadow:
This let you setup auto connect for ship windows.
It reserves the 'autoconnect' values from 1000 to 1360 for ship windows with angles between 0° and 360°.

I opted for this way rather than adding a new ECrewPosition value, as ship windows are, like main screen, a special case and handled differently. Also we don't need an extra option just for the angle that way, and it is unlikely that we'll ever get to more than 1000 different stations :-)

### DIFF
--- a/src/menus/autoConnectScreen.cpp
+++ b/src/menus/autoConnectScreen.cpp
@@ -6,6 +6,7 @@
 #include "playerInfo.h"
 #include "multiplayer_client.h"
 #include "multiplayer_server_scanner.h"
+#include "screens/windowScreen.h"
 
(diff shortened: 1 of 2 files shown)


daid:
Yeah, I like this more then the crewposition method, not sure if that pull request is still open...

aBlueShadow:
oh.. I actually missed or forgot about #1837, and #1816 for that matter - so that one would be fixed now. I just mentioned ECrewPosition to address the obvious question why I have not done it that way.

kpanic23:
That's perfect! And even more elegant! Thanks!

PR #1991, opened 2023-07-02
Add focusdelete parameter to text entry fields

aBlueShadow:
This adds an optional boolean parameter to text entry fields. If true, the content of the text field is deleted when it gains focus. This is useful when it is very likely that the user wants to type in a completely new value instead of editing the old one, like numbers for example.

### DIFF
--- a/src/gui/gui2_textentry.cpp
+++ b/src/gui/gui2_textentry.cpp
@@ -3,12 +3,13 @@
 #include "clipboard.h"
 
 
-GuiTextEntry::GuiTextEntry(GuiContainer* owner, string id, string text)
+GuiTextEntry::GuiTextEntry(GuiContainer* owner, string id, string text, bool focusdelete)
 : GuiElement(owner, id), text(text), text_size(30), func(nullptr)
 {
     blink_timer.repeat(blink_rate);
     front_style = theme->getStyle("textentry.front");
     back_style = theme->getStyle("textentry.back");
+    delete_on_focus = focus;
 }
 
 GuiTextEntry::~GuiTextEntry()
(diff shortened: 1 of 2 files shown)


daid:
There are two more common ways this is done:
* Select all on focus, gives the option to start typing and erase what was there, or to use the cursor keys to edit the existing value.
* Placeholder text that only shows if the text field is empty: https://www.w3schools.com/tags/att_input_placeholder.asp

Both methods are less "destructive" to the current value.

Also, following the API design of these controls, it shouldn't be a boolean parameter to the constructor but in the same way as `setMultiline`.

aBlueShadow:
The initial idea of this request came from the thoughts of how I could replace the fiddly ship window slider (almost impossible to set some specific values) with a text input field - without make the editing feel awkward.

https://github.com/daid/EmptyEpsilon/assets/25465934/956d569f-cd2d-4f78-81a2-737f29919d99

I guess both of your suggested variants would work in this case, as you usually won't edit the angle more than once.

aBlueShadow:
> Also, following the API design of these controls, it shouldn't be a boolean parameter to the constructor but in the same way as setMultiline.
As doing so in another commit won't make make sense (a diff on that PR would be larger than on master), I made a new PR: #1992

daid:
Ok, then I'm closing this one, the other one is more to my liking.

Also, just a general bit of knowledge. boolean parameters to functions are considered bad API design, as you cannot see at the callee what is happening. `function(True, False, False, True)` isn't very obvious. A different way then these `set` functions would be to have flags, like the SDL2 API has https://wiki.libsdl.org/SDL2/SDL_CreateWindow (or the QT API)

Not that we need to go around and change all the current code, but it's good to keep in mind when writing code in general. I made quite a few "mistakes" while making EE, and learned a lot ;-)

PR #1994, opened 2023-07-07
convert window angle input from slider to textentry

aBlueShadow:
The current solution for ship window angles has a big problem: Precise settings are virtually impossible. The possible steps depend on the current aspect ratio, but even on ultra wide screen it is very fiddly. This is especially an issue if you want to have two windows next to each other. Or on a rear window that feels slightly off because you can't set it to exactly 180°.
So I replaced the slider with a text input. 
windowangle

Downside of this solution is that you need a keyboard, but I don't think this is a big issue. On machines without a keyboard, you might want to set up auto connect anyway. And on android, the on-screen-keyboard works just fine.


Xansta:
Hooray! I've been looking for a way to get exactly 90 and 270 degree windows for a while

PR #2002, opened 2023-07-16
Improve weapon tube function for calculating firing solutions.

andrewchimkh:
Proposed improvement to the WeaponTube::calculateFiringSolution function for faster and more reliable convergence:
- Direct calculation of turning angle from aim point.
- Improved calculation of interception point / nearest approach point (from TODO).
- Harmonization of methods for moving and stationary targets.

Contact Chimp on discord for more information.


andrewchimkh:
BUG: Function does not work well when the parent ship is moving. Relaunching the pull request.

PR #2003, opened 2023-07-16
Improved function for calculating firing solutions (relaunched).

### LINKED ISSUE #1996: Missile trajectory lines glitching
Issue:
I often notice that the missile trajectory lines are glitching (side-mounted launchers). The lines are flickering between showing the proper aim trajectory and the default non-aimed line of a side-mounted launcher. And its not just corner-cases where the trajectory just loses LOS/aim. Very annoying. 
Expected: Proper trajectory shown without flickering between states.

andrewchimkh:
Proposed improvement to the WeaponTube::calculateFiringSolution function for faster and more reliable convergence:
- Direct calculation of turning angle from aim point.
- Improved calculation of interception point / nearest approach point (from TODO).
- Harmonization of methods for moving and stationary targets.

Relaunched pull request because the previous one had a bug that made the firing solution incorrect when the parent ship is moving. Bug is now corrected. Contact Chimp on discord for more information.

Fixes #1996

### DIFF
--- a/src/spaceObjects/spaceshipParts/weaponTube.cpp
+++ b/src/spaceObjects/spaceshipParts/weaponTube.cpp
@@ -306,63 +306,108 @@ string WeaponTube::getTubeName()
     return "?" + string(direction);
 }
 
+// Coordinate transform from global reference frame (X: right/east, Y: down/south)
+// to weapon tube reference frame (X: right of tube, Y: direction of fire)
+static glm::vec2 transform_global_to_tube(glm::vec2 vec, float angle)
+{
+    const float heading = angle + 0.5f*float(M_PI);
+    const float c = glm::cos(heading);
+    const float s = glm::sin(heading);
+    return glm::vec2(c*vec.x + s*vec.y, s*vec.x - c*vec.y);
+}
+
 float WeaponTube::calculateFiringSolution(P<SpaceObject> target)
 {
-    if (!target)
-        return std::numeric_limits<float>::infinity();
-    const MissileWeaponData& data = MissileWeaponData::getDataFor(type_loaded);
-    if (data.turnrate == 0.0f)  //If the missile cannot turn, we cannot find a firing solution.
-        return std::numeric_limits<float>::infinity();
-
-    auto target_position = target->getPosition();
-    auto target_velocity = target->getVelocity();
-    float target_velocity_length = glm::length(target_velocity);
-    float missile_angle = vec2ToAngle(target_position - parent->getPosition());
-    float turn_radius = ((360.0f / data.turnrate) * data.speed) / (2.0f * float(M_PI));
-    float missile_exit_angle = parent->getRotation() + direction;
-
-    for(int iterations=0; iterations<10; iterations++)
+    const MissileWeaponData& missile = MissileWeaponData::getDataFor(type_loaded);
+    float missile_angle;
+    if (!target || missile.turnrate == 0.0f)
     {
-        float angle_diff = angleDifference(missile_angle, missile_exit_angle);
-
-        float left_or_right = 90;
-        if (angle_diff > 0)
-            left_or_right = -90;
-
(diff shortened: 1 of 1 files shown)


src/spaceObjects/spaceshipParts/weaponTube.cpp
```diff
@@ -306,63 +306,108 @@ string WeaponTube::getTubeName()
     return "?" + string(direction);
 }
 
+// Coordinate transform from global reference frame (X: right/east, Y: down/south)
+// to weapon tube reference frame (X: right of tube, Y: direction of fire)
+static glm::vec2 transform_global_to_tube(glm::vec2 vec, float angle)
```
daid, line 311: https://github.com/daid/SeriousProton/blob/master/src/vectorUtils.h#L20

src/spaceObjects/spaceshipParts/weaponTube.cpp
```diff
@@ -306,63 +306,108 @@ string WeaponTube::getTubeName()
...
+            // Calculate turn angle
+            if (turn_direction == 0.0f)
+            {
+                turn_angle = 0.0f;
+            }
+            else
+            {
+                const float x = turn_direction * aim_position.x;
+                const float aim_radius = glm::length(aim_position - turn_direction*glm::vec2(turn_radius, 0.0f));
+                const float a = glm::acos((turn_radius - x) / aim_radius);
```
daid, line 367: Instead of acos could atan2 be used? Or https://github.com/daid/SeriousProton/blob/master/src/vectorUtils.h#L15C26-L15C26
acos has a pretty high risk of a domain error causing NaN

daid (commented):
Not saying I fully understand the new code yet, I think I need to draw it out to follow it. But in general it looks good, just two nitpicks on code reuse and potential stability.

src/spaceObjects/spaceshipParts/weaponTube.cpp
andrewchimkh: I had considered using rotateVec2, and I would normally go with reusing a function as the preferable solution. But, I decided against it for the following reasons:
- rotateVec2 takes angles in degrees. But, there are various other operations inside the iterative loop where it would be much more convenient to keep the angles in radians by default. Thus, I thought it would be clearer if I consistently used radians inside the main body of the iterative loop, and only converted back into degrees at the very end.
- The game seems to work in a left-handed coordinate system, which I was struggling with during testing. So I wanted to switch to a right-handed system which I am more familiar with, for only the internal calculations of the function. This means a rotation + reflection transform. I thought it would be simpler to implement it directly instead of chaining rotateVec2 with a reflection operation.

src/spaceObjects/spaceshipParts/weaponTube.cpp
andrewchimkh: Good point. I will update the code to try this.

andrewchimkh:
Latest commit addresses feedback on acos behaviour. Remaining acos NaN risk in `calculateTurnAngle` is mitigated by the `d >= turn_radius` check.

andrewchimkh:
> Not saying I fully understand the new code yet, I think I need to draw it out to follow it. But in general it looks good, just two nitpicks on code reuse and potential stability.

Is there a particular part that you would like an explanation for?

daid:
Note that I haven't forgotten about this, I just haven't had the mental ability yet to dive into it. I do apricate the work you put into this a LOT!
I want to understand it before I merge it, also because I need to port it to the ECS branch. But overall it does look good now from a general "scan".

andrewchimkh:
Okay. I will work on a technical note to help explain how it works.

andrewchimkh:
[technical_note_firing_solution.pdf](https://github.com/daid/EmptyEpsilon/files/12223157/main.pdf)

Et voila !

daid:
Holy, that is a clever bit of geometry. I wouldn't have been able to pick that up directly from the code, but the picture at step 3 with the explanation was really helpful in understanding this.

PR #2007, opened 2023-08-02
create scenario_79_kessler.en.po and fr.po

muerteFR:


aBlueShadow:
You can tell how fast a translator is, when the translation made in into the game sooner than the content that is to be translated :-)

PR #2008, opened 2023-08-08
Update translation capability Planet Devourer scenario and create devourer.en.po and fr.po

muerteFR:
Seen with @Xansta 

1st part

- Added many string.format
- Adding translation tags
- Standardization of certain translation contexts


muerteFR:
I finished that PR @daid  :-)

Issue #1947, opened 2023-03-18
[ShipsLogControl] Log might not update if first entry text is the same as the latest


oznogon:
Steps to reproduce:

1. Send a log-filling set of `addToShipLog` calls to a PlayerSpaceship during a scenario via the exec.lua HTTP API endpoint:

```lua
for i=1,100,1
do
player:addToShipLog("ack", "yellow")
end
```
1. Send the exact same set again a few seconds later, after no other log activity on the player ship.

Expected behavior:

The newest entries in the log have the newer timestamp.

Observed behavior:

The log doesn't appear to update.

This appears to be due to checks like these in ShipsLogControl:

https://github.com/daid/EmptyEpsilon/blob/b3b6e9127de6cbbae566d0d9eaf07a0408e9b02b/src/screenComponents/shipsLogControl.cpp#L42-L47

Specifically:

```cpp
log_text->getEntryText(0) != logs[0].text
```

which returns true if two separate log entries from different times have identical text.

csibbitt:
This can be closed

PR #2018, opened 2023-10-10
[ShipsLogControl] Fix update failure with duplicate text

csibbitt:
* Add a sequence # to log messages
* Store seq in GuiAdvancedScrollText along with rest of log data
* Compare seq instead of text when sync'ing ScrollText to log
* Fixes Issue #1947

I'm hoping this sequence # will also come in useful when I look at optimizing the network sync of the ship's log.


daid:
Looks good enough for now. On the ECS branch this will be handled a bit differently. But, we're not there yet :)

PR #2019, opened 2023-10-10
[GuiAdvancedScrollText] Performance improvements

csibbitt:
Fixes Issue #1939

Most of the work is from @oznogon , I just added some finishing touches. 

This refactors GuiAdvancedScrollText to improve render performance.

- Font prep is moved to addEntry, and FontPreparedStrings are stored in the entries. This means fonts for an entry are prepared once upon the entry's addition, not each frame. This improves render performance, especially with large texts.
- Font prep is repeated across all entries only if the window width changes.
- Content outside the bottom of the scroll rect is ignored on the text-rendering step, rather than rendered and then clipped. This improves render performance when scrolling to the top of the text.
- max_prefix_width is calculated upon entry addition/removal rather than each frame.

Testing on a variety of systems with a 10,000-entry log, this maintains:

- 20fps on a 12th-gen/16-thread i5 on Intel graphics, with power-saving enabled (previously: 2.8)
- 40fps on a 12th-gen/16-thread i5 on Intel graphics (previously: 4)
- 51fps on a 16-thread Ryzen 9/RTX 3060 (previously: 9)

Known issues:

- Resizing the window with a large text is as expensive as before this PR, since this is now the only time the entire entries list is prepared at once.

### DIFF
--- a/src/gui/gui2_advancedscrolltext.cpp
+++ b/src/gui/gui2_advancedscrolltext.cpp
@@ -24,16 +24,16 @@ unsigned int GuiAdvancedScrollText::getEntryCount() const
     return entries.size();
 }
 
-string GuiAdvancedScrollText::getEntryText(int index) const
+string GuiAdvancedScrollText::getEntryText(unsigned int index) const
 {
-    if (index < 0 || index >= int(getEntryCount()))
+    if (index >= getEntryCount())
         return "";
     return entries[index].text;
 }
 
-GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(int index)
+GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(unsigned int index)
 {
-    if (index < 0 || index > int(getEntryCount()))
+    if (index > getEntryCount())
         return this;
     entries.erase(entries.begin() + index);
     return this;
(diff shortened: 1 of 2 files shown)


src/gui/gui2_advancedscrolltext.h
```diff
@@ -29,8 +36,8 @@ class GuiAdvancedScrollText : public GuiElement
     GuiAdvancedScrollText* setTextSize(float text_size) { this->text_size = text_size; return this; }
 
     unsigned int getEntryCount() const;
-    string getEntryText(int index) const;
```
daid, line 32: I think most of the changes in this pull request are good. But I don't like taking unsigned ints here. I sort of understand the reasoning and that it most likely won't cause any problems. It's inconsistent with the rest of the code which pretty much never uses unsigned ints.

src/gui/gui2_advancedscrolltext.cpp
```diff
@@ -24,16 +24,16 @@ unsigned int GuiAdvancedScrollText::getEntryCount() const
...
+    if (index >= getEntryCount())
         return "";
     return entries[index].text;
 }
 
-GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(int index)
+GuiAdvancedScrollText* GuiAdvancedScrollText::removeEntry(unsigned int index)
 {
-    if (index < 0 || index > int(getEntryCount()))
+    if (index > getEntryCount())
```
csibbitt: I think this should be `>=`

src/gui/gui2_advancedscrolltext.h
csibbitt: I'll eliminate the usage for getEntryCount() too, then, since that was likely part of the motivation for aligning the others (they do a casted comparison to getEntryCount())

csibbitt: Neverminded, I found many other cases where it's compared to other vector::size()'s or unsigned local iterators. Are you SURE you don't like unsigned ints? 😆

daid: It's fine for local iterators. But I don't think they are used anywhere in API, and sometimes in casts when comparing with vector sizes. But in general, I try to keep numbers unsigned, so you don't have issues with checks like `if (n - 3 < x)`

PR #2021, opened 2023-10-12
[GuiAdvancedScrollText] Base scrollbar click change on text size

### LINKED ISSUE #1942: [GuiScrollbar] Arrow scroll amount can be too small
See https://github.com/daid/EmptyEpsilon/issues/1940#issuecomment-1465271654. If a scrollable element has a sufficiently large number of lines or entries, the arrows can be limited to scrolling by 1 pixel per click. This means scrolling one 32-px entry using the arrows can require 32 arrow clicks.

Due to #1941, scrolling via arrow is required to view the bottom of sufficiently large scrollable elements.

csibbitt:
- Fixes #1942 

Credit to @oznogon , I'm just submitting the PR. 

I tested this and it works well; much better than scrolling 1px/click.

### DIFF
--- a/src/gui/gui2_advancedscrolltext.cpp
+++ b/src/gui/gui2_advancedscrolltext.cpp
@@ -1,10 +1,12 @@
 #include "gui2_advancedscrolltext.h"
 
 GuiAdvancedScrollText::GuiAdvancedScrollText(GuiContainer* owner, string id)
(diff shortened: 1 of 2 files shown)


daid:
There seems to be some conflict, currently not at a PC where I can check it locally.

PR #2049, opened 2024-03-15
Another update to the Scurvy Scavenger scenario

Xansta:
- Rearrange code
- Use external scenario utilities
  - Station creation
  - Generate call sign
  - CPU ship diversification
  - Spawn ships (from GM buttons)
- Make scenario more explicitly single player ship, single type of player ship
- Update enemy ship spawn routines
- Make asteroid size variance larger

### DIFF
--- a/scripts/scenario_47_scavenger.lua
+++ b/scripts/scenario_47_scavenger.lua
@@ -14,85 +14,16 @@
 -- Murphy[Hard]: Random factors are more against you
 
 require("utils.lua")
-function createRandomAlongArc(object_type, amount, x, y, distance, startArc, endArcClockwise, randomize)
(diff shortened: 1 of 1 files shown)


daid:
There seems to be some conflict?

Xansta:
There is no mechanism in github to resolve conflicts, even when I click the "Resolve conflicts" button. It tells me there are 3 conflicts and scrolls to those places, but there's no way to pick which block to use. Does this have to be resolved outside of github (ir manually)?

muerteFR:
When this happened to me at the time, I directly modified the file manually from the PR, choosing which block to keep if there were few conflicts.
Maybe it's different now, I don't know.

PR #2060, opened 2024-03-30
Use Arrow Keys to move Crew in Engineering

parracodea:
Avoids duplicate keybinding in Engineering+, where `S` moves the Crew down, but also toggles Shields

---

For `Weapons` station, `S` toggles shields and in `Engineering`, `W A S D` moves a Crew Member. However in `Engineering+`, they have the ability to do both, so moving a Crew member also causes the shields to be turned on/off.

The default keybinds have been reassigned as follows:

Action | Original Key | New Key | Additional Note
|--|-|-|----------------------|
"ENGINEERING_POWER_100"             |  `Space`   |  `\\`          | Backslash to reset power levels since it's next to the increase/decrease keys (`[`, `]`)
"ENGINEERING_POWER_INCREASE"        |  `Up`      |  `]`           |
"ENGINEERING_POWER_DECREASE"        |  `Down`    |  `[`           |
"ENGINEERING_COOLANT_INCREASE"      |  `Left`    |  `=`           | `=` since its the same key as `+`
"ENGINEERING_COOLANT_DECREASE"      |  `Right`   |  `-`           | See note below
"ENGINEERING_REPAIR_CREW_NEXT"      |  `Q`       | `C`            | `C` for "Crew" to follow the convention where the letter matches the action name
"ENGINEERING_REPAIR_CREW_UP"        |  `W`       | `Up`           |
"ENGINEERING_REPAIR_CREW_DOWN"      |  `S`       | `Down`         |
"ENGINEERING_REPAIR_CREW_LEFT"      |  `A`       | `Left`         |
"ENGINEERING_REPAIR_CREW_RIGHT"     |  `D`       | `Right`        |

Note: There's a bug in SeriousProton where it's unable to read `-` as a keybind from `keybindings.json`. See https://github.com/daid/SeriousProton/pull/243 for a fix (which should probably be merged before this PR).


daid:
Thanks for all of these! It's the little things that are a lot of work and easy to miss.

PR #2066, opened 2024-04-06
Increase contrast in Science colored bands

### LINKED ISSUE #2064: Colored Bands for the Science station has poor accessability
The blue band in the Scanning mini-game and around the Science station is a bit difficult to see against a dark background.

image

I entered the colors into a [Contrast Checker site](https://webaim.org/resources/contrastchecker/), and it came back with a "FAIL":

image

In fact, Red failed as well; only Green passed:

image

image

It would be nice to improve the accessibility. Possible solutions would be to pick different colors (Cyan/Magenta/Yellow?), or choose a lighter shade of Red/Green/Blue.

parracodea:
Use lighter version of Red, Green, Blue for better accessibility

Aims to fix #2064 (Colored Bands for the Science station has poor accessibility)

# Before
original_sci_bands_rgb

# After
modified_sci_bands_rgb

# Accessibility for new colors
## Red
Accessibility for Red
## Green
Accessibility for Green
## Blue
Accessibility for Blue

### DIFF
--- a/src/screenComponents/rawScannerDataRadarOverlay.cpp
+++ b/src/screenComponents/rawScannerDataRadarOverlay.cpp
@@ -171,7 +171,7 @@ void RawScannerDataRadarOverlay::onDraw(sp::RenderTarget& renderer)
     a_b.push_back(a_b.front());
 
     // Draw each band as a line.
-    renderer.drawLineBlendAdd(a_r, glm::u8vec4(255, 0, 0, 255));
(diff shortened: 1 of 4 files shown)


daid:
Good improvement, I didn't like all colors I saw in #2064 but these are good.

Another accessibility issue is the thickness of the lines. On larger monitors they are still 1 pixel which can be hard to see. This is very much an issue if you use an android phone for example, where individual pixels are so small.

PR #2067, opened 2024-04-06
Use Mine icon on Radar, instead of "blip" icon

parracodea:
Mines also appear as the same color as Nukes (orange)

---

Many new Helms players don't know that a white dot on their radar is a Mine, so they fly over it (much to the rest of the crew's dismay). 

There doesn't seem to be a particular reason either in Lore or Gameplay for it to be a white dot ("Our radars can't tell if it's a mine or an asteroid, we need Science to scan it"), since the radar already differentiates between asteroids and mines, appearing as yellow-brown or white dots, respectively.

# Screenshots

For reference, here's what the new mine looks like next to a Nuke and some HVLIs:

feature_radar_mine_icon-0 5-size

And it'll also appear as a mine icon in the GM screen. (This is slightly older screenshot where I had the size be `0.4 * 32` instead of `0.5 * 32`. `0.4 * 32` started to lose some definition and the spikes around the mine, which is why I used `0.5 * 32`)
feature_radar_mine_icon-0 4-size-gm


aBlueShadow:
> Many new Helms players don't know that a white dot on their radar is a Mine, so they fly over it (much to the rest of the crew's dismay).

The tutorial tells about what a mine looks like.
So keep in mind that those changes should also be reflected in the tutorial.

parracodea:
I didn't think about updating the tutorial text - thanks for that!

nigelkennington:
Here is a suggested alternative art for the mine specific "blip"
mine

hemmond:
Usually it is harder to differentiate between mine and artifact/supply drop. The color/size changes are just too subtle. But I think it is better to just use different icon (as long as there are no scenarios that are depending on placed artifacts/supply drops in a crowd of mines. I'd lean towards changing the image and keeping the white color (and keep orange/blue to weapons of mass destruction/disruption), mines are IMHO less dangerous than EMP/Nuke.

parracodea:
> Usually it is harder to differentiate between mine and artifact/supply drop. The color/size changes are just too subtle. But I think it is better to just use different icon

That's probably a good idea - you should create a new Issue/Pull Request to discuss that

parracodea:
Most of the feedback from [the Discord discussion](https://discord.com/channels/206597597455056898/468130064513368064/1226796971541528636) thought that the icon change was a positive change, but some had issue with the size of the icon, saying that it was too large, but making it smaller may have issues with readability of the icon.

I've created a new mine icon that has slightly larger "spikes" (trying to matching the size of the crosshair lines for the Homing missiles), which should help with the readability as the icon is shrunken down. This updated icon appears on both the radar, as well as the "Missile Selection" buttons on the right.

This screenshot shows the new mine icon in the radar at size `0.4` (committed to the branch) as well as size `0.3`, with HVLI of Large/Medium/Small next to it for comparison.

new-mine-icon-0_3-and-0_4

parracodea:
Using the Mine icon looks messy at smaller resolutions (and the science console, where everything is more zoomed out) and the orange color conflicts with the color for Arlenians in the "Game Master" screen.


image
_Science Console Screenshot Resolution: 1000x750_

image

nigelkennington:
I understand this be too late as these are closed now, but as a final suggestion: would it be possible to include this but by default have the "mine.png" be a renamed copy of the existing "blip.png" so that if someone does want to replace them they can do it as part of a mod but those who prefer the traditional blip and arrow will be unaffected?

PR #2068, opened 2024-04-06
Use missile icon on radar instead of arrow icon

parracodea:
Unscanned ships and missiles both appear as arrow icons, which can be difficult to differentiate them at a quick glance. Missiles also have large-sized variants, which get closer to the same size as an unscanned ship.

# Before
feature_radar_missile_icon

# After
feature_radar_missile_icon_new


oznogon:
I like the idea of differentiating missile icons from ship icons, but I don't really like the specific icon proposed, especially at small sizes (being two separate shapes makes it "darker"/"noisy"/harder to spot) or for HVLIs (implies a missile with thrust and/or guidance, but isn't).

nigelkennington:
Here is a suggested alternative png for the "missile blip".
missile

nigelkennington:
Also, because I quite like the idea, here's a bonus one for hvli/unguided blips:
unguided

parracodea:
Closing this due to the icon having poor readability (especially the small missile), and difficulty determining the heading, among other reasons.

nigelkennington:
I understand this be too late as these are closed now, but as a final suggestion: would it be possible to include this but by default have the "missile.png" be a renamed copy of the existing "arrow.png" so that if someone does want to replace them they can do it as part of a mod but those who prefer the traditional blip and arrow will be unaffected?

parracodea:
I'd suggest opening a new Issue to discuss and track that separately from
this Pull Request.

PR #2088, opened 2024-05-05
Theme selection in option menu

Tsht:
- Add a translatable menu in Option / Interface called Interface Theme
- Themes containing unparsable lines are now considered invalid (see https://github.com/daid/SeriousProton/pull/244)
- Only valid theme are listed
- Use default.theme.txt as fallback in case of issue with a theme
- Use first found font as fallback font (not to crash)
- If no (valid) theme found, then EmptyEpsilon exits with an explicit error message.

Every theme found in gui, *.theme.txt, are loaded when entering option screen. This is primarily to test validation.


src/menus/optionsMenu.cpp
```diff
@@ -104,31 +105,77 @@ OptionsMenu::OptionsMenu()
     }))->setSize(GuiElement::GuiSizeMax, 50);
 
     //Select the language
-    (new GuiLabel(interface_page, "LANGUAGE_OPTIONS_LABEL", tr("Language (applies on back)"), 30))->addBackground()->setSize(GuiElement::GuiSizeMax, 50)->layout.margin.top = 20;
-    
-    std::vector<string> languages = findResources("locale/main.*.po");
-    
-    for(string &language : languages) 
     {
```
Tsht, line 108: Only added this bracket, github merger seems lost :p

daid: Indentation is different.

src/gui/gui2_canvas.cpp
```diff
@@ -1,13 +1,14 @@
 #include "gui2_canvas.h"
 #include "gui2_element.h"
 #include "theme.h"
+#include "preferenceManager.h"
```
daid, line 4: Overall, this the changes look decent. But I'm not sure about this specific part, not sure if I like it that the gui2_canvas depends on the preferencesManager. I think I will like it better if there is some class static or global that gets set from the "consumer" and keeps the gui2 code reasonably clean from dependencies.

src/menus/optionsMenu.cpp
Tsht: Thanks, found the option :)

src/gui/gui2_canvas.cpp
daid: Yeah, I like it better this way. Not that the whole EE codebase is a showoff of great code design :P

src/main.cpp
```diff
@@ -332,8 +343,8 @@ int main(int argc, char** argv)
     soundManager->setMusicVolume(PreferencesManager::get("music_volume", "50").toFloat());
     soundManager->setMasterSoundVolume(PreferencesManager::get("sound_volume", "50").toFloat());
 
-    main_font = GuiTheme::getTheme("default")->getStyle("base")->states[0].font;
-    bold_font = GuiTheme::getTheme("default")->getStyle("bold")->states[0].font;
+    main_font = GuiTheme::getCurrentTheme()->getStyle("base")->states[0].font;
+    bold_font = GuiTheme::getCurrentTheme()->getStyle("bold")->states[0].font;
```
daid, line 347: While this is a bit of a hack to get a lot of legacy code to work, these main_font and bold_font globals are now not updated when you change a theme until you restart the game.

Tsht: Herrrr you're right I may update this when updating preferences.

muerteFR:
@Tsht By theme, do you mean just changing the font in the future ?

muerteFR:
And, the translation tag was not added to be able to translate the word "Default". Can you add it?

Tsht:
@muerteFR Hello. Theme is a mechanism to modify some elements for the UI, look for instance in : https://github.com/daid/EmptyEpsilon/blob/master/resources/gui/default.theme.txt
You can copy this file to another file and edit, or edit it directly.

I don't know if it makes sense to translate "default" as it's just the file name... If you create "other.theme.txt", then you will have the choice to also select "other". 
Themes really don't have specific names, so I showed them in UI by this convention.
Translating this would be confusing I think ? And even then I don't know how you would make translations.

muerteFR:
Thank you for these explanations. :-)
For the translation, I was just talking about translating the word. “Default” is "Défaut" in French or even “Par defaut”. Not necessarily to translate the others who will arrive later and whose names will surely be proper nouns. This is weird, it's the only untranslated word in the options.
default

Tsht:
Maybe someting like "default (Default theme)" translated : "default (thème par défaut)" ?

muerteFR:
Yes it would be good like that.

daid:
problem is, this name comes directly from the filename, so that isn't translation aware. Not sure how to fix that yet.

muerteFR:
ok

daid:
But, if there is only 1 theme, I made this selector hidden, so per default you won't see it right now. Gives us some time to test this and think about how to translate it before it's default exposed.

aBlueShadow:
> This is weird, it's the only untranslated word in the options.
Actually, not quite. The language indicator is not translated either. Basically the same thing, in that it is just part of the filename(s). I think that works because many people are used to those indicator, so they serve as some kind of proper name. Also it makes it easier to change the language back, if you don't know what a language is called in another one.
One easy fix for the default theme, later when more themes are added, might be to make make it proper-name-ish, like "EE-default". That works for desktop themes, so I think it can work for EE as well.

Issue #1686, opened 2022-02-09
[Scenario issue] Training: Cruiser DB link hs with some non-existent ships


muerteFR:
I don't know if we can put here problems concerning scenarios.
The link to the DB for some ships doesn't work. Like "Exuari freighter", "Exuari transport", "Exuari Shuttle"... even in EN language.
Which seems logical since these ships do not/no longer exist in the DB.

muerteFR:
Hi @Piglit I I believe you created this scenario, I raise this problem. Could you fix it?

oznogon:
This isn't trivial because of how ships are added to the Science database and linked to from the Science radar, and it's hard to pin this on @piglit or any scenario writer because the tools aren't there to do this easily.

Today, EE populates the Ships database on scenario launch by parsing ShipTemplates from files in scripts/shiptemplates. The Exuari ships that don't show up in the database aren't in shiptemplates files or ShipTemplate objects, but are standard ships created with a custom type name by using `ShipTemplateBasedObject:setTypeName()`:

```
function createExuariTransport()
	return CpuShip():setFaction("Exuari"):setTemplate("Personnel Freighter 1"):setTypeName("Exuari transport")
end
```

Doing this breaks the DB link to Personnel Freighter 1, because that link relies on a string comparison of the type name. (#1271) But aside from that, we probably want new entries for the new ships anyway.

To minimally "fix" this, the scenario can create entries for those ship types in the Science database:

```lua
-- Get the Exuari Ships ScienceDatabase object
exuari_ship_db = queryScienceDatabase("Ships","Exuari")
-- Add the entry
exuari_transport = exuari_ship_db:addEntry(_("Exuari transport"))
-- Put something in it
exuari_transport:setLongDescription(_("Exuari transport description goes here"))
```

To copy the base ship template's full data also in that new db entry, you need a Lua function that duplicates as much of an existing ScienceDatabase under a new name as it can:

```
function copyShipDBEntry(base_ship_template,new_ship)
  ship_db = queryScienceDatabase("Ships")

  -- Search each Ships class's entry for the base ship
  for k,entry in ipairs(ship_db:getEntries())
  do
    base_ship_entry = entry:getEntryByName(base_ship_template)
    if base_ship_entry ~= nil
    then
      break;
    end
  end

  -- Create a new Exuari DB entry for the new ship
  exuari_ship_db = ship_db:getEntryByName("Exuari")
[...]

PR #2095, opened 2024-05-09
makes scenario_20_training1 playable again for non-english-languages

aBlueShadow:
This addresses https://github.com/daid/EmptyEpsilon/issues/1686 and especially https://github.com/daid/EmptyEpsilon/issues/1686#issuecomment-1666091533 , as that scenario became unplayable for non-english languages.

This PR does not try to fix the actual issue (which seems to be very persistent) but offers a workaround, to make it at least playable again. 
- Deletes all database references, notably the ones that copy the ship names and cause the issue.
- The custom ship name is transferred to the ship description for simple scan. So it now can be considered a description of how this particular ship is used, rather than an custom model name. Positive side effect: Other than ship names, this can safely be translated.

### DIFF
--- a/scripts/scenario_20_training1.lua
+++ b/scripts/scenario_20_training1.lua
@@ -84,7 +84,6 @@ There will be not much resistance, so you can try different approaches and tacti
 
 Here's your chance to beat up some helpless opponents.
 Commander Saberhagen out.]]))
-    globalMessage("Test")
     end
 end
 
--- a/scripts/scenario_20_training1.lua
+++ b/scripts/scenario_20_training1.lua
@@ -20,37 +20,6 @@ function createExuariWeakInterceptor()
 	return CpuShip():setFaction("Exuari"):setTemplate("Dagger"):setBeamWeapon(0, 0, 0, 0, 0.1, 0.1)
 end
 
-function copyShipDBEntry(base_ship_template,new_ship)
-  ship_db = queryScienceDatabase("Ships")
-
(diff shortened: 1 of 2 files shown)


scripts/scenario_20_training1.lua
```diff
@@ -135,6 +84,7 @@ There will be not much resistance, so you can try different approaches and tacti
 
 Here's your chance to beat up some helpless opponents.
 Commander Saberhagen out.]]))
+    globalMessage("Test")
```
daid, line 87: "Test"?

aBlueShadow: Oops - Sorry that was a debugging remnant, I just removed it.

daid:
I do wonder if there isn't just a simpler fix, where we say `if ship_db == nil then return end`. Which would make it the translated scenario doesn't have the DB entry and the base one does.

aBlueShadow:
That would still leave 3 other errors on lines with setModelDataName, though.

daid:
also fixable with nil checks, but I understand that that isn't great. I'll think a bit about it and most likely accept your pull request. The database modification from scripts are messy, so maybe having less of them is better right now.

aBlueShadow:
Especially for that scenario, I don't see much benefit for those db modifications considering the problems. It is literally just the same ship with the same db entry just copied into the exuari category with another name. I don't know if that feature is worth the hassle. And the custom typenames are very descriptive, so sourcing them out to the descriptions wouldn't be that bad IMO.

PR #2105, opened 2024-05-14
Fix Voice in scenario 48 and 51, fix philips hue integration

Massimo53:
I didn't mean to include philips hue in this to keep to it separate from scenario fixes, but I can't take it out.

### DIFF
--- a/src/spaceObjects/spaceship.cpp
+++ b/src/spaceObjects/spaceship.cpp
@@ -902,7 +902,7 @@ void SpaceShip::drawOnRadar(sp::RenderTarget& renderer, glm::vec2 position, floa
     // Otherwise, draw the ship-specific icon.
     if (my_spaceship && (getScannedStateFor(my_spaceship) == SS_NotScanned || getScannedStateFor(my_spaceship) == SS_FriendOrFoeIdentified))
     {
-        object_sprite = "radar/arrow.png";
+        object_sprite = "radar/ship.png";
     }
 
     glm::u8vec4 color;


aBlueShadow:
Pull requests compare changes between two branches across two repositories.
So if you want to keep changes seperate, you have to create different branches. Which does not necessary mean you have to do it now, it might be that daid is okay with it in this instance. Just something to be considered for the future.

daid:
Sadly, the hue changes look like a work and progress and not something ready for merging.

aBlueShadow:
That's unfortunate

you now have several options @Massimo53 :

a) make sure you have a local copy of the hue file, then undo the changes of that file in your online repository. This might look a bit irritating in the commit history, but will do the trick.
b) make sure to have local copies of all three changed files, delete your EE fork, then make a new one. As your fork just had 3 commits and no other branches yet, that would be a viable option. This will reset all your changes. Then make sure to make branches this time. In this branches, you can change the files and make Pull requests.
c) you can also undo your last commit, but I think that only works if you have git setup locally.

If you are new to git and github, I would recommend a) or b)

daid:
(actually, you could just make a branch from commit 7a4212b and submit that)

aBlueShadow:
Or that :-)
That would fall in the same category as 3) though, at least I could not find a way to do this in the web interface, it seems you only can branch off from another branch on the web.

Massimo53:
no problem, it was a quick and dirty because I didn't have time to set up my form properly. I'll resubmit once I have it set up.

PR #2113, opened 2024-05-23
fix: prevent segfault when template does not exist

Piglit:
setTemplate now produces an error, if the chosen template does not exist. But the game no longer crashes with a segfault.


daid:
I'm quite sure this moves segfaults to other spots, there are quite some places in the code that assume that the template property is set.

daid:
In that case, it's mostly fine. There is still a chance that it runs into a crash because other code doing something with the object before the update runs, but those are then extreme edge cases

PR #2120, opened 2024-06-01
Responsive custom function layout

aBlueShadow:
https://github.com/daid/EmptyEpsilon/assets/25465934/221492a7-1a24-49b6-a005-fa9434d90d4e

This let the custom buttons on science/ops have an extra sidebar, if the screen is wide enough. That way, those buttons/labels won't be as easy to miss as before, at least on 16:10 and wider.
I also adjusted size and positions of the Reputation and clock stats, so they wont overlap with the changes. (I checked both with all supported languages, the text still fits in)
I also made the science radar just a little bit smaller (it is about the size as the ops radar now), as the layout works better this way.


daid:
As the radar is the most important thing on the science screen, I'm not sure that I like it being smaller, even if it is just a tiny bit.

aBlueShadow:
Well, it would be the same as the operations radar always has been (due to the ship log) , and both have the same purpose.

PR #2124, opened 2024-06-20
Autodock update

tiittapauppi:
Once fighter out of docking range, auto dock disappears from odysseus relay view.


tiittapauppi:
Pul lrequest to incorrect fork

PR #2132, opened 2024-07-26
Change Nebula 3D Rendering to enable Distant Nebulas

ragerin:
**Summary**

This PR changes the 3D rendering of the nebulas, such that instead of disappearing by a factor of the distance, they instead turn almost opaque at a distance, such that they can visually conceal other objects behind them.

This need came from a group I play with, that spotted a "hidden" object very easily on the main viewer, thus turning a longer scenario into a direct course into the nebula.

image

**Changes Made**

- Shader file:
  - Created a new shader file (`nebula.shader`) specifically for the nebula graphic.
- Nebula graphics:
  - The graphics for the Nebulae was changed to have transparency, to allow the shader to do the pre-multiplied alpha calculations.
- Nebula SpaceObject:
  - Added blend functions: `GL_SRC_ALPHA` and `GL_ONE_MINUS_SRC_ALPHA`.
  - Made the `alpha` calculation linear to the distance. 
    - This may need to be improved, to prevent the quads from "clipping" through the player ship.
  - Clamped the `alpha` to prevent the white-out effect from high distance nebulae.
  - Removed the `* 0.8f` scaling, since it's not apparently needed.

**Notes**

- The colors may not be super pleasing, but could be improved by adjusting or adding nebula-graphics.
- It may be a good idea to implement a faded "cutoff", to ensure the nebulae at extreme distances are not rendered.
- Performance-wise I saw no noticeable effect, compared to the official release, but it's only measured with the builtin FPS and timing-gizmos.
- Adding a user-setting for controlling the maximum rendering-distances may be a good idea, if the performance is affected too much on some hardware.


ragerin:
Oh, and by the way, regarding the notes: I'm most likely implementing the suggested changes in my fork at some point anyway, but if this PR is interesting enough to the main project, I'll gladly prioritise it to be included in this PR before an eventual merge. 

And of course, fully respect the design choices, if this PR does not align. 😊👍

PR #2133, opened 2024-07-28
Add beam frequency/targeting to Single Pilot station

GinjaNinja32:
Code copied from Tactical station.


daid:
With any UI change, I prefer seeing a before/after screenshot (saves me from compiling the branch, being the lazy person that I am)

daid:
Thanks, yes, that looks good :)

PR #2138, opened 2024-08-11
Re-implement set next target in ECS

kzsh:
- Maybe I'm just new to C++, but it seems difficult to distinguish between public and private methods without consulting the header file.  Is there a convention around this?
- The current implementation also targets incoming missiles.  This could be a feature if that has any gameplay effect, or it could just make it hard to cycle to an actual enemy if there are a lot of missiles in the air.


src/screenComponents/targetsContainer.h
```diff
@@ -37,7 +37,9 @@ class TargetsContainer
     int waypoint_selection_index;
     glm::vec2 waypoint_selection_position{};
 
+    void setNext(glm::vec2 position, float max_range, std::vector<sp::ecs::Entity>& entities);
+    void sortByDistance(glm::vec2 position, std::vector<sp::ecs::Entity>& entities);
     bool isValidTarget(sp::ecs::Entity entity, ESelectionType selection_type);
 };
 
-#endif//TARGETS_CONTAINER_H
+#endif//TARGETS_CONTAINER_H
```
kzsh, line 45: I need to update my editor not to alter existing conventions

daid:
> The current implementation also targets incoming missiles. This could be a feature if that has any gameplay effect, or it could just make it hard to cycle to an actual enemy if there are a lot of missiles in the air.

That is unintended, wonder why that happens. None of the components I check for should be on a missile...

PR #2139, opened 2024-08-15
Update website screenshots

parracodea:
The screenshots on the site don't match the UI in the current version of the game - it looks like they were taken in 2016.


daid:
Thanks! Those where indeed quite old

aBlueShadow:
The new sreenshots are very nice, although it seems like they are truncated (which is especially noticeable on Main Screen and Science)

parracodea:
True, it's especially cropped on the Science console. Most of the original screenshots were a ~3:2 aspect ratio and I was trying to keep that, so I had the game running in Windowed Mode instead of Full Screen, which may have been part of the problem.

If I have some time later, I'll try and retake the screenshots. (I'm glad I saved the scenario :D )

daid:
Maybe reusing the screenshots from the steam page works?

PR #2140, opened 2024-08-17
ECS: re-implement mesh rendering in database entries

kzsh:
In the first few commits, this was close to a copy/paste from `MeshRenderSystem::render3D` to `GuiRotatingModelView::onDraw`.  If duplicating the rendering pipeline in these two places is sufficient, I'm happy to simplify this PR back down to that.

That said, the rest of the work is an attempt at re-using some of the code between `render3D` and `onDraw` without compromising the clarity of the render system to serve the one-off needs of `onDraw`.  I will admit that the likelihood of render code changing often is pretty low, so maybe this kind of re-use isn't that critical here.

Aside: Again, I'm pretty new to C++, so I'm open to thoughtful feedback on language features I'm misusing, or weaknesses I might be inadvertently introducing, or patterns that make more sense than what I've put in. 

I've opted to save the debug work for a separate PR to keep this one a bit smaller, but I did leave one piece from my exploration there in this series of commits https://github.com/daid/EmptyEpsilon/pull/2140/commits/2d0405c3a9cf62d1ec288e346359e9aa3e45aadf.  I think it'll be useful to have an entity in order to enumerate and render emitters.

### DIFF
--- a/src/mesh.cpp
+++ b/src/mesh.cpp
@@ -144,28 +144,14 @@ uint32_t Mesh::greatestDistanceFromCenter(std::vector<MeshVertex>& vertices)
         return 0;
     }
 
+    glm::vec3 sum{};
+    for(auto vertex : vertices) sum += glm::vec3{vertex.position[0], vertex.position[1], vertex.position[2]};
+    auto average = sum / float(vertices.size());
 
-    float sum_x = 0.f;
-    float sum_y = 0.f;
-    float sum_z = 0.f;
-    for(auto vertex : vertices)
-    {
-        auto [px, py, pz] = vertex.position;
-        sum_x += px;
-        sum_y += py;
-        sum_z += pz;
-    }
-
-    auto vertex_count = vertices.size();
-    auto average_x = sum_x / vertex_count;
-    auto average_y = sum_y / vertex_count;
-    auto average_z = sum_z / vertex_count;
     auto greatest_distance = 0.f;
-
     for(auto vertex : vertices)
     {
-        auto [px, py, pz] = vertex.position;
-        float distance = sqrt(pow(px - average_x, 2) + pow(py - average_y, 2) + pow(pz - average_z, 2));
+        float distance = glm::distance(average, glm::vec3{vertex.position[0], vertex.position[1], vertex.position[2]});
         if(distance > greatest_distance) {
             greatest_distance = distance;
         }
--- a/src/mesh.cpp
+++ b/src/mesh.cpp
@@ -56,13 +56,15 @@ Mesh::Mesh(std::vector<MeshVertex>&& unindexed_vertices)
         glBindBuffer(GL_ARRAY_BUFFER, vbo_ibo[0]);
         glBufferData(GL_ARRAY_BUFFER, sizeof(MeshVertex) * vertices.size(), vertices.data(), GL_STATIC_DRAW);
         glBindBuffer(GL_ARRAY_BUFFER, GL_NONE);
-        
+
         if (!indices.empty())
         {
             glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, vbo_ibo[1]);
             glBufferData(GL_ELEMENT_ARRAY_BUFFER, index_count * sizeof(uint16_t), indices.data(), GL_STATIC_DRAW);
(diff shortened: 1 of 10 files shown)


src/systems/rendering.cpp
```diff
@@ -11,6 +11,21 @@
 
 std::vector<RenderSystem::RenderHandler> RenderSystem::render_handlers;
 
+bool MeshRenderComponent::ensureLoaded()
```
kzsh, line 14: This would probably be less error prone if the fetching/caching logic were moved into `getters` for the fields in question.  It's annoying to have to make sure to call `ensureLoaded()` in the right place before using mesh fields.

src/systems/rendering.h
```diff
@@ -67,8 +67,16 @@ class RenderSystem
...
 };
+
 template<typename COMPONENT, bool TRANSPARENT> Render3DInterface<COMPONENT, TRANSPARENT>::Render3DInterface() { RenderSystem::add3DHandler(this); }
 
+// FIX: This is obviously a very inelegant way to share behavior
+glm::mat4 calculateModelMatrix(glm::vec2 position, float rotation, MeshRenderComponent& mrc);
+glm::mat4 calculateModelMatrix(glm::vec2 position, float rotation, MeshRenderComponent& mrc, float scale);
+ShaderRegistry::ScopedShader lookUpShader(MeshRenderComponent& mrc);
+void activateAndBindMeshTextures(MeshRenderComponent& mrc);
+void drawMesh(MeshRenderComponent& mrc, ShaderRegistry::ScopedShader& shader);
```
kzsh, line 78: I think these being in a separate file.  Some kind of render utils. The hope is that by extracting these bits of behavior, we can re-use it in `rotatingModelView.cpp:draw` and `render.cpp:MeshRnderComponent::render3D` without forcing `render3D` to adapt to `draw`, but also to allow both to benefit from improvements / changes to render logic without having to separately track down and update them.

If it's simpler or more cost efficient just to maintain two separate implementations, I don't mind ripping this out.

src/systems/rendering.cpp
```diff
@@ -105,18 +107,51 @@ void MeshRenderSystem::render3D(sp::ecs::Entity e, sp::Transform& transform, Mes
...
+            mrc);
+
+    auto shader = lookUpShader(mrc);
+    glUniformMatrix4fv(shader.get().uniform(ShaderRegistry::Uniforms::Model), 1, GL_FALSE, glm::value_ptr(model_matrix));
+
+    auto modeldata_matrix = glm::rotate(model_matrix, glm::radians(180.f), {0.f, 0.f, 1.f});
+    modeldata_matrix = glm::scale(modeldata_matrix, glm::vec3{mrc.scale});
+
+    // Lights setup.
+    ShaderRegistry::setupLights(shader.get(), modeldata_matrix);
```
kzsh, line 145: I'm not entirely sure I understand why we invert the model for lighting.  I tried rendering with and without the inversion, and while there were differences, they seemed subtle.
Is it to keep things synchronized with other `render3D` rendering?

daid: I guess so. This code is old and has seen many revisions from multiple people :)

src/mesh.cpp
```diff
@@ -136,6 +138,41 @@ glm::vec3 Mesh::randomPoint()
...
 }
 
+uint32_t Mesh::greatestDistanceFromCenter(std::vector<MeshVertex>& vertices)
+{
+    if (vertices.empty()) {
+        return 0;
+    }
+
+
+    float sum_x = 0.f;
```
daid, line 148: maybe it would be possible to use `glm::vec3` instead? Not sure if you can initialize a `glm::vec3` from a `float[3]`.

But then it would look something like:
```
glm::vec3 sum{};
for(auto vertex : vertices) sum += glm::vec3(vertex.position);
auto average = sum / float(vertices.size());

for(auto vertex : vertices) {
    float distance = glm::length(glm::vec3(vertex.position) - average);
    ...
}
```

kzsh: I like that idea! That's where I started before changing it to try to match the style of the code above it, but I think it would be cleaner with `glm` structures/functions.

PR #2146, opened 2024-08-23
Use Environment::runWithStringResults for a better Lua console REPL

GinjaNinja32:
See also daid/SeriousProton#246.


GinjaNinja32:
https://github.com/daid/SeriousProton/pull/246#issuecomment-2427871331

PR #2153, opened 2024-08-31
Add radar signature to artifacts and suggest using probes in scenario "Birth of the Atlantis"

parracodea:
I played this scenario with some new players, and it took over an hour for them to complete Phase 2 (of 5 phases).

They felt that there was too much time "blindly wandering around" the nebula looking for the artifact. When they did come near one of the artifacts, it appeared on the edge of the radar, so they missed it and then got distracted with one of the Ghosts. I swapped to the GM screen to tell them to go back since we had already been searching for a while at that point. This was kind of discouraging to them (flying through a nebula isn't what makes EmptyEpsilon interesting) and they didn't get to do much combat since we had to stop playing right after that.

I've removed some of the nebula from 8 to 6 so there's less players have to search through. I also spaced them out to encourage players to jump between nebula (rather than flying slowly with impulse), and so that its easier to keep track of which nebula they've searched (it used to appear like one massive nebula on the player's screen).

## Before:
Before

## After:
After

### DIFF
--- a/scripts/scenario_08_atlantis.lua
+++ b/scripts/scenario_08_atlantis.lua
@@ -147,7 +147,7 @@ function init()
 
     nebula = table.remove(b20_nebula_list, math.random(#b20_nebula_list))
     x, y = nebula:getPosition()
-    b20_artifact = Artifact():setPosition(x + random(-1000, 1000), y + random(-1000, 1000)):setRadarTraceColor(250, 79, 255)
+    b20_artifact = Artifact():setPosition(x + random(-1000, 1000), y + random(-1000, 1000)):setRadarTraceColor(250, 79, 255):setRadarSignatureInfo(0.25, 0, 0)
     b20_artifact:setScanningParameters(3, 1)
     b20_artifact.nebula = nebula
     b20_artifact.beta_radiation = irandom(1, 10)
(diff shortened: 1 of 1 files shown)


aBlueShadow:
>  (flying through a nebula isn't what makes EmptyEpsilon interesting)
but searching a nebula with probes can be interesting.

You probably did not have a full crew, correct? I think this scenario assumes to have a dedicated relay station, as with probes, there is no need to aimlessly wandering around, and the artifact should be able to be found pretty quickly.

I am not sure if making wandering though nebulas less tedious is the way to go, as you are not supposed to wander through nebulas in the first place (IMO).

One way to account for limited crew setup would be to give additional info when relay is not occupied, like mentioning the specific sector name it is in. Or, maybe even better: give the artifact a strong signature, so it makes a "dent" in the outer lines of the long range radar, indicating the direction.

parracodea:
You're right, we were playing with Tactical/Engineering+/Operations, and using probes would have been a better solution - I forgot about them.

>  you are not supposed to wander through nebulas in the first place (IMO)

I think the original scenario creator expected that too, because after reporting the readings from the artifact, JC-88 tells Atlantis-1 to get closer for a better reading. Our crew made a comment about that since we were parked next to the artifact for the original scan.

I like your suggestions. I update the comms messages to suggest using probes and looking at the radar signature. The artifact now has a gravitational signature, which is also foreshadowing for JC-88 reports a huge gravitational spike and it turns into a wormhole.

Screenshot of the artifacts with a gravitational radar signature (most obvious in-game with the green line):

image

daid:
I like these changes, makes it more streamlined and gives more ways to progress.

You might be surprised to learn that I wrote this scenario before probes even existed :) so the original intent wasn't with probes. But using probes does speed it up quite a bit.

The "move closed" is actually "move to a certain distance", if you are close, it will tell you to take some distance. As I want you on the edge of the wormhole when it spawns, so that you have time to see it pull you in.

PR #2156, opened 2024-09-03
Server scenario improvements

GinjaNinja32:
- Don't crash if the server fails to start with `server_scenario` set, just exit nonzero instead.
- Don't ignore scenario options specified just before an empty comment line, to allow breaking up blocks of settings for ease of reading.
- Add `scenario_settings` option, to set scenario settings when using `server_scenario` or `headless`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -420,7 +420,7 @@ int main(int argc, char** argv)
             return 1;
 
         // Load the scenario and open the ship selection screen.
-        gameGlobalInfo->startScenario(PreferencesManager::get("server_scenario"));
+        gameGlobalInfo->startScenario(PreferencesManager::get("server_scenario"), loadScenarioSettingsFromPrefs());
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -516,3 +519,25 @@ void returnToOptionMenu()
...
+{
+    string preferenceValue = PreferencesManager::get("scenario_settings");
+
+    std::unordered_map<string, string> settings = {};
+    if (preferenceValue == "")
+        return settings;
+
+    for(string setting : preferenceValue.split(";"))
+    {
+        std::vector<string> key_value = setting.split("=", 1);
```
daid, line 533: You could do: 
```
auto [key, value] = setting.partition("=");
if (!value.empty())
    settings[key.strip()] = value.strip();
```

daid:
People have asked for this for a few times, so I'm happy to see it!

PR #2158, opened 2024-09-05
fix rotation behavior in tactical and single pilot (ECS version)

aBlueShadow:
ECS version of #2157


daid:
As I regularly merge master to ECS, I did the this there.

PR #2163, opened 2024-09-08
Liberation Day Scenario

Xansta:
**Depends on pull request #2162** 

Advanced scenario for one or more player ships. Terrain and other aspects randomized between scenario runs, so the scenario can be replayed multiple times.
- Time limit:
  - Default: 30 minutes
  - Shortest: 15 minutes
  - Longest: 60 minutes
  - The longer the time limit, the easier the scenario


aBlueShadow:
The script requires player_ship_upgrade_downgrade_path_scenario_utility.lua, which is not present

PR #2165, opened 2024-09-14
Convert variation in carriers and turrets to new format

aBlueShadow:
Converted the old scenario variation to a difficulty setting in the new format, so the update locale script can generate translation entries for it.


aBlueShadow:
i just realised, that the new settings cannot get recognised, as I would also need to change getScenarioVariation() to getScenarioSetting("Difficulty"). And then I realised that the old setting probably never did anything in the first place, as while there is a getVariation() function that would take care of it, it never gets called.
The new commit fixes both issues.

PR #2168, opened 2024-09-16
Scenario Surf's Up!

Xansta:
**Pull Request depends on pull request #2167** 

An open ended scenario for one or more player ships. Similar in concept to the Waves scenario in that wave after wave of enemies come trying to destroy the friendly bases. No victory condition. Difficulty increases with each wave.

#### Configurations
- Enemy ship difficulty: Easy, Normal, Hard, Extreme, Quixotic
- Prototype player ship available (fighter, cruiser, missile cruiser)
- Enemy ship speed, aka Pace (normal, 10% faster, 20% faster, 30% faster, ... 50% faster)
- Start at specific wave (1 [default], 2, 3, ... 9)
#### Game Play
- Different wave types (determined at random)
  - Normal
  - Base: Enemy base must be hunted down. The base will continue to spawn enemies until it is destroyed
  - Defense: Same as Base type, but the base has defense platforms around it
  - Formation: Enemy ships attack in formation
- Any wave type could include an asteroid storm (random determination)
- The services available are split between the friendly stations, motivating the players to protect all stations.
  - Services: charge ship batteries, repair ship hull, replenish scan probes, provide five types of ordnance
- Missions to implement or fix services that are not available at a station are possible
  - Goods to complete the mission may be obtained from independent stations or freighters
- Enemies may drop warp jammers
  - Warp jammers may also damage player ship systems

### DIFF
--- a/scripts/scenario_29_surf.lua
+++ b/scripts/scenario_29_surf.lua
@@ -1,7 +1,5 @@
 -- Name: Surf's Up!
--- Description: Wave after wave of enemy ships attack, each wave harder than the previous 
---- wave. Single or multiple player ships may participate. Scenario is over when the
---- friendly bases are destroyed. Loosely based on the Waves scenario.
+-- Description: Wave after wave of enemy ships attack, each wave harder than the previous wave. Single or multiple player ships may participate. Scenario is over when the friendly bases are destroyed. Loosely based on the Waves scenario. The differences between Waves and Surf's Up: Initial configuration options (faster moving enemy ships, set the start wave), random wave type (normal, hunt enemy base, formation flying, asteroid storm), services split between friendly stations, side missions to enhance stations, enemies may drop deadly warp jammers. 
 ---
 --- No victory condition. How many waves can you complete?
 ---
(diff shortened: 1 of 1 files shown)


scripts/scenario_29_surf.lua
```diff
@@ -0,0 +1,1762 @@
+-- Name: Surf's Up!
+-- Description: Wave after wave of enemy ships attack, each wave harder than the previous 
+--- wave. Single or multiple player ships may participate. Scenario is over when the
+--- friendly bases are destroyed. Loosely based on the Waves scenario.
```
daid, line 4: I think it would be good to mention that this is a more complicated scenario then the "Waves" scenario, where waves is quite basic, this provides different challenges.

We should update descriptions so they are helpful for people to know to pick a certain scenario over another for a particular crew.

Xansta: I updated the description. I think it's available as part of this pull request

aBlueShadow: About that: would it be possible to integrate the old waves scenario into this? Like, in a similar fashion Basic battle and Quick Basic merged a while ago? Haven't played it yet, but from the looks of it, the default settings of this ones seem to be similar to the default setting of waves. The playable area also seems to have a similar scope. Or is there something that makes both scenarios incompatible to each other?

Xansta: It's quite possible. I think I tried making version of Waves with more options a couple of years ago that did not get accepted. The changes in Surf's Up are different than what I did before.

PR #2175, opened 2024-09-22
Add relay map with no interactions as captains map

Matombo:
Gives the captains laptop something to display when he is not just providing the main screen. Arguably makes the captains role easyer as he needs to communicate less with relay.

This patch has a desgin bug that I'm not really sure how to fix: the new entry in alternative options sections in the ship station select screen does not fit the current layout. (and as such only the upper pixel row of the button is clickable). Any idea on how to make this look right again and make the button clickable in whole?


aBlueShadow:
You would need to rearrange an resize the Gui elements in shipSelectionScreen.cpp (CrewPositionSelection)
That said, I am not sure if there is much of a benefit over using the strategic map and just don't touching the buttons.

Matombo:
The benefit would be a cleaner look and another option for LARP settings.

Also "just not touching the buttons" might not work for every player group.

daid:
I personally like it if the captain has zero direct interaction with the game.

The mainscreen can currently be switched to short and long range radar, maybe it makes sense to add the relay map as option there as well?

PR #2180, opened 2024-10-01
add MetaInfo file

jstsmthrgk:
* copied summary from `.desktop` file
* copied description from website (left headings out)
* copied categories from `.desktop` file
* took the screenshots from the website, copied captions from website
* copied description of last release from github

stuff still missing (see https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines):
- [x] Metadata License, daid must agree that the metadata in this file is under a looser license (is needed for all the aggregation that is done by flathub and such) the guidelines ask for FSFAP or CC0.
- [X] Content rating, is ships shooting at each other and destroying them violence and in which bucket of OARS does it fall? Also probably "Uncontrolled chat functionality between users" and "Uncontrolled audio or video chat between users" (used [Pioneer](https://github.com/flathub/net.pioneerspacesim.Pioneer/blob/master/pioneer.metainfo.xml) for reference)


io.github.daid.EmptyEpsilon.metainfo.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2024 daid -->
+<component type="desktop-application">
+  <id>io.github.daid.EmptyEpsilon</id>
+  
+  <name>EmptyEpsilon</name>
+  <summary>Spaceship bridge simulator game</summary>
+  
+  <!-- of course I just copied the stuff from the website, daid needs to agree first <metadata_license>FSFAP</metadata_license>-->
+  <project_license>GPL-2.0-only</project_license>
```
daid, line 10: This isn't true. Some of the assets are under a closed license.

io.github.daid.EmptyEpsilon.metainfo.xml
```diff
@@ -0,0 +1,74 @@
...
+      <caption>View of the science officer's screen, which has long-range scanners and a large database of information.</caption>
+    </screenshot>
+    <screenshot>
+      <image>https://daid.github.io/EmptyEpsilon/screenshots/relay.png</image>
+      <caption>View of the relay officer's screen, which can see and hail allies, share their short-range radar, and launch remote probes.</caption>
+    </screenshot>
+  </screenshots>
+  
+  <releases>
+    <release version="2024.08.09" date="2024-08-09">
```
daid, line 67: I don't like having hardcoded release numbers and links in the repo.

daid (changes requested):
In this pull request, I am not sure why this file is needed (I might just be lacking background info)

jstsmthrgk:
It is required because else submission to flathub is not possible.

https://docs.flathub.org/docs/for-app-authors/requirements
> All submissions must provide a Metainfo file that passes validation.

The file could of course also be hosted somewhere else or maybe even be generated, but I guess it would be easiest to just keep it in here.

io.github.daid.EmptyEpsilon.metainfo.xml
jstsmthrgk: Hmmm, it seems that only version name and date would be enough, maybe that part should be templated somehow. (similar to how [PrismLauncher](https://github.com/PrismLauncher/PrismLauncher/blob/develop/program_info/org.prismlauncher.PrismLauncher.metainfo.xml.in) does it)

io.github.daid.EmptyEpsilon.metainfo.xml
jstsmthrgk: Ok, now having looked at all files with license in their name, I suppose the correct one would be:
`GPL-2.0-only AND LicenseRef-TurboSquid-Royalty-Free-License AND OFL-1.1 AND Apache-2.0`

io.github.daid.EmptyEpsilon.metainfo.xml
jstsmthrgk: okay, so I basically copied the approach from PrismLauncher, it is now tested and working, I'll mark this as resolved

jstsmthrgk:
I have also ran the checks, it does complain about the captions for the screenshots being too long, but that is just a warning. (Also it errors on the missing metadata license, but when I uncomment that it passes.)

daid:
> It is required because else submission to flathub is not possible.
> 
> https://docs.flathub.org/docs/for-app-authors/requirements
> 
> > All submissions must provide a Metainfo file that passes validation.
> 
> The file could of course also be hosted somewhere else or maybe even be generated, but I guess it would be easiest to just keep it in here.

I think, it would be better then to have a template file and fill in the details during cmake configure. https://cmake.org/cmake/help/latest/command/configure_file.html can be used for this to fill in things like version numbers. Just like I do for the android manifest: https://github.com/daid/EmptyEpsilon/blob/master/cmake/android/AndroidManifest.xml.in

jstsmthrgk:
> I think, it would be better then to have a template file and fill in the details during cmake configure. https://cmake.org/cmake/help/latest/command/configure_file.html can be used for this to fill in things like version numbers. Just like I do for the android manifest: https://github.com/daid/EmptyEpsilon/blob/master/cmake/android/AndroidManifest.xml.in

way ahead of you, was already done before your comment (although it seems I used a different syntax for the variables)

as the metadata license is now also in there, it should be ready to merge (I think)

PR #2205, opened 2024-10-16
Add Flathub Badge on Website

jstsmthrgk:
I think it would be nice to have it there.

image


daid:
Not objecting to the concept, but I'm not liking the implementation:

- It's too prominent, acting as if it is the primary source, while it is a secondary source. Reality is that the largest user-base is on windows, and they will now be confusingly redirected to flathub.
- Styling of the html elements is done with html/css, not by making an image with text on it. See the `stations` tab for buttons that are styled with icons.

jstsmthrgk:
>  * It's too prominent, acting as if it is the primary source, while it is a secondary source. Reality is that the largest user-base is on windows, and they will now be confusingly redirected to flathub.

The link is inherently non-versioned, so I guess the location should be kept? But yeah, it is a bit too prominent, maybe just a text link styled like the others?

>  * Styling of the html elements is done with html/css, not by making an image with text on it. See the `stations` tab for buttons that are styled with icons.

Actually it is just coincidence that it looks like the other buttons, I've downloaded the one provided by flathub and it has the border already in it.

parracodea:
Would something like this be better? Note, the first table header (`OS`/`File`/`Version`) has been removed, since Flathub doesn't really fit under the columns.

ee-flathub-download-link

Issue #2210, opened 2024-10-24
compilation issue with ninja package using the wiki instructions. 


gwaland:
While attempting to build a Debian package using the ninja package command from the wiki I've run into this error: 

`~/buildtest/EmptyEpsilon/_build$ ninja package
[0/1] Run CPack packaging tool...
CPack: Create package using DEB
CPack: Install projects
CPack: - Install project: EmptyEpsilon []
CMake Error at /home/gwaland/buildtest/EmptyEpsilon/_build/cmake_install.cmake:82 (file):
  file INSTALL cannot find
  "/home/gwaland/buildtest/EmptyEpsilon/io.github.daid.EmptyEpsilon.metainfo.xml":
  No such file or directory.

CPack Error: Error when generating package: EmptyEpsilon
FAILED: CMakeFiles/package.util 
cd /home/gwaland/buildtest/EmptyEpsilon/_build && /usr/bin/cpack --config ./CPackConfig.cmake
ninja: build stopped: subcommand failed.`

I can work around this by making a symlink to the xml file that's created in the _build directory to the root of the codebase: 
`ln -sf io.github.daid.EmptyEpsilon.metainfo.xml ../io.github.daid.EmptyEpsilon.metainfo.xml`

I'm thinking there may be something in the cmake.list changes made for the flatpack packaging that is causing this.

gwaland:
I'm pretty sure this line is the culprit. 
https://github.com/daid/EmptyEpsilon/blob/b9223f90171a8ef150b94150222d8ed0747e1165/CMakeLists.txt#L750

I'm testing modifying it to this: 
configure_file(io.github.daid.EmptyEpsilon.metainfo.xml.in ${PROJECT_SOURCE_DIR}/io.github.daid.EmptyEpsilon.metainfo.xml)

I think this will allow it to build the way I'm expecting and also allow for people to build in the root path which is what the configuration apppears to expect.

PR #2211, opened 2024-10-24
Fixing ninja package build error using wiki instructions. 

gwaland:
I believe this will resolve the issue described in #2210 
I tested this with the wiki instructions and by doing cmake in the root path of the repository and both functioned as I expected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -747,10 +747,10 @@ set(CPACK_INSTALL_CMAKE_PROJECTS
 
 include(CPack)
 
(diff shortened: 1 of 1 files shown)


jstsmthrgk:
I guess it would be better to instead change the `install` rule to:

```cmake
install(FILES ${CMAKE_CURRENT_BINARY_DIR}/io.github.daid.EmptyEpsilon.metainfo.xml DESTINATION "share/metainfo")
```

`configure_file` takes from `CMAKE_CURRENT_SOURCE_DIR` and puts into `CMAKE_CURRENT_BINARY_DIR` by default, I think it is better to keep it that way.

daid:
Yeah, it's not normal to change any files in the "source dir" during configure/build. So the install command should get it from the binary dir instead of configuring it to the source dir.

PR #2227, opened 2024-10-29
Simplify signatures

aBlueShadow:
This simplifies the current signature system by removing one abstraction layer. 

Instead of influencing multiple colored bands, each signal now directly correspond with one colored band:
electric: red
biological: green
gravitational: blue
signatures

I don't think that makes things too easy, it still don't give too much info away, and there is still plenty to learn about interpreting the signals - especially as objects often have multiple signals.

There is probably still stuff that can be tweaked - For example, that change now shows more clearly that electric is hardly visible on some objects. The signal of that artifact at bearing 81 in the screenshot is at full 1.0 strength. It is slightly more visible in motion, but it is still subtle.

But this was also the case with the old system, and I think we can use this PR as a foundation to built upon, and maybe it will also encourage scenario writers to use signatures more often.


daid:
Thanks, I wanted to look into this but hadn't had a moment yet.

PR #2230, opened 2024-10-30
Website: Add Flathub to the downloads page

parracodea:
An alternative to https://github.com/daid/EmptyEpsilon/pull/2205. Would something like this be better? 

Note, the first table header (`OS`/`File`/`Version`) has been removed, since Flathub doesn't really fit under the columns.

A link to the Steam page should probably be added here at some point, so any design decisions chosen now should also be applicable to Steam.

ee-flathub-download-link


daid:
yeah, I like this a lot more. Guess I could add steam in the way way as soon as that is released...

PR #2236, opened 2024-11-04
Add translation markers for delta patrol duty scenario

aBlueShadow:
The gm buttons here are supposed to be removed at certain events, like clicking said button. As there were translation markers for creating the buttons, but not for removal, ramoving does not work for non-english languages. This PR adds the missing marker.
**Note for translators**: This doesn't add any additional translation strings, so if you have this scenario already translated, there is no additional work to do.


muerteFR:
It is prudent to systematically notify @Xansta when we make small modifications to these scenarios on Daid's repo, so that he takes them into account when he works on these scenarios, otherwise these small modifications will be overwritten by those that 'he does on his side. :-)

aBlueShadow:
I would have assumed that Xansta will regularly merge daids repo from time to time, as keeping an eye on the master repo is important for any kind of change, but a notification will not hurt, of course.

PR #2247, opened 2024-11-23
GitHub Pages: Update screenshots

parracodea:
Updated screenshots to address the issue of some images being cropped at the top and bottom, as mentioned in https://github.com/daid/EmptyEpsilon/pull/2139#issuecomment-2291146816

For future reference, the "screenshot scenario" used can be found [here](https://github.com/parracodea/EmptyEpsilon-Scenarios/blob/main/scenario_907_screenshot_scenario.lua).


daid:
I see how that is better, but I kind miss the scanning dialog on the scanning screenshot.

PR #2248, opened 2024-11-23
GitHub Pages: Center Debian logo

parracodea:
(remove text from SVG)

At a glance, the Debian logo looks off center:
image

However, that's due to the `Debian` text being black over a black background.

After removing the text:
image


parracodea:
Updated to use the Debian logo from https://www.debian.org/logos/ (instead of the hand-edited version that I originally posted)

PR #2252, opened 2024-11-28
Scenario Surf's Up! Update

Xansta:
Minor changes:
- Enable Current Orders
  - Set up the function to provide those
  - Enable the current orders via the comms utility
  - Set primary and secondary orders
- Add translation contexts
- Tweak wave advancement curve
- Comment out diagnostic lines


Xansta:
Note this depends on #2253

PR #2273, opened 2024-12-30
Proofreading 60% Scavenger.fr.po

muerteFR:


muerteFR:
After many hours of testing, I stopped proofreading of this scenario at 60%.

PR #2293, opened 2025-02-05
Add de translation for Planet Devourer

aBlueShadow:
While this scenario currently doesn't work in master (probably due to ECS stuff), I figured I should probably still finally upload this translation.


daid:
yeah, there are a few `.typeName ==` in the code which doesn't work with ECS so that still needs some fixing.

PR #2298, opened 2025-02-22
Relay Button to center on ship

csibbitt:


src/screens/crew6/relayScreen.cpp
```diff
@@ -162,6 +162,13 @@ RelayScreen::RelayScreen(GuiContainer* owner, bool allow_comms)
     });
     launch_probe_button->setSize(GuiElement::GuiSizeMax, 50)->setVisible(my_spaceship.hasComponent<ScanProbeLauncher>());
 
+    // Center on ship
+    center_button = new GuiButton(option_buttons, "CENTER_ON_SHIP", tr("Center On Ship"), [this](){
+        auto transform = my_spaceship.getComponent<sp::Transform>();
+        radar->setViewPosition(transform->getPosition());
```
daid, line 168: Ships that are internally docked do not have a transform. So you need to check if getComponent might return null before using it

csibbitt: Thanks, I'll experiment with this!

csibbitt: I've added null checks for everything and tested the internally docked case.

src/screens/crew6/relayScreen.cpp
```diff
@@ -162,6 +163,17 @@ RelayScreen::RelayScreen(GuiContainer* owner, bool allow_comms)
     });
     launch_probe_button->setSize(GuiElement::GuiSizeMax, 50)->setVisible(my_spaceship.hasComponent<ScanProbeLauncher>());
 
+    // Center on ship
+    center_button = new GuiButton(option_buttons, "CENTER_ON_SHIP", tr("Center On Ship"), [this](){
+        if(!my_spaceship) return;
+        auto transform = my_spaceship.getComponent<sp::Transform>();
+        if(!transform) // If my ship has no position, it must be docked inside another ship
+            transform = my_spaceship.getComponent<DockingPort>()->target.getComponent<sp::Transform>();
```
daid, line 171: in theory, a ship could have no transform and no docking port. So docking port also needs to be checked for nullptr.
```
if (!transform) {
  if (auto dp = my_spaceship.getComponent<DockingPort>()) {
    transform = dp->target.getComponent<sp::Transform>();
  }
}
```

daid:
For UI changes/additions, I prefer a screenshot as well, so I know how it looks after the change.

csibbitt:
> For UI changes/additions, I prefer a screenshot as well, so I know how it looks after the change.

image

PR #2299, opened 2025-02-28
Fix ShipLog replication

csibbitt:
* Entire ship log was replicating on every tick
* Using 17.5kb/s after filled to 100 entries
* Now uses 0.2kb/s (I'm sending bursts of messages at 10/s)
* Tested that back-replication still works to a new client


csibbitt:
## Before
Capture7

## After
Capture8

csibbitt:
Tested this with a server and two clients. Both clients got full back replication and received updates as expected.

PR #2303, opened 2025-03-17
[Payload] A series of ECS-related fixes for Payload scenario

csibbitt:
* Replace `.typeName` with component checks
* Delay path clearing until physics starts
* Suppress errors when ships or stations destroyed at end of game
* Fix NPC orders problems stemming from new default roaming

### DIFF
--- a/scripts/scenario_31_payload.lua
+++ b/scripts/scenario_31_payload.lua
@@ -246,13 +246,13 @@ function ClearHazardsNear(centerX, centerY, radius, maxRoids, maxMines)
 
   -- Destroy any hazards beyond the limits
   for _, potentialHazard in ipairs(objects) do
-    if potentialHazard.typeName == "Asteroid" then
+    if potentialHazard.components.explode_on_touch ~= nil then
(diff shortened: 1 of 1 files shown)


scripts/scenario_31_payload.lua
```diff
@@ -658,7 +654,7 @@ function SpawnWave(faction)
       if i == queenIndex and WaveSize > 1 then
         enemyTemplate = "Phobos M3"
       end
-      local enemy = CpuShip():setTemplate(enemyTemplate):setFaction("Kraylor"):setPosition(enemyX, enemyY)
+      local enemy = CpuShip():setTemplate(enemyTemplate):setFaction("Kraylor"):setPosition(enemyX, enemyY):orderIdle()
```
daid, line 657: That idle isn't the default should be a bug.

csibbitt: Noted. I'll see if I can implement a fix on the engine side, but this may also fix another edge case for me so might as well stay.

PR #2306, opened 2025-03-17
More distinguishable graphic for mines

Piglit:
Use the weapon tube icon for mines on the radar.
This should make mines more recognizable for new players.


daid:
See previous discussions on this.

PR #2311, opened 2025-03-17
options: add multimonior toggle to options menu

Piglit:
When the setting is changes, the program needs a restart. This is shown in the lable of the toggle button.


daid:
There are still problems with this mode, so I prefer to keep it hidden.

PR #2314, opened 2025-03-18
Borderline Fever scenario for defenders

Xansta:
Bridge between current and ECS
Check for ECS to check object type appropriately

### DIFF
--- a/scripts/border_defend_station.lua
+++ b/scripts/border_defend_station.lua
@@ -13,13 +13,17 @@
 --	}
 require("utils.lua")
 function init()
+	ECS = false
+	if createEntity then
+		ECS = true
+	end
 	check_interval = random(4,6)
 	check_timer = check_interval
 	inactivity_count = 0
 	inactivity_max = 300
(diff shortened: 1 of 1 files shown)


scripts/border_defend_station.lua
```diff
@@ -33,6 +37,34 @@ function init()
 	--name, faction _id and template set by calling script
 	my_ship = CpuShip():setCallSign(string.format("%s %s",station_name,name)):setCommsScript(""):setCommsFunction(commsDefendShip):setFactionId(faction_id):setPosition(position_x, position_y):setTemplate(template):setScanned(true):orderDefendTarget(my_station)
 end
+function isObjectType(obj,typ)
```
daid, line 40: Wouldn't it be better to use the new generic function you introduced? I merged that one.

Xansta: Yes it would. At the time that I made this pull request, I had not made the generic utility yet. I made a bunch of other Lua file changes with their own internal function before the light bulb went off about making a generic routine for other Lua code to use. I have not gone back to those other Lua document pull requests to use the utility yet.

PR #2331, opened 2025-03-20
Lua utils for scenarios

Xansta:
Add helper function to test for object type in either an ECS world or a pre-ECS world.


Xansta:
Added information about this pull request to the wiki on ECS

PR #2342, opened 2025-03-23
orbiting: Exception for time=0 and fixing description

aBlueShadow:
This adds a special case that stops orbital movement when time is set to 0, preventing divide by zero issues. I think it makes sense to hardcode this in C++ rather than changing it in lua.
I also updated the description of the setOrbit() function in planet.lua

### DIFF
--- a/scripts/api/entity/planet.lua
+++ b/scripts/api/entity/planet.lua
@@ -131,8 +131,11 @@ function Entity:setAxialRotationTime(rotation_time)
     end
     return self
 end
---- Sets a SpaceObject around which this Planet orbits, as well as its orbital period in orbital degrees per tick.
---- Example: moon:setOrbit(planet, 20)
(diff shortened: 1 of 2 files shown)


scripts/api/entity/planet.lua
```diff
@@ -131,8 +131,11 @@ function Entity:setAxialRotationTime(rotation_time)
     end
     return self
 end
---- Sets a SpaceObject around which this Planet orbits, as well as its orbital period in orbital degrees per tick.
---- Example: moon:setOrbit(planet, 20)
+--- Sets a SpaceObject around which this SpaceObject orbits, as well as its orbital period in seconds. Setting time to 0 will stop movement, while still being locked in orbit.
+--- An orbit can be cancelled by setting the SpaceObject as its own target.
```
daid, line 135: I think it's better to have `obj:setOrbit(nil)` cancel the orbit. Which can be done by deleting the orbit component:
`obj.components.orbit = nil`

aBlueShadow: Oh I thought I tried that, obviously not. Done

PR #2344, opened 2025-03-23
Random Transport Scenario Utility Updates

Xansta:
Make scenario utility work with ECS and pre-ECS

Dependent on #2331

### DIFF
--- a/scripts/util_random_transports.lua
+++ b/scripts/util_random_transports.lua
@@ -1,7 +1,7 @@
 --- Script for random transports.
 --
 -- @script util_random_transports
-
+require("utils.lua")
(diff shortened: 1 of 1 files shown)


aBlueShadow:
I would actually prefer the current check rather than the one for SpaceStation in the helper function. I makes the scenario more flexible (and In general, I think it makes more sense to check for a missing impulse drive instead of specific physics types).
Also, IMO there is no real need to make this work with older releases, as it is very unlikely that someone will copy this file into a an older pre-ECS installation (for scenarios this is of course different).

Xansta:
That works for me

PR #2357, opened 2025-05-01
Add various preferences to improve autoconnect server selection

GinjaNinja32:
`server_name`, `server_password`, `server_internet` are for `server_scenario=...` what the corresponding `headless_*` are for `headless=...`

`autoconnect_servername` prevents the autoconnect screen from connecting to any server whose name is not the preference value.

`autoconnect_password` allows specifying a password for the autoconnected server. Likely mostly (only?) useful in combination with `autoconnect_servername`.

### DIFF
--- a/src/menus/autoConnectScreen.cpp
+++ b/src/menus/autoConnectScreen.cpp
@@ -72,6 +72,7 @@ void AutoConnectScreen::update(float delta)
         if (autoconnect_address != "") {
             status_label->setText("Using autoconnect server " + autoconnect_address);
             connect_to_address = autoconnect_address;
+            tried_password = false;
             new GameClient(VERSION_NUMBER, autoconnect_address);
             scanner->destroy();
         } else {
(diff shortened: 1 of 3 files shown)


src/menus/autoConnectScreen.cpp
```diff
@@ -92,7 +99,17 @@ void AutoConnectScreen::update(float delta)
             else
                 status_label->setText("Connecting...");
             break;
-        case GameClient::WaitingForPassword: //For now, just disconnect when we found a password protected server.
+        case GameClient::WaitingForPassword:
+            if (!tried_password) {
+                auto password = PreferencesManager::get("autoconnect_password");
+                if (password != "") {
+                    game_client->sendPassword(password.upper());
+                    tried_password = true;
```
daid, line 107: this is never set to false after this, is that intentional?

GinjaNinja32: nope, just me mixing up scopes.

PR #2359, opened 2025-05-02
Allow tweaking warp/jump drives

eXeC64:
The GM menu doesn't yet allow tweaking of the warp/jump drive components. This PR adds that functionality, and also tidies up the existing tweak mapping a little bit by factoring out the ship system macros.


daid:
Yeah, the ECS tweak still needs a lot of extra work done to be as feature complete as the legacy version. Thanks!

PR #2371, opened 2025-05-21
fix internal docking for util_random_transports.lua

aBlueShadow:
If a dockable object has internal ports for the transports, they will never undock after  docking.
This fixes the issue by recreating a transform component before ordering to leave the port.

_There is still the issue that large transports never dock in the first place, but that is a general issue that can't be solved in this script - other than excluding size-5- freighters from the list_


daid:
Don't think this is the right solution, if transform needs to be created the undocking code should do that?

aBlueShadow:
The underlying issue seems to be solved now

PR #2387, opened 2025-06-07
Add handling of headless_port, server_port, autoconnect_port

oznogon:
When running a server via `server_scenario` or `headless`, there doesn't appear to be a way to set the server's port. Likewise, when running a client via `autoconnect`, there doesn't appear to be a way to specify a port to connect to.

This attempts to add `server_port` (used only when `server_scenario` is set), `headless_port`, and `autoconnect_port` preferences that take advantage of port parameters in SeriousProton server and client classes/functions that EE doesn't appear to use.

If left undefined, these preferences should default to port 35666, leaving the default behavior of existing installs unchanged.

This is a naive implementation that doesn't attempt to check for port usage (including by the HTTP API) and allows any port from 10 to 65535. It also adds INFO-level log lines that output the server's IP and port, or the IP and port that the autoconnect client is attempting to connect to.

### DIFF
--- a/src/menus/autoConnectScreen.cpp
+++ b/src/menus/autoConnectScreen.cpp
@@ -65,36 +65,47 @@ void AutoConnectScreen::update(float delta)
     {
         std::vector<ServerScanner::ServerInfo> serverList = scanner->getServerList();
         string autoconnect_address = PreferencesManager::get("autoconnect_address", "");
+        int autoconnect_port = PreferencesManager::get("autoconnect_port").toInt();
 
-        if (autoconnect_address != "") {
-            status_label->setText("Using autoconnect server " + autoconnect_address);
+        if (!autoconnect_address.empty())
+        {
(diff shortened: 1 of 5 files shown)


src/menus/autoConnectScreen.cpp
```diff
@@ -65,36 +65,47 @@ void AutoConnectScreen::update(float delta)
     {
         std::vector<ServerScanner::ServerInfo> serverList = scanner->getServerList();
         string autoconnect_address = PreferencesManager::get("autoconnect_address", "");
+        int autoconnect_port = PreferencesManager::get("autoconnect_port").toInt();
```
daid, line 68: would make more sense to parse the address in `host:port` format, which is the usual way on how to specify this, and I'm quite sure that's also how the manual IP entry of connecting works.

oznogon:
Reimplemented as #2438.

PR #2391, opened 2025-06-08
Fixed text position indicator

### LINKED ISSUE #2385: [Bug] Radar bearing indicator drifts from tap/click point when Helms Radar Lock is enabled
The bearing indicator on Helms that appears when tap/clicking on the radar to change course doesn't track the coordinates of the tap/click. The indicator doesn't appear to factor in the current radar rotation.

**Steps to reproduce**

1. Enable Options -> Interface -> Helms Radar Lock.
2. Join a player ship on Helms.
3. Rotate the ship to a bearing other than 0.
4. Tap/click the radar to change the ship's heading.

**Expected behavior**

As with Radar Lock not enabled, the indicator appears on the radar near the point where the tap/click occurred.

**Observed behavior**

The indicator appears at an unexpected position that correlates to the ship's heading at the time the radar is tap/clicked.

Image

Ownezx:
Fixes #2385

### DIFF
--- a/src/screens/crew6/helmsScreen.cpp
+++ b/src/screens/crew6/helmsScreen.cpp
@@ -52,7 +52,20 @@ HelmsScreen::HelmsScreen(GuiContainer* owner)
             {
                 auto r = radar->getRect();
                 float angle = vec2ToAngle(position - transform->getPosition());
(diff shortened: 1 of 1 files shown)


Ownezx:
I did an ugly rotation, but it could be simplified by just getting the mouse to screen position.

Issue #2398, opened 2025-06-09
[Enhancement] F1 menu for GM


Ownezx:
The GM should have an F1 menu like other screens.

PR #2402, opened 2025-06-09
[#2398] Add F1 keyboard reference to GM screen

oznogon:
Adds keyboard reference menu to GM screen.

The `General` hotkey category is mostly irrelevant to the GM screen (tabbing through player screens or selecting via function key), but it also includes the Lua console hotkey. This PR also reassigns the Lua hotkey to its own category `Console` so it can be included on both the GM and player screens.

Screenshot 2025-06-09 at 8 40 48 AM


oznogon:
I'm not sure how voice chat works on the GM screen, but those hotkeys might benefit from being split from the `Basic` category if they aren't relevant here.

Issue #1800, opened 2022-12-18
Don't show unbound hotkeys in hotkey help popup


oznogon:
If a hotkey isn't bound, don't show it in the F1 help popup.

![Screenshot from 2022-12-17 23-57-21](https://user-images.githubusercontent.com/19192104/208287553-a7fbee01-64b1-49d3-aa89-9de7bf49c09d.png)

Ownezx:
I think they should be greyed but not removed to allow users to see what can be assigned.

PR #2403, opened 2025-06-09
Added a check for unbound keys in F1 menu

Ownezx:
Handled #1800, 

- I do not show unbound keys
- If some keys are hidden I show an indicator at the bottom of the help screen.

image

### DIFF
--- a/src/screens/crewStationScreen.cpp
+++ b/src/screens/crewStationScreen.cpp
@@ -78,7 +78,8 @@ CrewStationScreen::CrewStationScreen(RenderLayer* render_layer, bool with_main_s
     if (PreferencesManager::get("music_enabled") == "1")
     {
         threat_estimate = new ThreatLevelEstimate();
-        threat_estimate->setCallbacks([](){
+        threat_estimate->setCallbacks([]()
+        {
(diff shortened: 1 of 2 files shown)


src/screens/crewStationScreen.cpp
```diff
@@ -108,56 +108,98 @@ string CrewStationScreen::populateShortcutsList(CrewPosition position)
...
 
+    bool show_additional_shortcuts_string = false;
+
     // Add shortcuts for this position.
     for (auto binding : sp::io::Keybinding::listAllByCategory(getCrewPositionName(position)))
     {
-        ret += binding->getLabel() + ": " + binding->getHumanReadableKeyName(0) + "\n";
+        if(binding->isBound()){
+            ret += binding->getLabel() + ": " + binding->getHumanReadableKeyName(0) + "\n";
+        }
```
daid, line 118: please have the code style consistent. All this code is having curly brackets on new lines except for your new code.

daid: Not wanting to be annoying, but `else{` isn't fixed ;-) I know it's just minor but it helps me in reading the code faster.

Issue #2373, opened 2025-05-30
[Feature Request] Warp jump abortion


Morby21:
Thx to Xansta help me found this place to post an idea.

One idea I’d like to suggest is the ability to abort a warp jump. I fully understand and appreciate the current delay due to warp charging — it adds tension and makes escaping from combat less trivial. However, the inability to cancel a warp once initiated feels somewhat limiting and occasionally frustrating from a gameplay perspective.

Yes, I’m aware that cutting power to the warp drive in Engineering can effectively pause the jump, but you can’t adjust the target distance anymore at that point.

To prevent players from abusing warp toggling, a reasonable penalty could be that the warp drive has to recharge from zero, and the energy cost is still deducted as if the jump had completed. If this change doesn’t fit everyone’s vision of how warp should work, maybe it could be added as an optional gameplay setting.

PR #2408, opened 2025-06-09
[#2373] Allow JumpControl to abort a jump in progress

oznogon:
Instead of disabling the Jump button in `GuiJumpControls` after a jump is initiated, keep it enabled but change its theme and text to "Abort". If tap/clicked, the jump in progress is aborted.

Screenshot 2025-06-09 at 4 43 40 PM

To discourage Helms from frequently aborting jumps, aborting a jump consumes a portion of the charge from a full jump, and floods the jump system with heat. Aborting a jump close to completion consumes a larger proportion of the charge, and generates more heat than allowing the jump to complete. The additional heat can also cause significant system damage to the jump drive if it isn't properly cooled before the jump is aborted. (If coolant is not in use on the ship, aborting the jump directly damages the jump drive system.)

Aborting a jump doesn't directly consume reactor energy because energy will be spent to recharge the wasted drive charge.

### DIFF
--- a/src/systems/jumpsystem.cpp
+++ b/src/systems/jumpsystem.cpp
@@ -48,6 +48,7 @@ void JumpSystem::update(float delta)
                 if (entity.hasComponent<Coolant>())
                     jump.addHeat(jump.heat_per_jump);
 
+                jump.charge -= jump.distance;
                 jump.delay = 0.f;
             }
         } else {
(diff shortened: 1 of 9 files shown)


oznogon:
Open to changes in abort behavior, such as always fully draining the drive charge, damaging the jump drive system directl or damaging other systems, removing the jump drive component completely ("ejecting the core"), etc.

src/systems/jumpsystem.cpp
```diff
@@ -48,6 +48,7 @@ void JumpSystem::update(float delta)
                 if (entity.hasComponent<Coolant>())
                     jump.addHeat(jump.heat_per_jump);
 
+                jump.charge -= jump.distance;
```
oznogon, line 51: This moves the jump charge consumption to after a jump is successful, which allows the aborted jump to consume a different amount. Another solution might be to consume `jump.charge` `onUpdate` as we do `jump.delay`.

daid:
Maybe also set the `jump.just_jumped` on an aborted jump to indicate the jump drive did something? Maybe not to 2.0 but like .5 to just have a minor effect.

Overall, looks good. I think the "cost" aborting is good. Only thing I am thinking about is that aborting is pretty much free if you directly do it. Maybe we should clamp it to always cost at least 10%?

PR #2412, opened 2025-06-11
Stopped rotation when key up

### LINKED ISSUE #2389: [Enhancement] Increase precision for key movement
Problem,

Because the movement is done in 5 degrees increment when using the turning keys, it is difficult to have a precise bearing adjustment leading to a necessity of using the mouse for precise bearings.

Solution,

By sending a command to stay on the same baring on key up might lead to better precision and intuitive interaction with stations maneuvering the ship.

Ownezx:
Fixes #2389.

Needs some play testing to compare how it feels when it comes to lag.

### DIFF
--- a/src/screens/crew6/helmsScreen.cpp
+++ b/src/screens/crew6/helmsScreen.cpp
@@ -136,12 +136,23 @@ void HelmsScreen::onUpdate()
 {
     if (my_spaceship && isVisible())
     {
-        auto angle = (keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue()) * 5.0f;
-        if (angle != 0.0f)
+        static bool was_turning = false;
+
+        auto transform = my_spaceship.getComponent<sp::Transform>();
+
+        if (transform)
         {
-            auto transform = my_spaceship.getComponent<sp::Transform>();
-            if (transform)
+            auto angle = (keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue()) * 5.0f;
+            if (angle != 0.0f)
+            {
+                was_turning = true;
                 my_player_info->commandTargetRotation(transform->getRotation() + angle);
+            }
+            else if (was_turning)
+            {
+                was_turning = false;
+                my_player_info->commandTargetRotation(transform->getRotation());
+            }
         }
     }
 }
--- a/src/screens/crew1/singlePilotScreen.cpp
+++ b/src/screens/crew1/singlePilotScreen.cpp
@@ -162,13 +162,24 @@ void SinglePilotScreen::onUpdate()
 {
     if (my_spaceship && isVisible())
     {
-        auto angle = (keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue()) * 5.0f;
-        if (angle != 0.0f)
-        {
-            if (auto transform = my_spaceship.getComponent<sp::Transform>())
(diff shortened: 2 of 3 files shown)


src/screens/crew6/helmsScreen.cpp
```diff
@@ -136,12 +136,23 @@ void HelmsScreen::onUpdate()
...
+            if (angle != 0.0f)
+            {
+                was_turning = true;
                 my_player_info->commandTargetRotation(transform->getRotation() + angle);
+            }
+            else if (was_turning)
+            {
+                was_turning = false;
+                my_player_info->commandTargetRotation(transform->getRotation());
+            }
```
oznogon, line 155: I think the bigger issue is that we're still sending a network command every tick, and each one overrides the last. A better solution might be to track keydown/up between updates and send `commandTurnSpeed` only when it changes.

With `float turn_request = 0.0f;` in the class on `helmsScreen.h`:

```diff
-        static bool was_turning = false;
+        auto current_turn_request = keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue();
 
-        auto transform = my_spaceship.getComponent<sp::Transform>();
-
-        if (transform)
+        if (current_turn_request != turn_request)
         {
-            auto angle = (keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue()) * 5.0f;
-            if (angle != 0.0f)
-            {
-                was_turning = true;
-                my_player_info->commandTargetRotation(transform->getRotation() + angle);
-            }
-            else if (was_turning)
-            {
-                was_turning = false;
-                my_player_info->commandTargetRotation(transform->getRotation());
-            }
+            LOG(INFO) << "commandTurnSpeed network request happening";
+            turn_request = current_turn_request;
+            my_player_info->commandTurnSpeed(turn_request);
```

Otherwise, if we instead make 1- or 2-degree commands per tick, we then don't lose enough precision to benefit as much from precisely tracking keydown state. You could simplify this code back down and just use a smaller modifier than `5.0f` to the angle, such as:

```
        if (auto transform = my_spaceship.getComponent<sp::Transform>())
        {
            auto angle = 2.0f * (keys.helms_turn_right.getValue() - keys.helms_turn_left.getValue());
            if (angle != 0.0f)
                my_player_info->commandTargetRotation(transform->getRotation() + angle);
        }
```

This can still be less precise on ships with slower rotation rates, like an Atlantis (1/10 second keydown at 10.0f rotation = 1 degree expected, but this would send a 2-degree command, so the ship would turn one degree more than expected after keyup). For ships with rotation rates between 20 and 120, it's not noticeable; AFAICT the fastest-turning stock player ship is a Hornet at 300% maneuvering power, which doesn't break 100.0 deg/sec (32.0 default * 3 = 96).

Ownezx:
I just tested with joystick and I need to fix it to not send a stupid amound ot network data.

Ownezx:
First draft for joystick support with deadzone and a extreme value deadzone to clamp the high an low values to -1, 0, 1.

Additionally I added a resolution so users do not spam the server with requests as the user moves the joystick.

Ownezx:
This depends on a new PR on serious proton : https://github.com/daid/SeriousProton/pull/259

Ownezx:
Should be good now with a PR in serious proton: https://github.com/daid/SeriousProton/pull/259https://github.com/daid/SeriousProton/pull/259


Default deadzone setting is in main but could move it.

Issue #1275, opened 2021-01-15
Italian translation issues


oznogon:
Post-#1272, there are some UI issues with strings.

- Some new strings include accented characters, which don't render (#293).
- Some longer strings overflow their GUI elements.

<img width="758" alt="Screen Shot 2021-01-14 at 7 26 06 PM" src="https://user-images.githubusercontent.com/19192104/104677512-625cc300-569e-11eb-8476-3236a2d81bce.png">
<img width="452" alt="Screen Shot 2021-01-14 at 7 25 45 PM" src="https://user-images.githubusercontent.com/19192104/104677517-62f55980-569e-11eb-9405-a00c5fe3b3ab.png">
<img width="604" alt="Screen Shot 2021-01-14 at 7 25 35 PM" src="https://user-images.githubusercontent.com/19192104/104677518-638df000-569e-11eb-9287-7e4b638c6a76.png">
<img width="127" alt="Screen Shot 2021-01-14 at 7 24 40 PM" src="https://user-images.githubusercontent.com/19192104/104677519-638df000-569e-11eb-8147-74ebdc39ce0d.png">
<img width="857" alt="Screen Shot 2021-01-14 at 7 24 35 PM" src="https://user-images.githubusercontent.com/19192104/104677520-638df000-569e-11eb-82b5-0e54967c3599.png">
<img width="249" alt="Screen Shot 2021-01-14 at 7 23 42 PM" src="https://user-images.githubusercontent.com/19192104/104677522-638df000-569e-11eb-89ef-cc9d10ec247d.png">
<img width="389" alt="Screen Shot 2021-01-14 at 7 23 37 PM" src="https://user-images.githubusercontent.com/19192104/104677523-64268680-569e-11eb-952b-0738c99ad620.png">
<img width="320" alt="Screen Shot 2021-01-14 at 7 23 32 PM" src="https://user-images.githubusercontent.com/19192104/104677524-64268680-569e-11eb-8370-160b576bbacd.png">
<img width="421" alt="Screen Shot 2021-01-14 at 7 23 28 PM" src="https://user-images.githubusercontent.com/19192104/104677526-64268680-569e-11eb-8b9b-76d19bde7951.png">
<img width="251" alt="Screen Shot 2021-01-14 at 7 23 04 PM" src="https://user-images.githubusercontent.com/19192104/104677528-64bf1d00-569e-11eb-8070-74e13fc26294.png">
[...]

PR #2430, opened 2025-06-16
Add tube size labels to missile fire button

### LINKED ISSUE #1728: [Request] Indicator for tube size inside the firing button
I know that there is the possibility of displaying the tubesizes as text via lua.
But I think having them directly visible at the Buttons would be great.

Some possible ways to to this:
- capitalization like: right(small), Right(medium),RIGHT(large)
- resizing of the icon(rocket/mine)
- resizing of the button

hawson:
Adds labels to missile tube fire buttons; closes #1728?


GinjaNinja32:
perhaps something like this instead?


(icons are smaller with a `-` for small tubes, larger with a `+` for large tubes, unchanged from present for medium)

unsure which characters exactly it should be, but from a quick peek at the current translation files we probably can't do untranslated S-M-L in context of other translated strings, and probably not even translated ones:

| Language | "S" | "M" | "L" |
|--------|--------|--------|--------|
| CZ | **M** | **S** | V |
| DE | K | M | G |
| FR | **L** | M | **L** |
| IT | P | M | G |

oznogon:
> One idea I had was to use stacked dots (or bars), like these: `⠁⠃⠇` or `⠸⠰⠠` (unicode Braille).

Could pips in the icons themselves indicate this?

image

aBlueShadow:
> The text overflow is a fair critique, but as you said: it happens anyway; it has been brought up in #1275 and elsewhere. 

It happens on italian anyway, but rarely on the other languages.  But with that PR, it does overflow on all currently supported languages except english

hawson:
I find the "pips" a bit hard to see on small screens, so I re-worked the icons with "bars" instead:
montage
Since all of the icons are 128x128, try viewing at 33% or 25% zoom.

Ownezx:
Bars I really like. I didn't even know there were tube sizes before this issue!

nallath:
> Bars I really like. I didn't even know there were tube sizes before this issue!

Yeah, I added those because i felt that fighters were too strong. Although we often gave them a small amount of missiles, if you had 5 fighters, they could each fire their single missile and devastate a huge ship.

So now most fighters have small tubes (and some bombers have a single medium one)

In the same vein you can also have medium ships with a single huge missile if they are meant to be more like an "artillery" class.

oznogon:
+1 to bars

hawson:
I'm spending more time rotating images than I am coding.  At least the rotation and tiling is scripted. ;-)
tile

hawson:
And here's an in-game display of all 3 sizes at once:
missle_bars

daid:
"Ribbed for extra pleasure"

Ownezx:
Perhaps using 1/2/4 lines would make more sense with the damage of the missiles (see #2453)

hawson:
I don't think the icons are large enough to support 4 lines/bars? Three bars use most of the linear space on the current icons, but if the icon is reworked to be a bit "longer," it might work.  I'll see what I can do with the SVGs (and thus new PNGs).  

The Database mentions the damage ratios in passing on the "Weapons" page: "...wary of [large missiles] doubled damage potential" and "[small missiles] only deal half the damage of their basic counterparts...", but it's a little buried.

PR #2432, opened 2025-06-16
Higher resolution nebulas assets

Jadael:
This PR replaces the original nebula assets with NASA-based, fractal-detailed textures and implements several technical improvements to maintain gameplay balance and visual consistency.

3D view, in the exact middle of the nebula:
image

First person, 10U away:
image

Radar:
image

### Radar Readability (Fixed)
Initial high-brightness textures made asteroids invisible and text unreadable on radar
- Created separate, purpose-built radar traces instead of reusing 3D textures
- Reduced additive brightness to 70% (max level 180/255)
- Added 7 new radar trace variants designed specifically for 2D radar view

### Visual Consistency (Maintained)
New textures initially changed the established aesthetic to orange-heavy palette
- Adjusted color profiles to match original reddish/greenish/blueish scheme
- Maintained "big background object" design philosophy
- Preserved clear distinction from ship traces

### 3D Rendering Improvements
Implemented golden spiral arrangement for more natural volumetric appearance
- 24-48 cloud layers (vs previous 32 fixed)
- Size gradient from 5120 units (center) to 1024 units (edge)
- Consistent volume appearance from all viewing angles

## Technical Changes

| | Old | New | Rationale |
|-----------|-----|-----|-----------|
| Textures | 3 files, ~32KB each | 3 files, ~546KB average | Higher detail, fractal noise |
| Radar Traces | Reused 3D textures | 7 dedicated files, ~1.7MB each | Optimized for 2D readability- could probably be scaled down significantly without visual quality loss |
| Arrangement | Random within a circle | Golden spiral (137.508°) | Consistent volumetric illusion from all angles |
| Cloud Count | Fixed 32 | Random 24-48 | To maintain variation between nebula |
| Render Distance | 20,000 units | 40,000 units | Visible in main view whenever they are visible in long range view- and out to the same distance as radar signature range (2x Long Range range) |

## Performance Considerations
- Asset Size: ~13MB additional storage for radar traces- Helms and Weapons do see them at about "screen size" though- maybe use JPG instead of PNG?
- **Runtime Impact**: Keeping the image assets from exceeding 70% brightness feels about right in-game- could maybe be done in-code instead

## Visual Verification Needed
Need to verify these scenarios look acceptable:

1. **Ship & Asteroid visibility in nebulae** (short-range radar)
2. **Text readability** over nebula areas 
3. **Consistency** with original aesthetic
4. **Distance rendering** (close/medium/far views)

## Files Changed
- `resources/Nebula[1-3].png` - Replaced with new high-res versions
- `resources/radar/nebula[1-7].png` - New dedicated radar traces based on actual top-down appearance of nebula
- `scripts/api/entity/nebula.lua` - Updated generation algorithm- golden spiral instead of random for visual cohesiveness, random number of cloud layers to maintain natural variation

## Backward Compatibility
- Maintains same radar blocking radius (5U), assets scaled to match
- Preserves existing gameplay mechanics

---

#### Original Post:

Higher resolution nebulas, based on NASA images, with fractal noise for interesting detail in different view scales.

image

Please be aware, it does seem to make a visual bug with nebulas in 3D view more obvious, where they are rendered a little smaller than they should be when you are further away and they don't overlap visually in the same way their real size does.

### DIFF
--- a/scripts/api/entity/nebula.lua
+++ b/scripts/api/entity/nebula.lua
@@ -10,7 +10,7 @@ function Nebula()
     local e = createEntity()
     e.components.radar_signature = {gravity=0, electrical=0.8, biological=-1.0}
     e.components.transform = {rotation=random(0, 360)}
-    e.components.radar_trace = {icon="radar/nebula.png" .. irandom(1, 7) .. ".png", min_size=0, max_size = 9999999, radius=radius*1.15, blend_add=true}
+    e.components.radar_trace = {icon="radar/nebula" .. irandom(1, 7), min_size=0, max_size = 9999999, radius=radius*1.15, blend_add=true, rotate=true}
     e.components.radar_block = {range=radius}
     e.components.never_radar_blocked = {}
     
@@ -36,6 +36,6 @@ function Nebula()
     end
     
     e.components.nebula_renderer = render_info
-    e.components.nebula_renderer.render_range = 20000
+    e.components.nebula_renderer.render_range = 40000
     return e
 end


oznogon:
These nebula images' color intensity causes some readability issues on radar, which might not be undesirable but are notable changes.

Asteroids in nebulae are significantly more difficult to see on short-range radar:

Screenshot_20250616_171713

Areas of near-white intensity make text, radar traces, and heading indicators unreadable:

Screenshot_20250616_171641
Screenshot_20250616_171622

Jadael:
> The additive nature and high brightness can do that indeed. Reducing the color in the radar_trace component for the nebula might help in this case, while still giving a full on effect in the 3D view?

Actually, now that you say that it makes me wonder, should nebula have a proper radar trace as ships and other objects do? Are there any other objects that show on the radar as full-color objects like this besides nebulas? I could probably think up some options following the same design principles as the ship traces.

oznogon:
> Are there any other objects that show on the radar as full-color objects like this besides nebulas

Wormholes and black holes also do.

aBlueShadow:
While Wormholes and blackholes  have no "wireframe" like traces, but still have a different visuals on the radar and 3d view. So nebulae are the only outlier here, as they use the same image for both. Maybe we could use the old images for radar, and the new ones for 3d?

Btw. I just saw that the radar traces for wormholes seem to be broken in ECS

Jadael:
Ah, what I was interpreting as a bug with nebula's 3D view above was actually mostly me misunderstanding how their cloud array worked. I think what's actually going on is that nebula would just have to be tweaked to accommodate these new images, since they're so different. I'll be playing around with this more anyway, so my plan is to:

- Adjust the radar trace color, *maaaaybe* the size too if I zoom in real close and feel like I'm not happy with the radar blocking radius syncing up with the graphic, but I did try to follow the existing nebula's gradient so it's close.
- Adjust the cloud layers in nebula_renderer to better accommodate the new images and try to get it to look like a 3D interpretation of the 2D radar trace, as it did before

Just to put those thoughts out there in case anyone else is bored and tries them before I do.

daid:
> Actually, now that you say that, it makes me wonder, should nebula have a proper radar trace, as ships and other objects do? Are there any other objects that show on the radar as full-color objects like this besides nebulas? I could probably think up some options following the same design principles as the ship traces. (EDIT: Though, allow me say, I love the way both old and new nebula look on the scope and it would be a shame to lose that.)

The nebula are more big background objects, just like black holes and wormholes. So they are fine to have a different visual style, as they are different. If they would be some "blip" like trace, you could mistake them for something you could not fly through.

daid:
Another thing, the 2nd image has this rectangular shape it in:
image
It gives the illusion of that part being less in the nebula compared to other bits, which isn't the case.

Jadael:
- billboard sprites darkened a bit (max brightness level 180 of 255) so they add together more "softly" and block less from view in 3D view.

- nebula.lua tweaked to arrange the billboards in a golden spiral (so that they feel volumetric from all angles as consistently as possible), with sizes tweaked so that the new billboard images overlap "just right" (subjectively, based on flying around a bit and looking at different magic numbers for min and max size)

- radar traces now have separate graphics based on how the nebula now actually appear in-game in top down view, with a few artistic tweaks (some blurring) to make it look nicer in 2D, instead of showing one of the available 2D billboards. Since they are based on the now darkened billboard sprites, these are a bit darker now too and don't interfere with radar traces and labels as much.

image

image

image

image

---

Here is what is happening in top-down. Each cloud sprite's size increases from 1024 at the rim to x5 that at the center, so from the side they don't feel *round*, but they do feel like solid clouds.:

image

aBlueShadow:
There is maybe not much we can do about it without loosing density, but on first person view, the nebulae is still very bright due to the additive rendering. So this post is basically for the sake of completeness
nebulae

Jadael:
They're currently at, effectively, "70% brightness", so that could be brought down more (probably worth trying 50%, 25%, and 10% for thoroughness), and we could play with their gamma curve too. 

I think the question is design now: how opaque *should* they be?

Comparing old and new histograms:

Old:
image

New:
image

PR #2438, opened 2025-06-21
Add noninteractive server ports

oznogon:
The server creation menu provides a field for setting a port, and the server browser allows the port to be specified in a manually entered IP or domain.

When running a server via `server_scenario` or `headless`, there's no way to set the server's port. Likewise, when running a client via `autoconnect`, there's no way to specify a port to connect to.

This PR:

- Adds a `server_port` PreferencesFile setting. This is used when either `headless` or `server_scenario` is set.
- Allows the `autoconnect_address` PreferencesFile setting to optionally specify a custom port delineated by a colon (`server.example.com:35669`).
- Adds INFO-level log lines that output the server's IP and port, or the IP and port that the autoconnect client is attempting to connect to.
- Adds WARNING-level log lines if the port number is out of valid range (< 10 or > 65535).
- Displays the port number on the autoconnect connection screen.

For example, one can launch a server with IP 192.168.1.50 and port 35669 with

```
EmptyEpsilon headless=scenario_00_basic.lua server_port=35669
```

or

```
EmptyEpsilon server_scenario=scenario_00_basic.lua server_port=35669
```

and then autoconnect weapons to that server on a different client by running

```
EmptyEpsilon autoconnect=1 autoconnect_address=192.168.1.50:35669
```

Both take advantage of port parameters in SeriousProton's server and client classes/functions that EmptyEpsilon doesn't appear to use.

If left undefined, these preferences should default to port 35666, leaving the default behavior of existing installs unchanged. Invalid values, such as non-numeric strings or port numbers under 10 or over 65535, also revert to the default port.

This is a naive implementation that doesn't attempt to check for port usage (including by the HTTP API).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -219,7 +219,7 @@ int main(int argc, char** argv)
         if (server_port < 10 || server_port > 65535)
         {
             new EpsilonServer(defaultServerPort);
-            LOG(INFO) << "Launching server_scenario " << server_scenario << " on default port " << defaultServerPort;
+            LOG(WARNING) << "Invalid server_port " << server_port << ". Launching server_scenario " << server_scenario << " on default port " << defaultServerPort;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -270,18 +284,34 @@ void returnToMainMenu(RenderLayer* render_layer)
...
     {
-        new EpsilonServer(defaultServerPort);
+        // Create the server to listen on the assigned port.
+        // Use the default port if server_port isn't set or has an invalid
+        // value (toInt returns 0).
+        int headless_port = PreferencesManager::get("server_port").toInt();
+        // This is the same process as server_port and could be made DRY.
+        if (headless_port < 10 || headless_port > 65535)
+        {
+            new EpsilonServer(defaultServerPort);
```
daid, line 297: Instead of doing `new EpsilonServer` in both the if and else, I find it easier to read:
```
if (headless_port < 10 || headless_port > 65535) {
  LOG(Warning, ...);
  headless_port = default_port;
}
new Server(..);
```

PR #2440, opened 2025-06-23
Add F1 keyboard references to cinematic, top-down views

oznogon:
Screenshot_20250622_212922
Screenshot_20250622_212529


src/screenComponents/helpOverlay.h
```diff
@@ -2,6 +2,7 @@
 #define HELP_OVERLAY_H
 
 #include "gui/gui2_element.h"
+#include "gui/gui2_panel.h"
```
daid, line 5: why is this needed? It was forward declared to reduce header fanout.

PR #2444, opened 2025-06-23
Thingy improvements

GinjaNinja32:


Ownezx:
Do you think it would be worth pushing a tiny bit further and create a skybox system that allows to switch the skybox when you enter a nebula for instance ??

src/components/zone.h
```diff
@@ -14,6 +14,7 @@ class Zone
     string label;
     glm::vec2 label_offset;
     string skybox;
+    float skybox_fade_distance; // distance from edge of zone for skybox to fully fade in
```
daid, line 17: missing default value, which means it will be "random".

PR #2446, opened 2025-06-24
Add beam weapons to science db

oznogon:
The Weapons category of the Science db doesn't describe beam weapons. Add them to the base science DB script.

### DIFF
--- a/scripts/science_db.lua
+++ b/scripts/science_db.lua
@@ -107,6 +107,17 @@ It has been reported that some battleships started using larger variations of th
 
 Smaller variations of these missiles have become common in the galaxy, too. Fighter pilots praise their speed and maneuverability, because it gives them an edge against small and fast-moving targets. They only deal half the damage of their basic counterparts, but what good is a missile if it does not hit its target.]]))
 
+local item = weapons:addEntry(_('Beam weapons'))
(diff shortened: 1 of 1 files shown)


scripts/science_db.lua
```diff
@@ -107,6 +107,17 @@ It has been reported that some battleships started using larger variations of th
 
 Smaller variations of these missiles have become common in the galaxy, too. Fighter pilots praise their speed and maneuverability, because it gives them an edge against small and fast-moving targets. They only deal half the damage of their basic counterparts, but what good is a missile if it does not hit its target.]]))
 
+local item = weapons:addEntry(_('Beam weapons'))
+item:addKeyValue(_('Range'), 'Varies')
+item:addKeyValue(_('Damage'), 'Varies')
+item:setLongDescription(_([[Beam weapons emit an instantaneous, focused burst of energy or matter at a single target within a target arc. Many ships equip beam weapons for their precision and versatility.
```
oznogon, line 113: Beams default to energy-type damage, and I don't think any default ship templates set a beam's damage type. However, beams can be set to any damage type with `setBeamWeaponDamageType()`.

daid: Yeah, the idea is that scenarios could have EMP beams, but none of the default ships do this.

Issue #2459, opened 2025-06-25
[Enhancement] Move power-damage indicator from missile tube to missile load/unload button


oznogon:
**Issue**

When missile system effectiveness is 0 due to power or damage, the power-damage indicators (PDI) appear over only the launch controls, and not over the load/unload controls.

![Image](https://github.com/user-attachments/assets/84629121-6ce8-4eba-9e75-ee7deb00c5ec)

However, only missile loading is affected; loaded missiles can still be launched.

![Image](https://github.com/user-attachments/assets/4a2c1a23-9e47-4dc4-88a5-fe592588d691)

Also, if a tube is in the middle of being un/loaded when the system is fully degraded, the progress bar stops and the PDI does not appear.

![Image](https://github.com/user-attachments/assets/0291393d-30e9-40c4-ac62-c5552ae67074)

This results in confusion among Weapons/Tactical players as to whether missiles can be launched or un/loaded when the system is fully degraded.

**Suggested change**

Move the power-damage indicator from the tube to the load/unload button.

Credits to Bridge Command discord users Lt. Kane and Ymbirtt for the suggestion.

PR #2465, opened 2025-06-26
Move "Damaged" overlay to missile tube "Load" button

hawson:
Addresses #2459.

Also disables the Load (and "Unload") and Fire buttons when the missile system is 100% damaged, or at 0 power.

### DIFF
--- a/src/screenComponents/missileTubeControls.cpp
+++ b/src/screenComponents/missileTubeControls.cpp
@@ -70,6 +70,9 @@ void GuiMissileTubeControls::onUpdate()
             row.layout->hide();
         return;
     }
+    auto sys = ShipSystem::get(my_spaceship, ShipSystem::Type::MissileSystem);
+    float health = sys->health;
+    float power_level = sys->power_level;
     for (int n = 0; n < MW_Count; n++)
     {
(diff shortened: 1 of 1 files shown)


src/screenComponents/missileTubeControls.cpp
```diff
@@ -90,14 +93,21 @@ void GuiMissileTubeControls::onUpdate()
...
             rows[n].fire_button->setText(getTubeName(tube.direction) + ": " + tr("missile","Empty"));
             rows[n].loading_bar->hide();
             break;
         case MissileTubes::MountPoint::State::Loaded:
-            rows[n].load_button->enable();
             rows[n].load_button->setText(tr("missile","Unload"));
-            rows[n].fire_button->enable()->show();
+            if ((health <= 0) || (power_level <=0)) {
+                rows[n].fire_button->disable();
+                rows[n].load_button->disable();
```
daid, line 106: the lacking of show might cause a problem in some edge cases, like opening the weapon screen when the system is broken.

daid: Hope that makes things a bit clearer?

daid: Honestly, the code is a bit messy. And there might be edge cases where this currently would go wrong still :)

PR #2475, opened 2025-06-28
Database back button outside of entrylist

Ownezx:
Removed the button from the list and added a margin to make it stand out.

Screenshot 2025-06-28 at 00 00 37


daid:
I also like the more usage of layout managers in GUIs (those were added only in the 3th GUI system refactor, so not used by a lot of code)

PR #2477, opened 2025-06-29
Swap info and custom buttons bar

aBlueShadow:
Swaps the custom buttons with for the normal info sidebar so the science description and deep scan info do not overlap with the waypoint and comms buttons on the left.

I originally placed the custom sidebar intentionally to the right, as I had ideas for a compromise of the old glass cockpit idea, where the smaller bar would reach into the main screen section. However, readability has to be prioritised, of course.


aBlueShadow:
After your post, I saw that the overlap with zoom and the description text is even worse, but that should be rather easy to fix

PR #2482, opened 2025-07-02
Remastered, higher resolution radar traces for ships and items

Jadael:
Redrawn radar traces, based on the original designs, plus various other glyphs in the same style.

image


Jadael:
What's the max size for that "big texture" threshold? Are we thus *limited* to low resolution radar traces?
- If so, should this be combined with a change to how radar traces are stored and rendered?
- Or, if we must stick with low res, does scaling the icons to ship size even make sense at all?

> that where not discussed

These have been under discussion for weeks, and the extras especially have gone through many versions based on direct feedback, so I am sad I have not heard any feedback from you until just now 😥

If we go back to the drawing board, here was my rationale: Since most items in the game use 'blip.png', it seems like there are *missing* traces to be filled in to help make the game world more visible and readable to Helms, Weapons, and Science.

Ownezx:
While I do agree that some of these icons still need work I do think it's a step in the right direction towards having better icons.

Would there be a way for icon size to be a setting or is that too much of a pain to do with templates being in Lua?
Perhaps two folders with two resolutions and a setting?

Right now the main improvement I see are the following:
- Line width should be larger in general in the context of your screenshot resolution, I'm thinking of certain ships, asteroids and mines in particular
- Perhaps the circle be removed for mines and it just being the lines for more readability
- Same goes with asteroids
- the dots that for ship bearing indicator are less readable than lines used previously with the resolution shown on the screenshot
- The master SVG file for the icons should be included in the commit allowing to modify it in the future
- The [?] and [+] symbol seem to break in style from all the other icons
- I find the unscanned ship icon less visible compared to the previous filled arrow (perhaps just filling the shape would work)

daid:
> What's the max size for that "big texture" threshold? Are we thus _limited_ to low resolution radar traces?
> 
>     * If so, should this be combined with a change to how radar traces are stored and rendered?
> 
>     * Or, if we must stick with low res, does scaling the icons to ship size even make sense at all?
128x128 is the limit for putting in the atlas. Now, what I would propose, is that certain icons can be higher res then others. Stations for example could really do with some higher resolution radar traces. But no need for missiles to be that big, as they are always rendered small.

And for performance, that goes hand in hand there, things that are small are rendered in large quantities, and that's where the save on performance comes from. It's not an issue for any reasonably modern hardware, but there are people running EE on Raspiberry Pis, and those do not have great rendering performance.

> > that where not discussed
> 
> These have been under discussion for weeks, and the extras especially have gone through many versions based on direct feedback, so I am sad I have not heard any feedback from you until just now 😥
Well, to make it a bit more clear. I can only find some stuff from 2 weeks ago, at that time I was dealing with major headaches from a concussion. So I didn't partake in any discussions, and I assumed people would link you to the "new mine icon" issue, where I pretty much disliked all options there, for simple visual clarity at lower resolutions.

However, my KEY issue why I made that comment is, you name this change "high resolution traces". While it's two things in one "high resolution traces + major overhaul of visual style". [Bait&switch]

> If we go back to the drawing board, here was my rationale: Since most items in the game use 'blip.png', it seems like there are _missing_ traces to be filled in to help make the game world more visible and readable to Helms, Weapons, and Science.

Oh, let me be clear, I think it's good to have more options here. And clean up things. However, it's too much at once. Like, I cannot focus on "what sizes do certain icons need to be" or "this new mine trace isn't readable on small scale" or "these icons will be useful for specific scenarios without them even being used per default". The bigger the change, the larger the chance of it just being rejected for some reason.

Jadael:
I felt this was a minimal and faithful-as-possible redraw of the visual style, so I suppose I am mistaken about what your understanding of the visual style is, and would appreciate any guidance which could be given.

Issue #2466, opened 2025-06-26
Widget drop down on Science holds over to Relay


Xansta:
GUI edge case in the current release. I did not try it on the ECS branch. If I click the Tactical/Systems drop down (in this case it also has a description) and don't select one of those and then go over to the Relay screen, that drop down stays open and visible on the Relay screen

![Image](https://github.com/user-attachments/assets/38457399-eb83-4176-9dbc-2593c2e32b82)

oznogon:
This issue is not limited to this `GuiSelector`; all `GuiSelectors` appear to be affected. I've also reproduced with the beam frequencies and subsystem target selectors on Weapons and other stations with beam controls, and with the shield frequency selectors on Weapons, Eng+, etc.

PR #2484, opened 2025-07-04
[#2466] Hack GuiSelector popup fix using string IDs

oznogon:
Attempt to work around #2466 by assigning string IDs to popups generated by clicking a `GuiSelector`, then iterating through children of `CrewStationScreen` and hiding them when changing tabs.

### DIFF
--- a/src/gui/gui2_selector.cpp
+++ b/src/gui/gui2_selector.cpp
@@ -95,3 +95,9 @@ void GuiSelector::onMouseUp(glm::vec2 position, sp::io::Pointer::ID id)
         popup->show()->moveToFront();
     }
 }
+
+void GuiSelector::onFocusLost()
+{
+    // Explicitly hide the popup when the selector loses focus.
+    popup->hide();
+}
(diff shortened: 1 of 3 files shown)


daid:
I see how it fixes this specific instance of the problem, but I don't like the hackyness of the solution.

I wonder if somehow the loss of focus on the selector could be used to hide this, as I'm quite sure it will get focus as soon as you open the popup, and loses focus as soon as you click something else.

oznogon:
> I see how it fixes this specific instance of the problem, but I don't like the hackyness of the solution.

I don't like this solution either, but the popup implementation itself is hacky for being parented to the highest-level UI element.

> I wonder if somehow the loss of focus on the selector could be used to hide this, as I'm quite sure it will get focus as soon as you open the popup, and loses focus as soon as you click something else.

I had experimented with implementing `GuiSelector::onFocusLost() { popup->hide() }` before coming to this solution, but `onFocusLost()` doesn't trigger when the crew screen is changed via keybind.

oznogon:
> With the focus lost, it does then disappear as soon as you click something else I assume?

No. Even if I click something else, the popup persists until I return to the crew screen containing the selector that created the popup.

If I have clicked something else on a different crew screen, the popup disappears only after I return to the originating crew screen.

If I haven't clicked something else on a different crew screen and then return to the originating crew screen, the popup persists until I click on something else on the originating crew screen.

----

I've reverted the string check and tried again with `GuiSelector::onFocusLost() { popup->hide() } `, this time also setting `focus(main_panel)` in `CrewStationScreen::showTab()` to force a focus change as if clicked. This seems to work.

PR #2494, opened 2025-07-08
Make the GM mouse cursor modal

oznogon:
This PR builds upon #2493.

Add modal mouse cursor sprites for different GM screen selection modes:

- mouse_add if holding <kbd>Shift</kbd> (add to selection)
- mouse_faction if holding<kbd>Alt</kbd> (select only entities with the same faction as the faction selector)
- mouse_ship if holding <kbd>Ctrl</kbd> (select only ships and entities with docking bays)
- mouse_create if creating an entity (Create... button)

[Screencast_20250707_231736.webm](https://github.com/user-attachments/assets/db892cb2-6a55-4caa-a1d1-08619309a924)


daid:
I think the cursor change can persist outside of the GM screen if you hold shift and then press home? Other then that, nice!

PR #2496, opened 2025-07-09
Make cross-type templates behave more sensibly, like legacy

GinjaNinja32:
Two example cases:
- `PlayerSpaceship():setTemplate("MT52 Hornet")` ("Player"), a player ship with a "CPU ship" template
- `CpuShip():setTemplate("MP52 Hornet")` ("CPU"), a CPU ship with a "player ship" template

| Component           | Current | New    | Notes |
|---------------------|---------|--------|-------|
| ai_controller       | Both    | CPU    | If the template does not set AI, then CPU/CPU. |
| comms_receiver      | Player  | CPU    | |
| comms_transmitter   | CPU     | Player | |
| long_range_radar    | CPU     | Both   | The AI system expects CPU ships to have radar ranges. |
| coolant             | CPU     | Player | |
| hacking_device      | CPU     | Player | |
| reactor             | CPU     | Player | |
| scan_probe_launcher | CPU     | Player | |
| science_scanner     | CPU     | Player | |
| self_destruct       | CPU     | Player | |

### DIFF
--- a/scripts/api/entity/playerspaceship.lua
+++ b/scripts/api/entity/playerspaceship.lua
@@ -21,6 +21,14 @@ function PlayerSpaceship()
         transform = {rotation=random(0, 360)},
         callsign = {callsign=generateRandomCallSign()},
         scan_state = scan_state,
+        reactor = {},
+        coolant = {},
+        self_destruct = {},
+        science_scanner = {},
+        scan_probe_launcher = {},
+        hacking_device = {},
+        long_range_radar = {},
+        comms_transmitter = {},
     }
     e:setFaction(__default_player_ship_faction)
     return e
--- a/scripts/api/entity/shiptemplatebasedobject.lua
+++ b/scripts/api/entity/shiptemplatebasedobject.lua
@@ -26,10 +26,19 @@ function Entity:setTemplate(template_name)
             comp[key] = value
         end
     end
-    if template.__type == "station" then
+    if not comp.ai_controller and not comp.player_control then
+        -- no AI or player control, this is a station
         comp.physics.type = "static"
-    elseif template.__type == "playership" then
-        if comp.shields then comp.shields.active = false end
+    elseif comp.shields then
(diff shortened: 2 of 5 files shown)


scripts/api/entity/playerspaceship.lua
```diff
@@ -21,6 +21,14 @@ function PlayerSpaceship()
         transform = {rotation=random(0, 360)},
         callsign = {callsign=generateRandomCallSign()},
         scan_state = scan_state,
+        reactor = {},
```
daid, line 24: I don't like this. Part of the reason I added all these to playership templates, is that you can create fighter templates that don't need engineering. And thus those need to be able to remove a whole bunch of systems that aren't really for fighters.

scripts/api/entity/shiptemplatebasedobject.lua
```diff
@@ -26,13 +26,27 @@ function Entity:setTemplate(template_name)
             comp[key] = value
         end
     end
-    if template.__type == "station" then
+    if not comp.ai_controller and not comp.player_control then
+        -- no AI or player control, this is a station
```
daid, line 30: This doesn't have to be true, a station could have lasers/missiles and then it also needs AI. It's not something implemented yet, but this would rule that out completely.

daid (commented):
In general, I think this is the wrong direction. The templates are not designed for cross-use like this. And I much rather see some simple checks in :setTemplate() to check for certain odd combinations and report that. (like having both player and ai_controller, or an ai_controller and energy/coolant)

GinjaNinja32:
Updated.
Ship capability components (reactor, coolant, hacking, etc) are back on the templates.
Control-type components (AI/player control, comms tx/rx) are still in CpuShip/PlayerSpaceship/SpaceStation and are maintained across template changes.
Long range radar is still on all ships, since the AI code does process it. There's TODOs in the component, there's definitely some player-only bits in there (waypoints!), but the long/short radar range bits should be on AI ships too.

PR #2511, opened 2025-07-15
Update scenario Kessler to use getSectorName()

Xansta:
Use `getSectorName()` instead of hard coded sector names. Prep for sector name changes.


Xansta:
This is similar to the changes to scenarios in requests #2343 #2341 #2340 etc., that were made because `typeName` was deprecated with ECS.

PR #2515, opened 2025-07-16
Expose i18n plural handling to Lua

GinjaNinja32:
Depends on https://github.com/daid/SeriousProton/pull/270

Changes applied to scenario_00_basic as an example; this does not affect the existing translations since both French and German agree on 1 being singular and 2+ being plural; if the scenario was translated into Czech (as the only other language with a `main.*.po` file), it would require three forms (for 1, 2-4, and 5+); assuming Google Translate is in any way close to correct this would look something like:

```po
msgctxt "time-incCall"
msgid "%s, you have one minute remaining."
msgid_plural "%s, you have %d minutes remaining."
msgstr[0] "%s, zbývá ti %d minuta"
msgstr[1] "%s, zbývají ti %d minuty"
msgstr[2] "%s, zbývá ti %d minut"
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -761,7 +761,7 @@ if(Python3_Interpreter_FOUND)
 endif()
 
 add_custom_target(update_locale
-    COMMAND xgettext --keyword=tr:1c,2 --keyword=tr:1 --keyword=trMark:1c,2 --keyword=trMark:1 --add-comments=TRANSLATORS --omit-header -d resources/locale/main.en ${MAIN_SOURCES} ${GUI_LIB_SOURCES}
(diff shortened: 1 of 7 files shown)


aBlueShadow:
> if the scenario was translated into Czech (as the only other language with a main.*.po file). 
Italian has a main.*.po file as well, so has the pending PR for simplified chinese #2374.
I don't know anything about plurals in those languages though, so I have no idea it would matter in this context.

GinjaNinja32:
not sure how I missed the Italian file.

there's no plural header in either set of files, but based on gettext docs Italian follows the same pattern as English. Not sure what Chinese does, gettext docs don't mention it.

PR #2522, opened 2025-07-17
Make Relay's "Center On Ship" button a toggle button

oznogon:
Requires #2521.

Use `RadarView::auto_center_target` to convert Relay's "Center On Ship" button from a one-off view change to a toggle that follows the ship when enabled. Since this uses code in RadarView that continues following an internally docked ship, it no longer needs to apply its own workaround.

[Screencast_20250717_160352.webm](https://github.com/user-attachments/assets/3adb5823-cfcb-464c-bfea-b7fb3cafd663)


oznogon:
Rebased to use #2521 as merged.

PR #2524, opened 2025-07-18
Allow callsign toggle and reimplement on Spectate

oznogon:
The callsign toggle hotkey on the Spectator screen doesn't work. Implement a RadarView state for callsigns, allow the hotkey to toggle it, and prevent callsign rendering when turned off.

[Screencast_20250717_230511.webm](https://github.com/user-attachments/assets/caba5907-f988-41e6-9ace-7f55151626a2)


src/screenComponents/radarView.cpp
```diff
@@ -86,7 +86,7 @@ GuiRadarView::GuiRadarView(GuiContainer* owner, string id, float distance, Targe
     show_waypoints(false),
     show_target_projection(false),
     show_missile_tubes(false),
-    show_callsigns(false),
```
daid, line 89: Don't understand why this default was changed? Seems like a change instead of a regression from stable?

oznogon: This is the code in screenComponents/radarView.cpp on 2024.12.08 that uses `show_callsigns` but is no longer present in `master`:

```
        if (obj != *my_spaceship && rect.overlaps(object_rect))
        {
            obj->drawOnRadar(renderer, object_position_on_screen, scale, view_rotation, long_range);
            if (show_callsigns && obj->getCallSign() != "" && obj->getDockedStyle() != DockStyle::Internal)
                renderer.drawText(sp::Rect(object_position_on_screen.x, object_position_on_screen.y - 15, 0, 0), obj->getCallSign(), sp::Alignment::Center, 15, bold_font);
        }
```

daid: Ok, then the display of callsigns on all radars could be a regression from legacy compared to master. I see legacy as leading for behavior, not master.

oznogon: OK, I'll revert to align to the legacy implementation.

Issue #2526, opened 2025-07-20
Database: Set the order of the "Faction Relationship" labels to be alphabetical order


parracodea:
In the Database > Factions page, the "Friendly/Neutral/Enemy" labels appear to be in an arbitrary order, which makes it a bit difficult to find what you're looking for. Setting the labels to be alphabetical would help, which is also the same order as the buttons on the left.

<img width="1202" height="932" alt="Image" src="https://github.com/user-attachments/assets/99887090-b41e-4492-9b84-3e6e0ae227ec" />

oznogon:
Fixed by #2529.

PR #2529, opened 2025-07-20
[#2526] Sort faction relationship key/values alphabetically by locale name

oznogon:
Requires #2528.

To sort the faction relationship key/value entries by locale name and match the sorting in the DatabaseView list, sort the internal keys by their locale name equivalents.

image

### DIFF
--- a/scripts/science_db.lua
+++ b/scripts/science_db.lua
@@ -254,7 +254,23 @@ function __fillDefaultDatabaseData()
 	local faction_database = ScienceDatabase():setName(_("database", "Factions"))
 	for name, info in pairs(__faction_info) do
 		local entry = faction_database:addEntry(info.components.faction_info.locale_name);
+
+		--[[ DB entries use locale-independent internal names.
+		     Sort the list of internal names by localized name to match the DatabaseView list.
+		     Lua in EE sorts by character index (for instance uppercase A > lowercase z), so lowercase the locale name for sorting purposes.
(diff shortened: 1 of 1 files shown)


oznogon:
As noted in the comment, this is an incomplete solution because accented letters are sorted after all unaccented letters, so for instance the "É" in French "Éxuaris" is sorted after the "M" in French "MSU". Because this behavior is also true in DatabaseView, it isn't addressed in this PR, which aims to be consistent with the DatabaseView sort.

scripts/science_db.lua
```diff
@@ -253,22 +253,39 @@ function __fillDefaultDatabaseData()
...
 	local faction_database = ScienceDatabase():setName(_("database", "Factions"))
 	for name, info in pairs(__faction_info) do
-        local entry = faction_database:addEntry(info.components.faction_info.locale_name);
+		local entry = faction_database:addEntry(info.components.faction_info.locale_name);
+
+		--[[ DB entries use locale-independent internal names.
+		     Sort the list of internal names by localized name to match the DatabaseView list.
+		     Lua in EE sorts by character index (for instance uppercase A > lowercase z), so lowercase the locale name for sorting purposes.
+		     Accents also break sorting, which for example affects the French translation of "Exuari" as "Éxuaris", which are listed last after MSU (USN).
+		     Since this isn't handled in DatabaseView on the C++ side either, this sorting is also "broken" her in the same manner for consistency.
```
oznogon, line 262:
```suggestion
		     Since this isn't handled in DatabaseView on the C++ side either, this sorting is also "broken" here in the same manner for consistency.
```

PR #2534, opened 2025-07-23
Enhance CrewPositionSelection panel

oznogon:
Refine the CrewPositionSelection panel's layout.

- Use layout features to compose the UI with consistent padding and margins around component panels.
- Constrain the parent panel's width to a fixed value.
- Remove the panel border on the right column to provide more space for text.
- Split the right column into info and crew listings with a label heading.

Before:

image

After:

Screenshot_20250722_175604


Xansta:
Just that little bit of width management makes it less likely to require scrolling and makes the text look better even if scrolling is required.

PR #2535, opened 2025-07-24
Extend GuiScrollText scroll length to fit descenders

oznogon:
When the final line of a `GuiScrollText` box ends with characters containing descenders, such as j, y, p, and q, the descender is cut off. Extend the scrolling height of `GuiScrollText` and `GuiScrollFormattedText` by 25% of a single line's height to accommodate these descenders.

Before (final line of the Doomed Outpost description on the scenario selection screen):

Screenshot_20250723_224749

After:

Screenshot_20250723_224722

### DIFF
--- a/src/gui/gui2_scrolltext.cpp
+++ b/src/gui/gui2_scrolltext.cpp
@@ -31,7 +31,8 @@ void GuiScrollText::onDraw(sp::RenderTarget& renderer)
 {
     auto text_rect = sp::Rect(rect.position.x, rect.position.y, rect.size.x - scrollbar->getSize().x, rect.size.y);
     auto prepared = sp::RenderTarget::getDefaultFont()->prepare(this->text, 32, text_size, selectColor(colorConfig.textbox.forground), text_rect.size, sp::Alignment::TopLeft, sp::Font::FlagClip | sp::Font::FlagLineWrap);
-    auto text_draw_size = prepared.getUsedAreaSize();
+    // Extend the scroll by a fraction of a line height to accomodate descenders (jypq etc.)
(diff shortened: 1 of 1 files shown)


oznogon:
It's wrong for calculating the height of the last line of scrolling text, but it's correct for the line height everywhere else. Consequently fonts without descenders, such as the old all-uppercase font that EE formerly used, never looked wrong.

SeriousProton uses the glyph's `horiBearingX` and `horiBearingY` metrics for the bounds `position` in `FreetypeFont::getGlyphInfo`. The `bearingY` value extends _only from the font's baseline_, so it never includes the descender:

https://freetype.org/freetype2/docs/tutorial/step2.html:

> horiBearingY
>     For horizontal text layouts, this is the vertical distance from the current cursor position (on the baseline) to the topmost border of the glyph image's bounding box.

image

Since `getUsedAreaSize` relies on the glyph's `position` y coordinate for the height, it measures the glyph height only from the baseline, ignoring the descender. So we can either track the glyph's `height` metric which includes ascenders and descenders, or track a descender value (`height` - `horiBearingY`) and tack it onto the `max_y` value in `getUsedAreaSize`.

FreeType's metrics designed for descenders aren't reliable across fonts since fonts aren't obligated to follow FreeType's standards:

>  Unfortunately, font formats don't define the descender in a uniform way. For some formats, it represents the descent of all capital latin characters (without accents), for others it is the ascent of the lowest accented character, and finally, other formats define it as being equal to bbox.yMin. ... Notice that the values of the ascender and the descender are not reliable (due to various discrepancies in font formats), unfortunately.

So I'm not sure if there's a better solution. Filed daid/SeriousProton#271.

PR #2552, opened 2025-10-18
Revert default (non-autoconnect) tutorial repeating behavior to pre-#2310 defaults

oznogon:
PR #2310 changed the default behavior of the `tutorial` Preferences File option to avoid repeating tutorials, with the intent of allowing running a tutorial in EE only once when `autoconnect` was enabled.

Revert this change in default behavior and instead implement the non-repeating tutorial behavior only when `autoconnect` is set.

Filed as an alternative to #2548, which instead creates a new Preferences File option to define whether tutorials repeat.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -199,7 +199,7 @@ int main(int argc, char** argv)
 #endif //STEAMSDK
 
     // Tutorial scenario to load relative to the scripts path, such as
-    // "tutorials/00_all.lua" to run all tutorials.
+    // "tutorial/00_all.lua" to run all tutorials.
     string tutorial = PreferencesManager::get("tutorial");
 
     if (!tutorial.empty())


src/tutorialGame.cpp
```diff
@@ -221,7 +222,8 @@ void TutorialGame::finish()
         sp::ecs::Entity::destroyAllEntities();
         instance->hideAllScreens();
 
-        gameGlobalInfo->startScenario("tutorial.lua");
+        gameGlobalInfo->startScenario(instance->filename);
+        engine->setGameSpeed(1.0);
```
oznogon: This `setGameSpeed` call prevents the reloaded scenario from being paused upon start.

oznogon:
Duplicated by merged #2548

Issue #2550, opened 2025-10-09
[Enhancement] Destroy Asteroids with Missiles


ewingwm:
Currently Asteroids are static and cannot be destroyed with weapons, like missiles.

It would be beneficial if the asteroids could be destroyed to enable passage through fields.

oznogon:
Asteroid physics are set to `sensor`, same as missiles. The issue seems to be in `MissileSystem::collision` L147-148:

```cpp
    auto hull = b.getComponent<Hull>();
    if (!hull) return;
```

Asteroids lack a `Hull` component, so missiles do nothing when they collide with them. 

Adding a `Hull` component to asteroids in `scripts/api/entity/asteroid.lua` allows them to be destroyed upon contact with missiles:

```diff
 function Asteroid()
     ...
     local e = createEntity()
     e.components = {
     ...
+        hull={
+            current=1,
+            max=1,
+            allow_destruction=allow_destroy
+        },  
     ...
 end
```

This also makes them targetable by Weapons/Tactical, and destroyable with beams if the Asteroids are set to a hostile faction.

However, this also makes asteroids targetable by Science and Relay, which report on them as if they're ships and allow hacking by default, which isn't ideal.

PR #2553, opened 2025-10-19
Add a non-Hull Health component

oznogon:
The `Hull` component serves the function of a health property for entities, but its presence also enables other features in an entity, including:

- allowing missiles to deal damage to the entity upon collision
- making the entity a valid target of Weapons, selectable/hackable by Relay, and scannable by Science
- triggering hull damage indicator effects on player-controlled entities
- factoring into faction-based reputation awards when the entity is destroyed
- factoring into threat level estimates
- being repairable if docked in a `DockingBay` with the `Repair` flag enabled

These additional properties complicate making certain non-ship entities destructible via damage without enabling these additional properties.

For example, adding the `Hull` component to asteroids allows them to be destroyed when a missile collides with them. However, it also allows Science to select asteroids as if they were ships, and also allows Relay to select and attempt to hack asteroids.

This PR adds, replicates, and exposes to scripting a `Health` component, and also makes `Hull` inherit its features. This makes `Health` effectively identical in function to `Hull`, including separate callbacks for taking damage and being destroyed.

However, `Health` doesn't return as `true` when other features check whether the target has a `Hull` component. This allows scenario authors to allow entities lacking a `Hull` component to sustain and be destroyed by definable amounts of damage, but doesn't also make them targetable, scannable, or hackable.

If an entity has both `Hull` and `Health` components, depleting the entity's `Hull` no longer destroys it, but should instead begin depleting its `Health`. Once the entity's `Health` is depleted, the entity is then destroyed.

With new callbacks on Health damage and upon destruction from Health depletion, entities without hulls can now also have scripted callbacks upon damage and destruction from damage.

This PR also:

- Adds `Health` to asteroids and assigns them 1 point of health by default.
- Displays Health in the GM and Spectate screens' info displays when/if an entity has the `Health` component.
- Adds `Health` as a component to the GM Tweak menu.
- Modifies `MissileSystem::collision` to deal damage to entities with either `Hull` or `Health`.
- Adds `get`/`setHealth()`, `get`/`setHealthMax()`, `onTakingHealthDamage()` and `onHealthDestruction()` functions to the Lua SpaceObjects API.

This PR does not add the `Health` component to any other entities. The primary change in default EE behavior should be that asteroids are destroyed upon taking damage from a missile collision (#2550).

### DIFF
--- a/scripts/api/entity/shiptemplatebasedobject.lua
+++ b/scripts/api/entity/shiptemplatebasedobject.lua
@@ -1,3 +1,4 @@
+require("utils.lua")
 local Entity = getLuaEntityFunctionTable()
 
 ----- Old ShipTemplateBasedObject API -----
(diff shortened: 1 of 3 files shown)


oznogon:
Closing to look into alternatives

PR #2608, opened 2025-11-29
Add Lua isGamePaused() function

oznogon:
Provide a Boolean isGamePaused() function for
headless server confirmation and HTTP API usage.

Example:

```
> pauseGame()
> isGamePaused()
true
> unpauseGame()
> isGamePaused()
false
```


daid:
maybe "isGamePaused" as a function name sounds more natural?

Issue #2450, opened 2025-06-24
[Enhancement] Smoother line draw


Ownezx:
Perhaps doing a shader would be something to do.

Possble example https://github.com/mitkus/linren

Issue #2380, opened 2025-06-05
[Feature request] Bearing line for helms and single pilot


Ownezx:
Problem,

When moving using the arrow key, the only indication of bearing is on the top left or bottom right. When in a dogfight, the player must either change to use the mouse (and potentially not be able to do combat maneuvers if they use the mouse for them), or look outside of the radar screen to get the precise bearing.

This makes precision maneuvers difficult during a fight.

Solution,

Adding a line parallel to the ship direction to the specific bearing allows for the player to have an approximate idea of where they are heading while keeping their attention on the relevant information.

Example as follows:

<img width="715" alt="Image" src="https://github.com/user-attachments/assets/f675841a-66e7-4a02-8420-871901626792" />

Ownezx:
After thinking a little more, I thought having the target bearing could also be something interesting.

A system with two indicators, one with the current heading and one with the target heading (after clicking) could be interesting.

I did a quick draft of what such a thing could look like. A limitation would be that the arrow would at times be hidden behind the combat maneuver screen. A mix of line and arrow might make it better.

![Image](https://github.com/user-attachments/assets/a1d6fe18-d1c9-43bf-9634-f2e227d0b4e1)

PR #2611, opened 2025-11-29
Helms Heading Indicator and Keyboard Turning Adjustment.

JonathanILevi:
I think this new indicator for target heading in Helms is nice and helpful.  I think it also makes the turning behavior more intuitive (I had a couple new players yesterday on Helms that didn't get it at first, I think this would have helped them. ...I could also just get better at explaining XD)

For the keyboard turn left and right keys: It always annoying how the ship keeps turning after letting go of the turn key.  I don't think the current behavior is preferred, if it is, I think a new turn-by key should be added.  I think the fix I did here is adequate and easy.

image

### DIFF
--- a/src/screenComponents/radarView.cpp
+++ b/src/screenComponents/radarView.cpp
@@ -807,13 +807,8 @@ void GuiRadarView::drawHeadingIndicators(sp::RenderTarget& renderer)
     {
         renderer.drawLine(
             radar_screen_center + vec2FromAngle(target_heading - 90 - view_rotation) * (scale - 0),
-            radar_screen_center + vec2FromAngle(target_heading - 90 - view_rotation) * (scale - 30),
+            radar_screen_center + vec2FromAngle(target_heading - 90 - view_rotation) * (scale - 40),
             {255, 255, 255, 255});
-            
-        renderer.drawSprite("redicule.png", radar_screen_center + vec2FromAngle(target_heading - 90 - view_rotation) * (scale - 50), 48);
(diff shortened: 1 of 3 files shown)


oznogon:
The keyboard changes seemed familiar: @Ownezx's #2389 and #2412, which stalled out.

EDIT: Same for the helms heading indicator, #2380 and #2390, which also added target bearing indicators and options toggles.

JonathanILevi:
This is a simpler change, not having a heading number, just a simple long tick.  I think a heading number would be good, but not sure how, and this is simpler...so harder to disagree with?

image

oznogon:
I like that indicator quite a bit better, especially when paired with active waypoints since the line clearly connects the heading request to the waypoint's bearing. The heading number stays readable and above the touch contact area so hands/fingers don't obscure it.

I also like it a bit better than #2390. The line connects it to the heading interface better IMO than the red ship-like deltas. If this lands and #2450 for line-drawing enhancements also gets addressed, I'd come back to this and see if the line could be made to stand out a little more.

For me the only question is whether this approach to Helms turning by keyboard is better than #2412, or perhaps if that part of this PR should be split off into its own PR since it doesn't depend on the indicator implementation. Keyboard controls on Helms aren't something that I have much playing experience with, and #2412 also attempted to address deadzones (and the SP dependency for that did land in daid/SeriousProton#259).

JonathanILevi:
I can separate out the turning adjustment, I only included it because I thought they were related and cumbersome to have a bunch of tiny pull requests.

PR #2634, opened 2025-12-12
Allow manual control of GuiRotatingModelView

oznogon:
Implement optional manual control of GuiRotatingModelView elements, such as those used in DatabaseView.

The default initial behavior of the model rotating within the view is unchanged, except for new FoV checks that attempt to better fit the model in the viewport when the viewport is vertically oriented. On load, the model spins automatically within the viewport. When clicked/tapped, this feature allows the model view to be manually manipulated.

- Implement mouse click, drag, scrollwheel on element
  - Single left click resets to default spinning view behavior
  - Left click-drag rotates the model around its center, with constraints
  - Mouse wheel up zooms in, down zooms out
- Default view FoV constrained to avoid clipping in vertical aspect ratios; see https://www.litchiutilities.com/docs/fov.php on horizontal/vertical FoV conversion logic
- Enabled by default; `GuiRotatingModelView::setManualRotationAllowed(false)` to disable
- Touch interactions that are read as mouse inputs also work

https://github.com/user-attachments/assets/d1b4fba4-4cd7-4bb9-aea3-e47fb1c5bf8e

### DIFF
--- a/src/screenComponents/rotatingModelView.cpp
+++ b/src/screenComponents/rotatingModelView.cpp
@@ -25,9 +25,7 @@ GuiRotatingModelView::GuiRotatingModelView(GuiContainer* owner, string id, sp::e
 
 void GuiRotatingModelView::onDraw(sp::RenderTarget& renderer)
 {
-
-    if (rect.size.x <= 0) return;
-    if (rect.size.y <= 0) return;
+    if (rect.size.x <= 0 || rect.size.y <= 0) return;
 
     auto mrc = entity.getComponent<MeshRenderComponent>();
     if (!mrc) return;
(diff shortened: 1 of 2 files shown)


src/screenComponents/rotatingModelView.cpp
```diff
@@ -42,27 +40,45 @@ void GuiRotatingModelView::onDraw(sp::RenderTarget& renderer)
...
+    float aspect_ratio = rect.size.x / rect.size.y;
+    auto projection_matrix = glm::perspective(glm::radians(camera_fov), aspect_ratio, near_clip_boundary, 25000.f);
+
+    // Calculate distance needed to fit the model in both dimensions based on
+    // the element's aspect ratio.
+    float vertical_fov_rad = glm::radians(camera_fov / 2.0f);
+    float horizontal_fov_rad = glm::atan(glm::tan(vertical_fov_rad) * aspect_ratio);
+
+    // Calculate distance needed to fit each dimension to fill viewport, then
+    // apply fill percentage and zoom.
```
oznogon: Redundant comment.

```suggestion
```

PR #2662, opened 2025-12-25
[#171] Implement subtle banking rotation on turns

### LINKED ISSUE #171: Request: Ship Banking Animation
I'm sure this is probably a huge amount of work, but I'd like to suggest that at some point a banking animation is added to the ships (or at least the option to toggle banking animation), I know that it makes little practical sense in space, but it helps break the feeling of "driving" in space.

oznogon:
Resolves #171.

Before:

[Screencast_20251225_014448.webm](https://github.com/user-attachments/assets/7807f845-e341-44ad-b253-6f3ec724b115)

After:

[Screencast_20251225_014100.webm](https://github.com/user-attachments/assets/68a15208-7c13-4ba5-9cb2-1f8d73e4c00c)

### DIFF
--- a/src/systems/rendering.cpp
+++ b/src/systems/rendering.cpp
@@ -7,7 +7,7 @@
 #include <glm/gtc/type_ptr.hpp>
 #include "tween.h"
 #include "random.h"
-
+#include "components/maneuveringthrusters.h"
 
 std::vector<RenderSystem::RenderHandler> RenderSystem::render_handlers;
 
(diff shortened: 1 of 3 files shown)


src/systems/rendering.cpp
```diff
@@ -120,6 +120,13 @@ void drawMesh(MeshRenderComponent& mrc, ShaderRegistry::ScopedShader& shader)
 
 void MeshRenderSystem::update(float delta)
 {
+    // Update banking angle for all physics entities.
+    for (auto [entity, mrc, transform, physics] : sp::ecs::Query<MeshRenderComponent, sp::Transform, sp::Physics>())
+    {
+        float target_bank_angle = 0.0f;
+        target_bank_angle = std::clamp(physics.getAngularVelocity(), -4.0f, 4.0f);
```
daid: Won't this cause things like asteroids (that always rotate) to bank as well?

oznogon: I thought the Spin component set rotation directly each tick independent of physics, instead of setting angular velocity:

https://github.com/daid/EmptyEpsilon/blob/9405eaf272e1f7e92be6808e1dcc63e8987f88cb/src/systems/basicmovement.cpp#L16-L18

But I can narrow this PR's component selection further to the ManeuveringThrusters component

PR #2666, opened 2025-12-27
Implement SP setBaselineOffset as theme parameter

oznogon:
Adds the `offset` parameter to styles to define a float value used to offset font rendering via SeriousProton's setBaselineOffset() function.

Requires https://github.com/daid/SeriousProton/pull/287.


resources/gui/default.theme.txt
```diff
@@ -1,6 +1,8 @@
 [base] {
     font: gui/fonts/BigShouldersDisplay-SemiBold.ttf
-    
+    # Positive offset moves the font down, negative moves it up.
+    offset: 0.0
```
daid: I think `font_offset` might be a better name then `offset` as now it isn't clear that it refers to the font

PR #2668, opened 2025-12-27
Implement theme inheritance

oznogon:
Add an optional `inherit` keyword to the `[base]` section of themes. When defined with one or more theme names in a comma-separated list, all style properties not defined in the theme are inherited from the specified parent themes.

All themes inherit from the `default` theme, which EE already requires.

Also adds an example inherited theme, `default-bebas`, which overrides fonts with the BebasNeue font already in the repository.

### DIFF
--- a/resources/gui/default-bebas.theme.txt
+++ b/resources/gui/default-bebas.theme.txt
@@ -3,7 +3,7 @@
 # The `inherit: default` property is implied.
 
 [base] {
-    font: gui/fonts/BebasNeue Book.otf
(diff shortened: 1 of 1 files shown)


daid:
This is a very good idea!

I am wondering on the implementation. Wouldn't the code become a lot simpler if you merge at the `getFlattenNodesByIds` step?

oznogon:
I was trying to avoid traversing nested maps, especially at onDraw, but maybe in this case nested maps might not be an issue if I only do it when loading/processing the theme.

PR #2684, opened 2026-01-04
[#1062] Add MissileTubes MountPoint Fired state

### LINKED ISSUE #1062: trigger hardware/DMX event when missile is launched

oznogon:
Adds a Fired state to missile tubes, equivalent to the Empty state, for triggering scripting events when a missile is fired. The Fired state persists for 0.1 seconds after firing a tube, or after HVLIs exit the Firing state. Also adds TubeFired0... hardware events.

Fixes #1062.

Minimal hardware.ini example:

```ini
[hardware]
device = VirtualOutputDevice
channels = 1

[channels]
FiringMissile = 1

[event]
trigger = TubeFired1
target = FiringMissile
runtime = 0.5
value = 1.0
```

[Screencast_20260104_100120.webm](https://github.com/user-attachments/assets/e31b400c-0569-4bde-86a7-f38855d9ec11)

### DIFF
--- a/src/multiplayer/missiletubes.cpp
+++ b/src/multiplayer/missiletubes.cpp
@@ -40,7 +40,6 @@ BASIC_REPLICATION_IMPL(MissileTubesReplication, MissileTubes)
         VECTOR_REPLICATION_FIELD(type_loaded);
         VECTOR_REPLICATION_FIELD(state);
         VECTOR_REPLICATION_FIELD(delay);
-        VECTOR_REPLICATION_FIELD(fired_delay);
         VECTOR_REPLICATION_FIELD(fire_count);
     VECTOR_REPLICATION_END();
 }


src/multiplayer/missiletubes.cpp
```diff
@@ -40,6 +40,7 @@ BASIC_REPLICATION_IMPL(MissileTubesReplication, MissileTubes)
         VECTOR_REPLICATION_FIELD(type_loaded);
         VECTOR_REPLICATION_FIELD(state);
         VECTOR_REPLICATION_FIELD(delay);
+        VECTOR_REPLICATION_FIELD(fired_delay);
```
GinjaNinja32: does anything actually read this on the client? the `-= delta` loop is in an `if (game_server)` and as far as I can see that's the only thing that _reads_ the value, rather than just setting it - everything else is driven by `state`

oznogon:
```suggestion
```

PR #2718, opened 2026-02-14
Fix Lua API issues

oznogon:
- Fix broken `string.sub()` calls re: model data that forget to pass the string into the function's first argument
- Add missing coordinate arguments to `ModelData:addTubePosition()`
- Use the `offset_x`/`_y` arguments passed to `orderFlyFormation()` instead of invalid globals `x`/`y`
- Fix deprecated shield functions' dot/colon notation
- Fix broken ScienceDatabase `hasEntries()`, which still doesn't work as documented anyway due to daid/EmptyEpsilon#2615 / daid/EmptyEpsilon#2617
- Use `deepcopy()` to copy modelData by value for `ScienceDatabase:setModelDataName()`, aligned with other uses of `deepcopy()` for modelData
- Fix DockingPort state enum converter incorrectly using copied/pasted Comms values (use `docking` instead of `opening`, `docked` instead of `hailed`)
- Fix `setPlanetCloudRadius()`, warpjammer `get`/`setRange()` missing `components` reference
- Replace incorrect `spin` with function argument `rotation_time` in planet `setAxialRotationTime()`
- Remove unused `hailed_player` state in `isCommsBeingHailed()`
- Fix spaceship `addBroadcast()` color assignments; hostile is gray and neutral is red
- Validate tube index on spaceship `get`/`setWeaponTube...()` functions
- Fix broken value validation in spaceobject `take`/`addReputationPoints()`
- Fix function chaining in stbo `setRadarTrace()`
- Validate mesh_render.scale is present before using values in `addEngineEmitter()`
- Clamp spaceobject `getHeading()` to 0-360 range
- Fix `orderFlyFormation()` assigning a different case to the order string than the enum

### DIFF
--- a/scripts/api/entity/cpuship.lua
+++ b/scripts/api/entity/cpuship.lua
@@ -94,7 +94,7 @@ end
 --- Give multiple CpuShips the same SpaceObject and different offsets to create a formation.
 --- Example: ship:orderFlyFormation(leader, 500, 250) -- fly 0.5U off the wing and 0.25U off the tail of the SpaceObject `leader`
 function Entity:orderFlyFormation(target, offset_x, offset_y)
-    if self.components.ai_controller then self.components.ai_controller = {orders = "fly in formation", order_target=target, order_target_location={offset_x, offset_y}} end
+    if self.components.ai_controller then self.components.ai_controller = {orders = "Fly in formation", order_target=target, order_target_location={offset_x, offset_y}} end
     return self
 end
 --- Orders this CpuShip to move toward the given coordinates, and to attack hostiles that approach within its short-range radar range during transit.


scripts/api/entity/spaceobject.lua
```diff
@@ -27,13 +27,18 @@ end
...
---- Unlike SpaceObject:setRotation(), a value of 0 points to the top of the map ("north").
+--- Sets this entity's heading, in degrees ranging from 0 to 360.
+--- Unlike setRotation(), a value of 0 points to the top of the map ("north"), which matches player radars.
 --- Values that are negative or greater than 360 are converted to values within that range.
 --- SpaceObject:setHeading() and SpaceObject:setRotation() do not change the helm's target heading on PlayerSpaceships. To do that, use PlayerSpaceship:commandTargetRotation().
 --- Example: obj:setHeading(0)
 function Entity:setHeading(heading)
-    if self.components.transform then self.components.transform.rotation = heading + 270 end
+    if self.components.transform then
+        local heading = self.components.transform.rotation + 270
```
daid: This looks wrong, as it's not taking the parameter passed but the current rotation.

PR #2734, opened 2026-02-21
Write log to config dir/EmptyEpsilon.log when launched as a Mac app bundle

oznogon:
When running a packaged build of EE on macOS through typical end-user/non-development means (double-clicking the icon, running `open EmptyEpsilon.app`), the log is still written to STDOUT, which isn't accessible.

Detect when EE is run as an app bundle on macOS, and if so, write EE's log to a file in the config directory (`~/.emptyepsilon/EmptyEpsilon.log`). This captures log output on macOS dev, server, and player runs for troubleshooting without resorting to running the binary on the command line from within the app bundle. This is also a more similar behavior to running EmptyEpsilon on Windows, which also writes the log to a file unless EE is built in DEBUG mode.

Existing development and server/player workflows should be unaffected. Running the binary from within the app bundle (i.e. `cd EmptyEpsilon.app/Contents && MacOS/EmptyEpsilon`) should still log to STDOUT.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -92,21 +92,50 @@ int main(int argc, char** argv)
 #else
     Logging::setLogLevel(LOGLEVEL_INFO);
 #endif
+
+// Log to STDOUT unless on non-debug Windows builds, which won't have
+// terminals for log output.
 #if defined(_WIN32) && !defined(DEBUG)
     Logging::setLogFile("EmptyEpsilon.log");
 #else
     Logging::setLogStdout();
 #endif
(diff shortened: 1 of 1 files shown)


oznogon:
macOS CI failure doesn't seem related. Build succeeds without errors but it hits an `hdiutil: create failed - Resource busy` error and fails.

daid:
The CI failure also happened on master, and I think it's just github action mac runnings changing something again breaking the builds. It's not the first time.

PR #2736, opened 2026-02-22
Add a null guard on my_player_info access

oznogon:
Add a null check for `my_player_info` to prevent a crash on accessing ShipSelectionScreen when my_player_info doesn't exist yet or was destroyed, which can intermittently occur when a proxy is in use.


oznogon:
I'm still trying to figure out a reliable reproduction case, but it's happened enough times in testing reverse proxy instructions to at least want to check nil here. There's also a `my_player_info` nil check further down in `SecondMonitorScreen::update()`.

PR #2769, opened 2026-03-08
hotkeys for adjusting step-by-step

aBlueShadow:
This adds keybindings for 15 degree steps for the manual weapon aiming. When the binding is used, the aim will snap to the next multiple of 15 rather than just adding 15°.
That way you can use rotary encoders or knobs on macro keyboards for aiming. As they only send a short signal each step, they do not work reliably with the weapons_aim_left/weapons_aim_right bindings.
Each step is a 1/24th of a full circle, which means you will hit both 8 cardinal and 12 "clock" directions. It is also pretty close to the 20 steps per rotation that many rotary encoders have, and you won't get a perfect positional match anyway.


aBlueShadow:
Also added keybinding for adjusting scanning parameters stepwise.
I wasn't sure about the step size, but 0.2 seems to be a reasonable compromise between accuracy and speed. I was able to reliably deep-scan multiple objects with a 2-knob keyboard. Of course a 3 or even 4 param that might look different, but those are pretty rare anyways in scenarios, and can be avoided for normal ship scans.

oznogon:
I ask because if we can make it so that any applicable keybind's rate and behavior are tuneable, we wouldn't have to implement stepwise logic ad hoc for each input. Could do it once instead of dozens of times and again for any new binds.

PR #2797, opened 2026-03-18
[#2796] Reorder comms data assignment in station placement

### LINKED ISSUE #2796: [Bug] lua `setCommsFunction(commsStation)` not registering the function
ECS compiled 18Mar2026
Test scenario: Close the gaps (scenario_50_gaps.lua)
When player ship contacts friendly station, they should see the interaction as defined by the function commsStation. Instead they get the default comms script dialog. I verified that the lua code on line 220 of place_station_scenario_utility.lua was being executed:
`station:setCommsScript(""):setCommsFunction(commsStation)`
I tried taking out the `setCommsScript("")` bit, but it did not make a difference.

My apologies if this has already been noted. I looked for an issue that resembled this and did not see one.

oznogon:
In ECS, `setTemplate()` overwrites the entity's CommsReceiver component to use the default comms script.

In `place_station_scenario_utility.lua`, re-initialize `setCommsFunction` immediately after applying `setTemplate` in `placeStation`. On legacy this should be idempotent, on ECS this should restore the expected state.

Should fix #2796.

### DIFF
--- a/scripts/place_station_scenario_utility.lua
+++ b/scripts/place_station_scenario_utility.lua
@@ -96,6 +96,9 @@ function placeStation(x,y,name,faction,size,diagnostic)
 			station:setTemplate(szt())
 		end
 	end
+	if commsStation ~= nil then
+		station:setCommsScript(""):setCommsFunction(commsStation)
+	end
 	if diagnostic == nil then
 		diagnostic = false
 	else


scripts/place_station_scenario_utility.lua
```diff
@@ -96,6 +96,9 @@ function placeStation(x,y,name,faction,size,diagnostic)
 			station:setTemplate(szt())
 		end
 	end
+	if commsStation ~= nil then
+		station:setCommsScript(""):setCommsFunction(commsStation)
```
daid: The setCommsScript is redundant (setting a function overrules a script) and this looks odd, does it this depend on some global?

oznogon:
Closing in favor of #2798

PR #2800, opened 2026-03-19
correct getfont/cachefont loading error. 

gwaland:
This fixes an issue that prevents the application from loading. 

[WARNING ]: Can't find gui/fonts/BigShouldersDisplay-ExtraBold in theme default. Falling back to 'fallback' style.
[WARNING ]: Can't find fallback in theme default. Falling back to 'fallback' style.
[WARNING ]: Can't find fallback in theme default. Falling back to 'fallback' style.

There may be an additional issue where there is a loop if the fallback theme can not initialize before being called as the fallback theme. 
I'm not sure it's actually an issue except in this case where the fallback theme can't be generated because the code can't load it.

### DIFF
--- a/src/hardware/hardwareController.cpp
+++ b/src/hardware/hardwareController.cpp
@@ -292,7 +292,7 @@ void HardwareController::createNewHardwareMappingState(int channel_number, std::
         {
             state.variable = condition.substr(0, condition.find(compare_string)).strip();
             state.compare_operator = compare_operator;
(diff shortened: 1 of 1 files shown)


oznogon:
If I'm understanding the problem correctly, #2801 should at least log and break the loop if that happens. Not sure how it would since the fallback theme is (should be?) populated on GuiTheme init.

PR #2837, opened 2026-04-06
Convert "Request dock" button to a menu

oznogon:
When multiple compatible DockingBay entities are within range, list them on the "Request dock" button and let the player select which entity to dock with instead of automatically choosing the nearest entity.

This converts DockingBay to subclass GuiPanel instead of GuiButton, and uses separate GuiButton and GuiListbox components in order to handle different docking states and dynamically update list contents.

This does NOT CHANGE keyboard shortcut behavior, which continues to dock with the nearest eligible target. It also does NOT CHANGE any existing interactions when only one eligible target is within range.

https://github.com/user-attachments/assets/088782cb-acbf-40e6-a90a-3f25da7e0aa4

### DIFF
--- a/src/screenComponents/dockingButton.cpp
+++ b/src/screenComponents/dockingButton.cpp
@@ -3,6 +3,7 @@
 #include "dockingButton.h"
 #include "gui/gui2_button.h"
 #include "gui/gui2_listbox.h"
+#include "gui/gui2_panel.h"
 #include "systems/collision.h"
 #include "systems/docking.h"
 #include "components/collision.h"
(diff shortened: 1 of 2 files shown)


dkapell:
Testing this out, I realized that the new GuiPanel is being drawn at the same spot/size as the docking button, so on the default theme it looks like a single button, but on my custom theme (for example) it ends up with a second border around the docking button.  To work around this i added:
* `bool visible = true` as a public member of GuiPanel (gui/gui2_panel.h)
* `if (visible) { ... } ` around the `renderer.drawStretchedHV` (gui/gui2_panel.cpp)
* `this->visible = false` at the start of `GuiDockingButton::GuiDockingButton` (screenComponents/dockingButton.cpp)

This allows for the docking button to be drawn the same as before, regardless of theme, while keeping the new functionality.

oznogon:
> Testing this out, I realized that the new GuiPanel is being drawn at the same spot/size as the docking button, so on the default theme it looks like a single button, but on my custom theme (for example) it ends up with a second border around the docking button. To work around this i added:
> 
>     * `bool visible = true` as a public member of GuiPanel (gui/gui2_panel.h)
> 
>     * `if (visible) { ... } ` around the `renderer.drawStretchedHV` (gui/gui2_panel.cpp)
> 
>     * `this->visible = false` at the start of `GuiDockingButton::GuiDockingButton` (screenComponents/dockingButton.cpp)
> 
> 
> This allows for the docking button to be drawn the same as before, regardless of theme, while keeping the new functionality.

Refactored it to no longer be a ~~GuiPanel~~GuiButton-based element, and to conditionally render the new `background_panel` only when expanded.

dkapell:
> Refactored it to no longer be a GuiPanel-based element, and to conditionally render the new `background_panel` only when expanded.

Thanks! I've confirmed on my end that that looks right with my (probably slightly incorrect) theme.

PR #2839, opened 2026-04-07
Feature/probe view toggle button

gwaland:
This allows the toggle probe view button on the screen to be assigned to a key or joystick.  It's useful for people who are creating control surfaces for science. 

It implements it by exposing a toggle method which just calls onclick() internally. 

toggle is called vi the hotkey that was configured.


src/gui/gui2_togglebutton.cpp
```diff
@@ -52,6 +52,10 @@ bool GuiToggleButton::getValue() const
     return value;
 }
 
+void GuiToggleButton::toggle() {
+    onClick();  // trigger the click event used to toggle probe view vie key.
```
daid, line 56: I don't like calling a handler from other code. Better to put the logic in the toggle function and call the toggle function from onClick.

PR #2840, opened 2026-04-07
trying to fix probe view next scannable object

gwaland:
I found that if you're in probe view the next scanable target wasn't working.  This was the quick solution I came up with. 
It checks to see if the probeview is active and if so it will use the probe_transform instead of the ship's transform as the closest target. 

This was tested in my rig.


gwaland:
https://github.com/daid/EmptyEpsilon/pull/2841
This is handled in another pr.

PR #2851, opened 2026-04-10
Refactor player transfers

### LINKED ISSUE #2848: [ECS Regression]: transferPlayersToShip does not reset radar focus
If the players move to a new ship via transferPlayersToShip, the radar screens still remain centered on the old ship. This resets when the players manually log out and into the new ship again.

Image

### LINKED ISSUE #2849: [Bug] transferPlayers... functions allow transferring to crew positions prohibited on the destination ship
If the destination ship has prohibited certain positions, transferPlayers... functions override these. This results in players who can be assigned to positions that they can't/shouldn't be able to access on the ship selection screen.

### LINKED ISSUE #2850: [Bug] transferPlayers... functions retain the old ship's cached password
...

oznogon:
- Fix #2848 by tracking when a radar view should lock onto `my_spaceship` and updating `auto_center_target` appropriately.
- Fix #2849 by checking against the `allowed_positions` mask and transferring players to allowed positions on the destination.
- Fix #2850 by clearing the cached password after transferring players to a new ship. It's up to the scenario author/GM to communicate the new ship's password to players if necessary.


src/script.cpp
```diff
@@ -464,18 +464,63 @@ static int luaGetEnemiesInRadiusFor(lua_State* L)
 
 static void luaTransferPlayers(sp::ecs::Entity source, sp::ecs::Entity target, std::optional<CrewPosition> station)
 {
-    if (!target.getComponent<PlayerControl>()) return;
-    // For each player, move them to the same station on the target.
-    for(auto i : player_info_list)
-        if (i->ship == source && (!station.has_value() || i->hasPosition(station.value())))
-            i->ship = target;
+    // Relevant only to player-controlled entities.
+    auto target_pc = target.getOrAddComponent<PlayerControl>();
```
GinjaNinja32: imo if target doesn't have a PlayerControl, this should be an error, not just add it anyway. if the script intends to transfer to a non-playership entity, it should be the one to add the component, since that typically requires other supporting components to work in a sensible way anyway.

oznogon:
```suggestion
    auto target_pc = target.getComponent<PlayerControl>();
```

src/script.cpp
```diff
@@ -464,18 +464,63 @@ static int luaGetEnemiesInRadiusFor(lua_State* L)
...
+    if (target.hasComponent<AIController>()) target.removeComponent<AIController>();
```
oznogon:
```suggestion
```

src/script.cpp
```diff
@@ -464,18 +464,63 @@ static int luaGetEnemiesInRadiusFor(lua_State* L)
...
+
+    if (!target_pc.allowed_positions.mask)
```
oznogon:
```suggestion
    if (!target_pc || (target_pc && !target_pc.allowed_positions.mask))
```

src/script.cpp
```diff
@@ -464,18 +464,63 @@ static int luaGetEnemiesInRadiusFor(lua_State* L)
...
+    {
+        LOG(Error, "transferPlayersToShip: destination ship has no allowed crew positions. Resetting to Any.");
```
oznogon:
```suggestion
        LOG(Error, "transferPlayersToShip: destination ship has no allowed crew positions.");
```

src/script.cpp
```diff
@@ -464,18 +464,63 @@ static int luaGetEnemiesInRadiusFor(lua_State* L)
...
+        target_pc.allowed_positions = CrewPositions::all();
```
oznogon:
```suggestion
        return;
```

src/screenComponents/radarView.h
```diff
@@ -46,6 +46,7 @@ class GuiRadarView : public GuiElement
     glm::vec2 view_position{0, 0};
     float view_rotation;
     sp::ecs::Entity auto_center_target;
+    bool track_my_spaceship;
```
daid: Put defaults in the header with the variable instead of in the constructor if possible. (I know a lot of older code doesn't do this, but I've been trying to improve that)

src/screenComponents/radarView.h
```diff
@@ -38,45 +39,45 @@ class GuiRadarView : public GuiElement
...
     std::function<void(sp::RenderTarget&)> overlay_func;
 
-    const GuiThemeStyle* radar_style;
-    const GuiThemeStyle* radar_outline_style;
-    const GuiThemeStyle* radar_range_indicators_style;
-    const GuiThemeStyle* radar_sector_grid_style;
-    const GuiThemeStyle* ship_waypoint_style;
-    const GuiThemeStyle* ship_waypoint_background_style;
-    const GuiThemeStyle* ship_waypoint_text_style;
+    const GuiThemeStyle* radar_style = theme->getStyle("radar");
```
daid, line 74: I'm surprised that this work, but, I do like that we removed a whole bunch of duplication :)

PR #2870, opened 2026-05-24
Hotkeys for custom buttons

aBlueShadow:
This adds hotkeys for custom buttons.
Instead of numbering the buttons in order of their appearance(green), I used the position/slot (purple).
buttonorder

Slot based hotkeys are pretty common in games, and this would also allow custom consoles with softkeys next to the screen. But if you prefer it the other way around, that would just be the removal of one line.
15 hotkeys, as that is the maximum that can be displayed (Engineering and most aux stations). 
I wasn't entirely sure whether or not I should make yet another config tab for the custom buttons, but had them placed in general for now.


daid:
As this is mostly to facilitate hardware buttons, I think your choice of making them position based is perfect.

Issue #2874, opened 2026-06-06
[Bug] Zone entities are all located at 0,0 and unexpectedly selectable on the GM screen


oznogon:
All Zone entities are located at coordinates 0,0 regardless of where their points are located or rendered. The GM screen can unexpectedly select the entity location at 0,0 and move it, which moves the offset zone points.

Via @aBlueShadow on Discord

> there seems to be at least some buggy/unintentional behavior. Here I am just doing a small rectangular selection (as I often do for single objects, as it is more reliable than clicking), but somehow selecting a whole array of zones

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/ef336f04-1426-4689-bef8-668f5bebeac5" />

@GinjaNinja32:

> zones don't get their transform.position set to COM like they did on legacy, so they're all at the origin with a really big offset in their polygon points

@daid:

> Dont think there was a technical reason beyond "this was easier"

oznogon:
#2875 is triage for the unexpected disconnect between Transform position and the Zone when using Lua `setPoints()`, but it doesn't address whether or why Zones should be selectable on the GM screen, and it can be circumvented by directly changing the component via member access.

PR #2875, opened 2026-06-06
[#2874] Update Transform on Zone entity to outline's centroid on setPoints()

oznogon:
When setting a Zone's points via Lua `setPoints()`, also set its Transform's position to the outline's centroid.

Before: all Zones' transforms are at the same set of coordinates, typically 0,0

[Screencast_20260606_112640.webm](https://github.com/user-attachments/assets/2bb47365-2b6f-48ff-8395-3fcb0d7a4dad)

After: each Zone's transform is at its centroid

[Screencast_20260606_112235.webm](https://github.com/user-attachments/assets/42c34dc0-953d-4503-814c-7b8a390694f9)


oznogon:
A potential workaround to #2874, but doesn't change the behavior when directly manipulating the Zone component's outline points via Lua member access.

PR #2879, opened 2026-06-14
Add CXX_STANDARD to force C++17 on legacy

oznogon:
Explicitly define [CXX_STANDARD 17](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) in EmptyEpsilon's CMakeLists.txt for the legacy branch, which fixes legacy builds on GCC 16. SeriousProton's legacy branch already does this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,9 @@ endif()
 
 cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 
+# Force C++17 across all compilers
+set(CMAKE_CXX_STANDARD 17)
+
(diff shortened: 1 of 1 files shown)


oznogon:
The macOS build failure is `Compatibility with CMake < 3.5 has been removed from CMake ... add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway`.

The Windows cross-compile build failure is consistent with the build box not using the legacy branch of SeriousProton.

The Windows RelWithDebInfo build failure is `Generator Visual Studio 17 2022 could not find any instance of Visual Studio`, which seems likely to be a build box/GitHub Actions issue.

PR #2881, opened 2026-06-14
Upscale radar traces from SVG and pngcrush them

oznogon:
Upscale ship, blip, and arrow radar traces from 32x32 to 128x128 via SVG recreations of stock traces, and run pngcrush against all radar traces to reduce file sizes.

Trace shapes are retained, with as few differences I could manage from their current sprites. These traces should also have more consistent transparent margins, which results in some scaling differences that are particularly noticeable with space stations.

`hugestation` isn't part of this PR, since the trace of the sprite wasn't symmetrical enough when upscaled despite producing the same output as the original sprite.


oznogon:
Closing to redraw everything from scratch on a grid instead of trying to match the original sprites, which will probably wind up being closer to the rejected/abandoned #2482

PR #2882, opened 2026-06-14
 Replace mine blip trace with mine weapon icon 

oznogon:
Stacked on/requires #2881.

- Use a 128x128 marginless version of the mine weapon icon for mine radar traces.
- Change Lua API mine sizes (currently `10`) to match the default player-launched mine size (`16`).

Before: mines use a slightly different blip than blip.png, and mines launched by players default to a larger blip size than mines placed by Lua scenario scripting

Screenshot_20260614_123151
Screenshot_20260614_123144
Screenshot_20260614_123134

After: mines use the mine weapon blip, and mines placed by Lua scenario scripting are the same size as mines launched from a `medium` tube

Screenshot_20260614_123012
Screenshot_20260614_123004
Screenshot_20260614_122953


oznogon:
To capture a Discord conversation from 2024, this was rejected due to the icon's design.

PR #2887, opened 2026-07-10
add slow down function to reduce game speed to 10%

Piglit:
available as:
* GM button
* Lua function

This feature was requested for some larp-scenarios. I'm not sure if it should be part of the base game.


oznogon:
#2680 implemented a time-scale control as a GuiSelector, but for accelerated time scales. daid rejected it for physics concerns. The interface and script APIs could be adapted to also (or only) reduce the time scale.

PR #2898, opened 2026-07-24
Fix segfault when adding 'too many' custom buttons

GinjaNinja32:


daid:
... How many are you adding to run into this...

PR #2906, opened 2026-09-18
Add T. Hill-Hand to graphics credits

Jadael:
Tried to keep this as minimal as possible - just wanted to be credited for the radar traces.


daid:
Perfectly fine. You did a great job on those.

